PR #18, opened 2016-04-28
Fix typos involving articles.

waywardmonkeys:
This fixes a variety of duplicated articles (``the the``) and the like.


david135:
on line 2294, the correction should be "an the application" -> "the application", not "and the application".

yxy:
I was curious about how `the the` comes out ;)

waywardmonkeys:
@david135 Thanks! I had misread that. Fixed.

akeep:
Thanks for the the (and related) fixes!

PR #31, opened 2016-05-01
fixed incorrect handling of library-extension when searching wpo files

hyln9:
Example: cannot make use of ".chezscheme.wpo" file.
All the tests have been fully tested before and after the fix and properly revealed the bug.

### DIFF
--- a/LOG
+++ b/LOG
@@ -59,3 +59,6 @@
   object to a time-utc object.
     stats.c,
     date.ms
+- fixed incorrect handling of library-extension when searching wpo files
+    compile.ss
+    7.ms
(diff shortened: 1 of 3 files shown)


hyln9:
All the tests have been fully tested before and after the fix.

hyln9:
rebased

akeep:
@hyln9 thanks for the fix!  I extended your tests (ever so slightly) and added comma to logs.

The main changes to the tests were:

1. I added a check that the program generated by compile-whole-program was producing the expected results.
2. I changed the "magic" variable to bind a pair instead of a fixnum. This is because fixnums can be optimized across library boundaries, so the library is ultimately no needed when the program is run, but pairs cannot be optimized.
3. I made a similar test in the compile-whole-library section to test this there as well.

Thanks again for the fix!

PR #37, opened 2016-05-05
to print {} in a better way

xieyuheng:
thanks to rain1 and hyln9 on #chez

### DIFF
--- a/s/print.ss
+++ b/s/print.ss
@@ -1128,6 +1128,9 @@ floating point returns with (1 0 -1 ...).
              (write-char #\- p)
              (print-hex-char c p))]
         [(#\+) (if (fx= n 1) (write-char #\+ p) (print-hex-char c p))]
+        [(#\{ #\}) (if (fx= n 1)
+                     (write-char c p)
+                     (print-hex-char c p))]        
         [$constituent? (if (print-unicode) (write-char c p) (print-hex-char c p))]
         [else (print-hex-char c p)]))
     (s1 s p n 1))


xieyuheng:
although this makes some of the output better,
but there still problems about pretty-print.

```scheme
(pretty-print
 '(match {'a 2 3}
    [{'a b c}
     (let ([a 1])       
       {{a b c} a {a b c} b {a b c} c {a b c}})]))
```
I expect to get ::
```scheme
(match {'a 2 3}
    [{'a b c}
     (let ([a 1])
       {{a b c} a {a b c} b {a b c} c {a b c}})])
```
but I actually get ::
```scheme
(match {
  [quote a]
  2
  3
  }
  [{
   'a
   b
   c
   }
   (let ([a 1])
     {
     {
     a
     b
     c
     }
     a
     {
     a
     b
     c
     }
     b
     {
     a
     b
     c
     }
     c
     {
     a
     b
     c
     }
     })])
```

xieyuheng:
the use of {} is not specified by r6rs.
while chez is using it as normal symbol,
for example, '({}) is viewed as a list of two symbols -- '{ and '}

this makes it possible for a user to define his/her own macros
to use {} in anyway he/she wishs.

it will be great to see this fact documented in the documents of chez.

also, it is really necessary for those printer functions, such as pretty-print,
to handle {} in a proper way.

dybvig:
This change would introduce an incompatibility with R6RS, so I would want it enabled only when a parameter (say `print-extended-identifier-syntax`)  is `#t`.  If we had such a parameter, I would want `write` to handle other Chez Scheme extensions to the syntax of identifiers as well, such as the identifier `1+`.   This would require wrsymbol to make two passes over the symbol's name: one to determine whether the symbol falls into the extended category and another to do the printing.

(These extensions are available outside libraries and top-level programs except when prefixed by `#!r6rs` and inside libraries and top-level-programs only when prefixed by `#!chezscheme` ).

Extending `pretty-print` to handle braces the way you want when the parameter is set probably wouldn't make anyone unhappy.

xieyuheng:
sounds like a homework assignment, sir.
I will try after I finish the porgram I am working on.

dybvig:
I'm closing this pull request pending completion of a more complete update.

dybvig:
[This commit](https://github.com/cisco/ChezScheme/commit/603019ea82afda1926462214576ef92df15e43c8) adds a `print-extended-identifier` parameter that, when #t, causes the printer to print `{` and `}` and other extended identifiers, like `1+`, without escapes.

PR #45, opened 2016-05-13
Fix two typos and three minor inconsistencies.

michaellenaghan:
There are also inconsistencies with "non-zero" vs. "nonzero", but I didn't fix those.

(Let me know if trimming trailing whitespace is annoying.)


akeep:
Thanks!

PR #46, opened 2016-05-13
Make hrefs to tspl4 absolute.

michaellenaghan:
Since anyone can now "make docs" for csug, all hrefs to tspl4 should be absolute. Right now all of the tspl4 links on, for example, http://gwatt.github.io/csug/csug_1.html and http://cdn.virgil.xyz/csug9/csug_1.html 404. (Those are the index pages, but the same is true of the tspl4 links in the content—e.g., the first link on http://gwatt.github.io/csug/io.html.)


dybvig:
Thanks for these changes.

PR #48, opened 2016-05-13
Begin a refresh of the profiler's HTML output.

michaellenaghan:
* Mark the files as HTML5 rather than HTML4.
 * Use target attributes rather than onclick events to open links in specific windows.
 * Add a missing table row element.
 * Replace the deprecated name attribute with an id attribute (and replace the anchors with spans).
 * Replace the deprecated valign attribute with a style attribute.

### DIFF
--- a/s/pdhtml.ss
+++ b/s/pdhtml.ss
@@ -1,13 +1,13 @@
 "pdhtml.ss"
 ;;; pdhtml.ss
 ;;; Copyright 1984-2016 Cisco Systems, Inc.
-;;; 
+;;;
 ;;; Licensed under the Apache License, Version 2.0 (the "License");
 ;;; you may not use this file except in compliance with the License.
 ;;; You may obtain a copy of the License at
-;;; 
+;;;
 ;;; http://www.apache.org/licenses/LICENSE-2.0
-;;; 
+;;;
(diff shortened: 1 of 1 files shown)


michaellenaghan:
If there are no objections I'd like to do a style refresh next. These changes were meant to be non-controversial; those might be. We'll see. :-)

michaellenaghan:
As far as a style refresh goes, here are before-and-possibly-after screenshots. Let me know what you think. 

(Btw, I purposely went a bit retro on the Profile page in order to make it better match the source file pages.)

screen shot 2016-05-13 at 3 57 53 pm
screen shot 2016-05-13 at 3 57 56 pm
screen shot 2016-05-13 at 3 57 07 pm
screen shot 2016-05-13 at 3 57 11 pm

jltaylor-us:
Are you planning to put the more and less controversial portions in separate PRs?

I suspect some people may be attached to the dark background.  Maybe this could be configurable via parameter (much like the rest of the color palette is configurable via the `profile-palette` parameter?

michaellenaghan:
Yes, separate PRs. I put the screenshots here to see whether or not I should bother with the second one.

Nearly all of these colors are already configurable. And, of course, my changes go beyond the color scheme, so if they don't like *these* colors other changes may still be worthwhile.

PR #51, opened 2016-05-17
Continue the refresh of the profiler's HTML output.

michaellenaghan:
Detailed notes about the changes are in the [commit](https://github.com/cisco/ChezScheme/commit/f7d7109de6845f6e91d4a7d77063d46e7e0dff3f). Some other notes:

 * These changes are meant to be conservative.

 * Two other improvements could be done. Any interest?

    * Use CSS classes rather than inline styles

    * Use CSS classes rather than tables

 * You may want to consider a lighter color scheme; just change the first palette entry pair to #EEEEEE/black.

profilehtml-white

fatfibhtml-white

PS If you create new screenshots, note that the timestamp is an Easter egg. :-)


michaellenaghan:
Hmmm, I just noticed that the old screenshots have a "/tmp/fatfib" prefix in the file paths. Let me know if you'd like new ones that match.

michaellenaghan:
I added the paths to the screenshots. Sorry about the noise!

david135:
I might recommend avoiding whitespace-only changes, such as the removal of trailing spaces, which make it harder to find the more interesting.  There were many of these, for example, in system.stex.

michaellenaghan:
Yes, I've been wondering about that. Btw, in case you weren't aware, you can ask GitHub to ignore whitespace in its diffs by adding "?w=1", like this: https://github.com/cisco/ChezScheme/pull/51/files?w=1.

dybvig:
Whitespace changes are fine with me if they help keep the code clean or eliminate tabs from source files (as I've just done).  It can be helpful to put the whitespace-only changes in a different commit, particularly if there are a lot of them or a lot of other changes.

michaellenaghan:
Should I close this...?

dybvig:
No.  I got caught up discussing the colors and the color-assignment algorithm with a data visualization expert (Rachael Brady) in our group at Cisco.  For colors, she has suggested adding a dedicated color for executed once (to go along with the existing dedicated not-profiled and profiled-but-not-executed colors) and trying different shades of blue for all of the others except the highest-count group, which would get something like red or yellow, so only the latter really stands out.  I was going to play with that before committing your changes, but I haven't had a chance yet to do so.

michaellenaghan:
Right.

The colors I chose were from the [Material Design palette](https://material.google.com/style/color.html#color-color-palette). There's nothing magical about Material, it just has the benefit of being well specified. For example, there are tints and shades of each hue of blue. Convenient!

Changing the palette will mean re-doing the screenshots. I can do that or you can, but if you do it: I made the two screenshots exactly the same size in order to make them feel more like screenshots (and less like rectangles copied from a screenshot).

michaellenaghan:
Btw, I asked if there was interest in two other possible changes. The first would be to switch from inline styles to CSS classes; the second would be to use CSS for layout rather than tables. (On the first point, I was going to use an Atomic CSS approach, along the lines of [Basscss](http://www.basscss.com). Atomic CSS is highly modular and composable. The result would look and feel somewhat like the current inline styles but be more DRY.)

Those changes may well be past the point of diminishing returns, but: I thought the idea of using HTML for visualization was interesting, and I was thinking that we could slowly make all of this a better foundation for other visualizations.

dybvig:
Michael, I'm finally getting around to this, and I would like to merge the changes you made without making any additional changes (yet, at least).  The only thing stopping me right now is a LOG entry.  If you wouldn't mind, please update your repo and append an appropriate entry to LOG, possibly crafted from the commit-log text.

Also, Rachael seemed to like the lighter background, and I don't really have a preference, so if you also prefer the lighter background,  you should go ahead and commit that.

dybvig:
I don't have an opinion about the CSS changes you suggest.  Since I've heard no objections from others, I'd say go it if you're so inclined.

michaellenaghan:
I've added a LOG entry. I'll consider creating a separate pull request for the other color scheme at some point, but it may be enough to simply say somewhere that you can switch color schemes by changing the first palette pair to #EEEEEE/black.

PR #54, opened 2016-05-22
Bash test(1) does not allow bare numbers with ==, so use -eq

Abhayakara:
I wasn't able to get "make install" to work on my Ubuntu 15.10 system because of a minor error in the makefiles/installsh script.   Enclosed patch fixes the problem.

### DIFF
--- a/makefiles/installsh
+++ b/makefiles/installsh
@@ -61,7 +61,7 @@ while ${TRUE} ; do
     for file in $files ; do
       destfile=$dest
       if [ -d $destfile ] ; then destfile=$destfile/`basename $file` ; fi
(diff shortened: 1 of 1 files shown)


jltaylor-us:
The bash builtin `test` works just fine with numbers - it compares them as strings.  This script uses `/bin/sh`, though, which isn't guaranteed to be bash.
In any case, using `-eq` is consistent with comparisons elsewhere in the file, which is a good thing.

Abhayakara:
Ah, that makes more sense.   I'm afraid I was just guessing as to what had
gone wrong, since I didn't need to know to get the fix working. :)

PR #57, opened 2016-05-28
Fix doc bug.

michaellenaghan:


michaellenaghan:
Fix doc bug. (Change "ftype-lock!" to "ftype-unlock!".)

PR #70, opened 2016-06-10
fix thread blocked by expeditor under windows

### LINKED ISSUE #58: thread is blocked unexpectly
I tried to embed chez scheme to a program and  created a thread running a loop. But after a short while, the thread was blocked unexpectly and left the work unfinished.

I try simplified code in repl, and similar things happened.
As follow, after the first time I check a's value, I found it increase similar numbers each time I check it no matter how long I wait. More exactly, it depends on how many times I input things to the repl. The CPU usage is always 0%.

```
> (fork-thread (lambda () (let loop ([i 0]) (set! a i) (loop (+ i 1)))))
#<thread 1>
> a
537411480
> a
537900424
> a
538421584
> a
538945213
> a
539470650
> a
539994892
```
Is there a way to avoid this?

I use a 32-bit threaded version In Windows 7.

yjqww6:
fixes #58


dybvig:
This looks about right.  Please make the following changes: (1) Condition the deactivate_thread and reactivate_thread calls on DISABLECOUNT(tc) == FIX(0).  Disabling interrupts _should_ prevent automatic collections.  (2) Move the declaration of tc to the top of the function body.  (The assignment can remain where it is.)  Some C compilers don't like inline declarations, so we avoid them generally.  (3) Add a LOG entry describing the change.  You'll should pull the latest version from master so you are modifying the current copy of LOG.

PR #83, opened 2016-07-19
Expeditor unicode support

eraserhd:
This works for Mac OS X.

On Linux, characters entered with Ctrl+Shift+U #### aren't decoded properly (it seems to zero one of the bytes?)  It might be a couple days before I can investigate, but I leave this here for review of the approach.

### DIFF
--- a/LOG
+++ b/LOG
@@ -256,3 +256,7 @@
 - fixed three instances of unchecked mallocs reported by laqrix in
   github issue #77.
     io.c, schlib.c, thread.c
+- add unicode support to the expression editor.  entry and display now work
+  except that combining characters are not treated correctly for
+  line-wrapping.  this addresses github issue #32 and part of issue #81.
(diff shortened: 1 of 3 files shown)


eraserhd:
This is now ready!

Tested on Mac OS X 10.11.6 and Ubuntu 14 with gnome-terminal and xterm.

akeep:
The changes to expeditor.c break the Windows compile.  It looks like the s_ee_write_char function was shared between the Windows and Unix-like operating system code.

I'd like to see if we can get the Unicode repl support to work on Windows, but at the very least we need to avoid breaking the Windows version of the build.  I'll see what I can figure out :)

eraserhd:
@akeep Hurrah!  Thanks!

akeep:
@eraserhd Thanks for the patch!  Hopefully, I'll get a chance to get the windows repl similarly updated and add Unicode support to the command line handling

PR #88, opened 2016-07-29
Update docs: add how to install xQuartz on MacOS X

micfan:


dybvig:
While tempting, adding instructions for installing XQuartz would be inconsistent unless we provide similar instructions for other things upon which we depend, such as the required package files for various flavors of Linux or the required tools for various versions of Windows.  This is a slippery slope with no clear stopping point, and I'd prefer in any case not to take on the responsibility of maintaining correct instructions for other packages as things change.

PR #89, opened 2016-08-06
fix expeditor.c for 32bit hosts

okuoku:
A change in #83 broken build on 32bit glibc. The function `s_ee_write_char` has `s_ee_write_char(INT)` prototype.

Despite it fixes build on 32bit Linux, perhaps, the correct fix is doing opposite; define some type for character-in-expeditor and change `s_ee_write_char` prototype. (Sorry, but i'm too lazy to explore on this.)

cc @eraserhd

### DIFF
--- a/c/expeditor.c
+++ b/c/expeditor.c
@@ -1017,7 +1017,7 @@ static ptr s_ee_get_clipboard(void) {
 }
 #endif
 
-static void s_ee_write_char(wchar_t wch) {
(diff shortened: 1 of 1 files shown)


dybvig:
[This commit](https://github.com/cisco/ChezScheme/commit/50aa8b34774a9e65e96481ae329a3f329fad7aca) should fix the problem.  I would rather have pulled your fix but as you pointed out your change was not the correct fix.

PR #99, opened 2016-08-22
case and exclusive-cond syntax-error calls

gavwhela:
This commit fixes calls to syntax-error in case and exclusive-cond (s/syntax.ss) which had their arguments reversed causing exceptions like:
`Exception in syntax-error: #<syntax ((0))> is not a string`
Instead of:
`Exception: invalid case clause ((0))`
In cases such as:
`(case 0 ((0)))`
All tests pass on my machine, and there weren't any other obvious (ie static strings) cases of this.

### DIFF
--- a/LOG
+++ b/LOG
@@ -296,3 +296,6 @@
   can be overridden, as we do in our own make files.
     cafe.ss,
     7.ms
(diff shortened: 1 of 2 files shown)


dybvig:
Thanks for the fix (and the LOG entry)

PR #100, opened 2016-08-27
remove a duplicate let-statement.

FreeBirdLjj:
I found in the original `loop2` in `lookup`:

```
(let loop2 ([i 0])
  (if (fx= i n)
      (loop1 (env-next env))
      (if (eq? (vector-ref old-rib i) id)
          (vector-ref new-rib i)
          (let ([i (fx+ i 1)])
            (if (fx= i n)
                (loop1 (env-next env))
                (if (eq? (vector-ref old-rib i) id)
                    (vector-ref new-rib i)
                    (loop2 (fx+ i 1))))))))
```

the last let-statement (`(let ([i (fx+ i 1)]) ...)`) is the duplicate one of the `loop2` and can be replaced with just the recursive call of `loop2`. It seems like a manual loop unrolling, but according to the third guidance mentioned in the *CONTRIBUTING.md*,

> All other things equal, shorter code is preferable to longer code.

I decide to open this pull request.

### DIFF
--- a/s/cp0.ss
+++ b/s/cp0.ss
@@ -214,12 +214,7 @@
                           (loop1 (env-next env))
                           (if (eq? (vector-ref old-rib i) id)
                               (vector-ref new-rib i)
-                              (let ([i (fx+ i 1)])
-                                (if (fx= i n)
-                                    (loop1 (env-next env))
-                                    (if (eq? (vector-ref old-rib i) id)
-                                        (vector-ref new-rib i)
-                                        (loop2 (fx+ i 1)))))))))))))))
+                              (loop2 (fx+ i 1))))))))))))
 
     (define cp0-make-temp ; returns an unassigned temporary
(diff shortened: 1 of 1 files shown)


dybvig:
Good thought, but all other things aren't equal in this case.  The loop is indeed intentionally unrolled, and I just measured the benefit of the unrolling on an Intel i7-4790k cpu, a6le build, running 'make all' in the 's' directory to be about a 6-7% reduction in cp0 running time.  The overall compile-time reduction is a modest 1% or so.  If you want to repeat my test, run `make all pps=t | egrep '^cp0|^TOTAL'` a dozen or so times with and without the change.  Cpu time is in the 3rd column.

It usually pays to optimize environment lookup.

akeep:
I'm going to go ahead and close this PR, since there are performance reasons for the unrolled loop.

PR #104, opened 2016-09-12
Support for DragonFly build.

realzhtw:
Originally it was a very tiny patch. But these commits also contain bootstrap images as well as some auto-generated files. Not sure you want all of them. If not, I can strip it down to a few-lines-patch, so that the DragonFly build will still be possible but only with help of a Linux machine.

Btw, it was extremely easy to port. Thanks!


akeep:
Thanks for the pull request, this looks like a really good start at a port to Dragonfly BSD.

I do have a couple of requests.

1. Please add an Mf-a6df and Mf-ta6df to the mats directory.  The mats directory is where our unit tests live, and part of how we test Chez Scheme to make sure all of its parts are working is to run make all, make allx, or make bullyx in the mats directory.  I'd like to make sure that Chez Scheme on Dragonfly BSD can pass the allx mats before we merge this pull request.

2. When we make changes to Chez Scheme we update the LOG file in the base directory of the repository.  Please add a log entry.

I also have a separate question:  Does the 64-bit version of Dragonfly BSD support a 32-bit tool-chain?  Or is there a 32-bit version of the DragonFly BSD OS?  If so, we should consider adding an i3df and ti3df for 32-bit support.

realzhtw:
Thanks for your answer.
I will try to do what you've asked in the next few days.

As for the 32-bit version of DragonFly, they discontinued support for it since release 4.0 (see https://www.dragonflybsd.org/release40/). Neither do they support 32-bit tool chain on 64-bit version as far as I know, but I will double check this.

realzhtw:
I submitted Mf-a6df in mats and run the tests. Some failed. During the investigation I found a bug in DragonFly's version of clock_gettime. I filed a bug report (https://bugs.dragonflybsd.org/issues/2951), but there is definitely a workaround: instead of clock_gettime getrusage can be used. If the bug is not fixed quickly in DragonFly, I'll submit my workaround.

mflatt:
The v9.9.9 merge 47daa9b34016de84fd111801d9d589d15a523abe includes adjustments to make a DragonFly build work, although it's classified as "fb" (for FreeBSD).

PR #106, opened 2016-10-12
Correcting language stats

cjfrisz:
Apparently Github detects .ms files as MAXscript, making the mats files look like they're not Scheme. It also thinks the .h files are C++ as opposed to C. This pull request simply adds a .gitattributes file to hint the correct languages to Github.


burgerrg:
Thank you!

PR #112, opened 2016-11-02
Fix profile counters for non-s-expression source

cjfrisz:
Check both the beginning file pointer (bfp) and end file pointer (efp)
of the source location associated with a profile counter when updating
its count.

Assuming that each expression has a unique bfp with respect to profiling
seems to give accurate execution counts for s-expression-based source
locations as in Scheme, but causes problems when targeting other kinds
of syntax. For instance, a C-style function call, referencing the called
function by name, such as "fn(arg)", can logically have profile counters
associated with 1) the function name reference ("fn") and 2) the entire
function call expression ("fn(arg)"), both of which begin at the same
source location. Only the bfp is checked when updating profile counters,
so the two source locations are conflated, and only one counter is
incremented, which gives inaccurate execution counts for both locations;
approximately twice as many for one, and zero for the other.

### DIFF
--- a/LOG
+++ b/LOG
@@ -322,3 +322,6 @@
     examples/csocket.c, examples/socket.ss
 - use high-precision clock time on Windows 8 and up
     c/stats.c
+- fixed profiling code that keyed profiling locations off of only the
+  bfp to instead key off of both the bfp and efp.
+    pdhtml.ss
--- a/s/pdhtml.ss
+++ b/s/pdhtml.ss
@@ -201,13 +201,17 @@
       (let ([fdatav (hashtable-values fdata-ht)])
         (vector-for-each
           (lambda (fdata)
-            (let ([entry* (sort (lambda (x y) (> (entrydata-bfp x) (entrydata-bfp y)))
+            (let ([entry* (sort (lambda (x y)
+                                  (or (> (entrydata-bfp x) (entrydata-bfp y))
+                                      (and (= (entrydata-bfp x) (entrydata-bfp y))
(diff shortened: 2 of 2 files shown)


akeep:
I was getting ready to incorporate this change, but I started thinking about this in depth, and I was wondering what the situation is where two expressions could have the same starting file position and different ending file positions?

Part of what I'm wondering is if ignoring the end file position is actually an optimization recognizing that the beginning file position is already unique.

Do you have an example?

cjfrisz:
The changes in the pull request probably wouldn't be unnecessary if not for the ``(profile source-object)`` form (documented [here](https://cisco.github.io/ChezScheme/csug9.4/system.html#./system:s130)) which is intended for use cases in which Chez Scheme is the target language of another compiler or interpreter.

I had written a longer explanation recapitulating the text of the original comment before I realized that the pull request never mentioned being able to leverage Chez Scheme's profiling from when compiling from other source languages. I believe your assertion that each expression's bfp is unique is correct in Scheme and similar s-expression-based languages. However, if the source language has a C-style ``function(argument [, argument]*)`` syntax, there can logically be a profiling counter for ``function`` and one for ``function(argument [, argument]*)``, both of which have a source object with the same bfp.  As implemented without the changes in the pull request, the profiling output code will increment whichever counter it finds first for both the reference to ``function``, as well as for the entire function call expression. This gives incorrect execution counts for both expressions.

akeep:
Yes, that makes sense.  I should have read the title more carefully :)

Thanks for the clarification!

-andy:)

PR #129, opened 2016-12-05
fix ee_read_char function to handle ^@ properly on linux

5pyd3r:
when ee_read_char read the ^@ or ^Space, the value in buf[0] is '\0'. Since this character also means the end of the string, it fails the mbrtowc function and raises an exception.

### DIFF
--- a/c/expeditor.c
+++ b/c/expeditor.c
@@ -675,11 +675,15 @@ static ptr s_ee_read_char(IBOOL blockp) {
 #endif /* PTHREADS */
 
     if (n == 1) {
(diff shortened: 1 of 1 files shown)


akeep:
This change looks good.

Please add a short description of what you've changed in the LOG file.

Thanks!

akeep:
Just checking back in, did you want to add a short description of the change to the LOG?

burgerrg:
Thanks for the fix! It affects macOS as well as Linux. I updated the LOG and committed in e962a03987470d0a3937446c10af3a94793ffc43.

PR #139, opened 2016-12-22
Write "twelfth" instead of "twelveth" in format

gwatt:
Just a typo fix


akeep:
I think this looks like a good change, but there is a unit test in the file ``format.ms`` that checks for the formatting of numbers which also needs to be updated with this change.  (Check line 669 of format.ms).

Please update the unit test and also add a comment to the LOG file.

You can run the unit test by running ``make test`` at the top-level directory if you've made your changes in the same directory created by ``make``.  You can also go directly into the machine type directory's ``mats`` directory (e.g. ``a6osx/mats`` for 64-bit Mac OS X or ``a6le/mats`` for 64-bit Linux) and run ``make all`` (which runs one set of tests) and check the ``report-*`` file or ``make allx`` (which runs several variations of the ``all`` tests) and check the ``summary`` file.

Thanks!

akeep:
Thanks for the updates!

I'll try to get this merged this weekend.

akeep:
I got a chance to run the tests tonight, and everything looks good!  Thanks again, and thanks for updating the LOG file.

PR #141, opened 2017-01-09
add cmake file

hyphenlee:
add cmake file,tested under x64 windows 10


dybvig:
Thanks for the pull request.  What problem with the current configure / make process does this address?  Unless you've identified something that doesn't work that can't reasonably be solved with the existing process, I'm not inclined to add support for a new one.

hyphenlee:
@dybvig I try to build under Windows, but current build system not work well. I think cmake is a more convinient and universal solution, though I aware that the conversion cost from make to cmake is too much to be impossible, and everyone has his own preferred build system. I just have found that make/configure is unfriendly for windows user, and I have done some basic work. Perhaps someone needs it.

PR #148, opened 2017-02-26
Add `date-dst?` procedure.

samth:
Accesses existing-but-hidden DST information in date records.

This PR does not provide a way to explicitly construct date
objects where `date-dst?` produces #t, since that would
require a change to the `make-date` signature. Instead, only
dates constructed with `current-date` or `time-utc->date`, or where the OS determines that the date provided to `make-date` is in DST
will have `date-dst?` as #t.

Todo:

- [x] rename to `date-dst?`
- [x] fix test failures 
- [ ] test on Windows
- [x] figure out whether `make-date` ever sets `isdst` to `#t`.


dybvig:
Thanks!  This is a model pull request, complete with LOG entry, release note, user's guide entry, and appropriate tests.

I suggest renaming `date-isdst` to something with a `?` on the end, e.g., `date-dst?` or `date-daylight-savings-time?`.

Looking at `S_mktime` (in stats.c), it appears `tm_isdst` is sometimes set to 1; if so, dates created by `make-date`, which calls `S_mktime`, _can_ have isdst set.  The setting of tm_isdst presumably depends on the current locale; if so, tests and examples should take this into account somehow.  In particular, tests should not fail if run in a different locale.

Have you verified that isdst is computed correctly under Windows?  The code within `ifdef win32` ... `endif` in `S_mktime` has me wondering.

samth:
I'll go with `date-dst?`, which is the name we use in Racket as well. I have not tested on Windows, I'll see about doing that, although that `#ifdef` block does not seem to change the `tm_isdst` value.

Reading [this man page for mktime](https://linux.die.net/man/3/mktime) suggests that it will only guess if the input is `-1` for `tm_isdst`, whereas we're setting it to 0 in `S_mktime` in `stats.c`. Should that be changed to `-1`?

jltaylor-us:
A note for the docs:  While "Daylight Savings Time" is certainly in widespread use, the official name is "Daylight Saving Time".

(You can go learn all kinds of fascinating stuff by poking around on Wikipedia, but here's a link to the last US legislation that update DST, wherein you can find its reference to adjusting "the Daylight Saving Time": https://www.gpo.gov/fdsys/pkg/PLAW-109publ58/html/PLAW-109publ58.htm)

dybvig:
The referenced man page goes on to say that `tm_isdst` is set "regardless of its initial value" to a positive value or 0, but unhelpfully doesn't say how the value is affected by the supplied `tm_isdst` value.

The MacOS page also mentions that negative values for `tm_isdst` forces it to guess the time zone.  Yet a quick experiment on my Mac shows that, with an initial value of 0 rather than -1, it sets `tm_isdst` non-zero for July 1, 2017 but not January 1, 2017.  So there appears to be some correction going on.  I didn't get into it far enough to determine if an incorrect initial `tm_isdst` value also causes other fields to be adjusted when the `tm_isdst` value is corrected.

This all leads me to believe you might be right that we should be setting `tm_isdst` to -1 before calling `mktime`.

dybvig:
You're right that the `ifdef` block for win32 doesn't modify `tm_isdst`; the question is whether it should.

burgerrg:
I think that setting tm_isdst to -1 before calling mktime will eliminate the need for the WIN32 code. See https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/mktime-mktime32-mktime64

samth:
One question while I work on testing these changes -- how do I run a subset of the tests, say just `date.ms`?

samth:
I think this is now done, except for testing on Windows.

dybvig:
To run just date.ms, cd into the workarea's mat directory and run 'make date.mo' (removing date.mo first if necessary), then grep for errors, bugs, etc., using `egrep "Bug|Error|invalid memory" date.mo`.  To run an even smaller set,  you can create a temporary .ms file.

burgerrg:
Why don't we add a tri-state dst argument to make-date? Then make-date would have all the options. I've written production code that uses make-date with UTC dates where I do not want the DST offset applied.

burgerrg:
I don't understand what the WIN32 code is doing. Why does it apply the time zone and DST offset of the _current time_ to the date/time specified? GetTimeZoneInformation doesn't look at the date specified.

dybvig:
That would probably be best.  I suggest adding an optional boolean argument at the end of the argument list.  The three states would be not present (corresponding to `tm_isdst` = -1), present and `#t` (1), and present and `#f` (0).

dybvig:
Sounds like the win32 code is wrong and should be removed, per your earlier suggestion.

burgerrg:
Unfortunately the #else part doesn't compile under Windows. I don't understand what's going on with the tz offsets in this code, so I don't know what to recommend.

samth:
Racket has very similar code for Windows, see here: https://github.com/racket/racket/blob/master/racket/src/racket/src/fun.c#L10091-L10115 which it also uses to compute whether the date is in DST.

burgerrg:
The Racket code for Windows looks good. Can the S_mktime code set the tmx.tm_gmtoff before calling mktime and avoid the difftime? Unfortunately Windows doesn't have the tm_gmtoff member.

mflatt:
I think the right repair for Windows is

```
  {
    TIME_ZONE_INFORMATION tz;
    long tzoff;

    /* The ...ForYear() function is available on Windows Vista and later: */
    GetTimeZoneInformationForYear(tmx.tm_year, NULL, &tz);
    
    if (tmx.tm_isdst)
      tzoff = (tz.Bias + tz.DaylightBias) * -60;
    else
      tzoff = (tz.Bias + tz.StandardBias) * -60;

    if (tzoff != orig_tzoff) tx = (time_t) difftime(tx, (time_t)(orig_tzoff - tzoff));
  }
```

Note that it uses a function that's only available in Vista and later. Racket checks dynamically for the function and uses an older one on XP and earlier, but I don't know if XP an earlier is an issue anymore.

A possible locale-independent test to demonstrate the problem with the old code: Get the difference between June 15 and January 15 in seconds, and check whether that difference is consistent with the DST results for each. There are 3 possible cases: no DST (or, more generally, same DST for January and June), shorter by an hour in the northern hemisphere with DST (i.e., January is not DST and June is), or longer by an hour in the southern hemisphere with DST – assuming that DST is always a 1-hour shift.

The `date-dst?` test in the current commit seems locale-specific, since June 15, 2016 was not in DST everywhere. The only idea I have for a test involves adding a way to lookup the time zone abbreviation (which would be handy for Racket on Chez Scheme), and then a test could work in a handful of recognized time zones.

mflatt:
I've pushed a variant of Sam's patch to mflatt/ChezScheme@b92ae167fc91f377f64e218a4ca144c0616b0a19 (rebased to master).

It ended up being more of a change than I planned. While looking into time-zone names, I saw that there was no way to create a date with an arbitrary day such that the zone offset is set to be consistent with Daylight Saving for that day. (Instead, the zone offset would stay as provided to `make-date`.) Also, if you provide a time-zone offset to `make-date`, then the result of `date-dst?` doesn't match up with that offset.

So, I made the zone offset optional for `make-date`. When not provided, the offset is determined by the system time zone setting as appropriate for the target day, `date-dst?` is set appropriately, and `date-zone-name` is now set. When a zone offset is provided, then `date-dst?` and `date-zone-name` of the date both return #f.

Date constructions without an explicit zone offset now allocate an extra string each time, but I expect that's ok. Similarly, there's a `copy-vector` in `date->time-utc` to avoid mutating the given date's time-zone information.

@samth If the variant looks ok, then I suggest moving it to your branch for this pull request. (Note: I removed the `true` flag from the "primdata.ss" spec for `date-dst?`, since the selector doesn't always return a true value.)

samth:
I've adopted @mflatt's changes, fixed an error on Linux discovered by @spall, and pushed here. Building bootfiles/running tests now, may need some line number fixes in the tests.

samth:
Bootfiles built, still some line number errors to repair, I think. @dybvig, any suggestion for how to maintain the mats files other than manually?

burgerrg:
The changes for Windows look good.

burgerrg:
The mat patch file changes are incorrect. They cause lots of false reports for o=0 spi=t p=t on a6nt.

dybvig:
`make patches` can be used after a mat run to update the relevant patch files, once you've verified there are no actual errors.  `make just-reports` can be used to verify that the patches were made correctly, without rerunning the mats.

samth:
This is now done, and successfully passes `make test`.

burgerrg:
@samth, all of the patch files need to be updated. In particular, patch-compile-0-t-f-f, patch-interpret-0-f-f-f, and patch-interpret-0-f-t-f need to be updated for make allx to succeed in the mats directory.

samth:
@burgerrg I tried running `make patches` in `ta6le/mats`, but I get errors like this:

```
rm -f patch-compile-0-t-f-f
diff --context errors-compile-0-f-f-f\
                        errors-compile-0-t-f-f\
                        > patch-compile-0-t-f-f
Mf-base:430: recipe for target 'xpatch-compile-0-t-f-f' failed
make[1]: [xpatch-compile-0-t-f-f] Error 1 (ignored)
```

burgerrg:
I haven't done this before, either, but I'm guessing that we need to run all the combinations first.

```
for EVAL in compile interpret ; do
  for O in 0 2 3 ; do
    for SPI in f t ; do
      for CP0 in f t ; do
        for CIS in f t ; do
          make report eval=$EVAL o=$O spi=$SPI cp0=$CP0 cis=$CIS
        done
      done
    done
  done
done
```

and then run `make patches`.

samth:
Ok, trying that.

samth:
That took quite a while, but didn't fix the errors in `make patches`.

burgerrg:
I don't think the ignored failures are relevant. Doing this procedure should have fixed several of the patch files and introduce a couple new ones. Be sure to `cp patch-* ../../mats` so that git observes the changes.

burgerrg:
@dybvig, should we add an all-reports target to mats/Mf-base that does $(MAKE) report on every combination? Do we need every combination? I don't completely understand the patch system for mat.

dybvig:
I generally update the patch files lazily as I need them.  If you want to be eager about it, running `make allx bullyx` in the workarea mats directory generates a more complete set of error files than just `make allx`.  The `bullyx` target takes a long time to run, so it's not run by default.

burgerrg:
@samth, your updated patches worked for Windows. Please squash that commit with the original commit, rebase onto master, and then update the boot files. Thanks!

samth:
@burgerrg Done.

burgerrg:
Thanks, Sam! I'm going to see if we can skip the bootfiles commit and combine with a fix for issue #182.

dybvig:
Merged manually along with fix for issue #182.

PR #151, opened 2017-03-10
add record-equal+hash

mflatt:
This patch adds `record-equal+hash` for getting or setting a
customization for `equal?` and `equal-hash` on an RTD to apply to its
instances.

Potential issues:

 * The API explicitly uses a pair of functions, and the identity of the
   pair is relevant for determining when the pair's `car` is used for
   `equal?`. Any suggestions for a more tasteful data structure or API?

 * Relatedly, although there's a check when a pair is provided that its
   `car` and `cdr` have procedures, the pair could be mutated at any
   time. The effect of mutation is not particularly specified.

 * When using `record-equal+hash` to get an RTD's current mapping,
   there's no way to distinguish an explicit mapping versus inheriting
   and ancestor's mapping. The difference can matter if the ancestor's
   mapping is changed (in which case the change is inherited).

 * The mapping is implemented by a property on the RTD's UID. Is there
   a convention for property keys to avoid conflicts?

 * The code's indentation doesn't quite match the rest of the Chez
   Scheme code. If that's important, what Emacs configuration would do
   the right thing?

### DIFF
--- a/LOG
+++ b/LOG
@@ -419,3 +419,10 @@
   occupy up to 1/2 of virtual memory, strings and fxvectors 1/4,
   and bytevectors 1/8.
     cmacros.ss
+- added record-type-equal-procedure, record-type-hash-procedure,
+  record-equal-procedure, and record-hash-procedure to enable
+  per-type customization of the behavior of equal? and equal-hash
+  for a record value
+     5_1.ss, newhash.ss, primdata.ss,
+     record.ms, root-experr*,
+     objects.stex
--- a/csug/objects.stex
+++ b/csug/objects.stex
@@ -2260,6 +2260,208 @@ requires a nongenerative clause.
 The default value is \scheme{#f}.
 The lead-in above describes why one might want to set this to \scheme{#t}.
 
+\section{Record Equality and Hashing\label{SECTRECORDEQUALTYANDHASHING}}
+
+\index{record equality}\index{\scheme{equal?} on records}%
+By default, the \index{\scheme{equal?}}\scheme{equal?} primitive
+compares record instances using \scheme{eq?}, i.e., it distinguishes
+non-eq? instances even if they are of the same type and have equal
+contents.
+A program can override this behavior for instances of a
+record type (and its subtypes that do not have their own equality
+procedures) by using
+\index{\scheme{record-type-equal-procedure}}\scheme{record-type-equal-procedure}
+to associate an equality procedure with the record-type descriptor
+(\var{rtd}) that describes the record type.
+
+When comparing two eq? instances, \scheme{equal?} always returns
+\scheme{#t}.
+When comparing two non-eq? instances that share an equality procedure
+\var{equal-proc}, \scheme{equal?} uses \var{equal-proc} to compare
+the instances.
+Two instances \var{x} and \var{y} share an equality procedure if
+they inherit an equality procedure from the same point in the inheritance
+chain, i.e., if
+\index{\scheme{record-equal-procedure}}\scheme{(record-equal-procedure \var{x} \var{y})}
+returns a procedure (\var{equal-proc}) rather
+than \scheme{#f}.
+\var{equal?} passes \var{equal-proc} three arguments: the two
+instances plus a \var{eql?} procedure that should be used for
+recursive comparison of values within the two instances.
+Use of \var{eql?} for recursive comparison is necessary to allow
+comparison of potentially cyclic structure.
+When comparing two non-eq? instances that do not share an equality
+procedure, \scheme{equal?} returns \scheme{#f}.
+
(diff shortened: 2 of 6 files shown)


dybvig:
The implementation of (even one-argument) `record-equal+hash` is not thread-safe due to the side effects of path compression, thus neither are `equal?` and `equal-hash`.  This is a show stopper.  Another concern I have is that `record-equal+hash` does not do path compression when no ancestor has been given equal and hash procedures.

I checked the existing `record-writer` implementation, and it has one of the same thread-safety issues.  While that aspect of its implementation is easily repaired, its use of an eq hashtable for recording record writers has the downside of holding onto an rtd in the common case where the associated record-writer procedure points back to the rtd, as you pointed out via email.  Since the `record-writer` and `record-equal+hash` mechanisms are related, we should continue to look for a better common solution for both.

Though some might be moot, I'll address the issues you raised:

- In place of `putprop`, `getprop`, and `remprop`, use `$sputprop`, `$sgetprop`, and `$sremprop`, which manipulate the symbol's "system" property list.  That way the properties can't conflict with user properties.

- You should be able to remove the pair from the interface (and implementation) by comparing the equal procedures rather than the pairs.  Another possibility would be to compare the rtds at which the most specific equal+hash record is found.

- I would be tempted to add an optional (or non-optional) boolean argument to your `record-equal+hash` accessor to say whether to follow the inheritance chain.  (This would require a switch to separate accessor/setter procedures.)

- I don't use emacs and so can't help you with a editing mode, but yes we would prefer that the code be indented similarly throughout.  That said, no indentation problems jumped out at me in your changes.

Also, rather than assign the outer `i` in `equal-hash`, please bind and assign a local `next-i`, like you've done with `next-k` in `equal?`, to localize the effect.  Please use `define-record-type` in place of `define-record`, with `(nongenerative)` and `(sealed #t)` clauses, and mark only the inheritors field mutable.

mflatt:
@dybvig Thanks for the review and corrections!

As long as the thread-safety issues are also addressed, I think path compression should always work with the existing implementation due to an "empty" record that is installed when an RTD has no setting and no parent.

Is using `with-tc-mutex` around the property manipulations (being sure to check for an existing system-property value within the mutex for an "empty" record) enough to avoid the thread-safety issues that you see? Of course, that wouldn't be enough to make equality-installing operations on RTD allowed concurrent to using `equal?` on instances of the RTD (or other installations on the RTD or its children), but that unsafety seems ok and consistent with other choices.

Thanks for the suggestions on excising pairs from the API. Relying on the identity of the customized RTD is closer to where I started, and I lost my way during some other bad turns. I have in mind making `record-equal+hash` return an RTD as well as the equality and hashing procedures,  which generally addresses the issue of distinguishing inherited versus direct results.

dybvig:
Ah,  I missed the installation of the empty record.

Using `with-tc-mutex` as you suggest should work except where it interferes with other calls to `$sputprop`.  There are only a handful of places where `$sputprop` is used at run time, and we can wrap them in `with-tc-mutex` forms as well.

Returning the rtd as well as the procedures (as multiple return values, I assume, now that the pair is gone) is a good solution.

BTW, using `set-who!` rather than `set!` for `record-equal+hash` would allow the quoted occurrences of `record-equal+hash` to be replaced with `who`, which is shorter and safer.

mflatt:
Only `$sputprop` and not `$sgetprop`? Is there something like a TSO guarantee so that `$getprop` doesn't need to be wrapped?

mflatt:
I tried to wrap all the `$sputprop`s that could affect uids, but maybe that's not right if any symbol can be a uid. If so, would it be better to put the `with-tc-mutex` in `$putprop`?

dybvig:
Yes, any call to `$sputprop` might be a problem.  I would not put the `with-tc-mutex` wrapper in `$sputprop`, however, since the wrapper might more appropriately be placed around a larger section of code at the call site, or might already be there.

We should also wrap run-time calls to `$sremprop`.

As property lists are currently implemented, `$sgetprop` is thread-safe even in the presence of concurrent calls to `$sputprop` on the same symbol and even the same key.  That's unlikely to change.

mflatt:
On closer inspection, I see that `with-tc-mutex` was generally in place around `$sputprop`s and `$sremprop`s, except in initialization places where no threads can be running. I ended up keeping only 2 additions in existing code (in "record.ss").

Just in case the patch is ready to merge, I've squashed and added a bootfiles commit.

mflatt:
This patch now has Kent's split into separate functions for setting equality and hashing functions and separate functions for getting the procedures to be used for a given set of equality or hashing arguments.

@dybvig - I noticed a few typos in the docs, and I added a small clarification on the default hashing function. Otherwise, this all works great for me.

PR #152, opened 2017-03-11
immutable vectors, strings, and bytevectors

mflatt:
This partial patch (no tests or docs, yet) adds support for immutable vectors, strings, and bytestrings, where `vector-set!` and similar operations check for mutability.

New `vector-set-immutable!`, `string-set-immutable!`, `bytevector-set-immutable!` operations change a value from mutable to immutable – as a one-way street – and `vector-immutable?`, etc., report whether a value is immutable.

For vectors, the high bit of the type+length field is used for an "immutable" flag. For strings and bytevectors, the length part of the field is shifted up by one to make room for the "immutable" bit. So, the runtime implications are:

 * halving of the maximum size of a vector, string, or bytevector – which could matter on a 32-bit platform;

 * adding a bit test in `vector-set!`, `string-set!`, etc.; and

 * adding a bitwise "and" to `vector-length` and the range check for `vector-ref` (but no new steps for `string-length` or `string-ref`, since a shift was needed already)

 * adding a range check in `make-vector`, etc., since the maximum size of a vector no longer matches the maximum fixnum.

The main question is whether those implications are acceptable enough to continue. I could also use help making sure that the generated code is a good as possible.

In case we get past those concerns: I didn't add an immutability bit to fxvectors, just because Racket doesn't have immutable fxvectors, but I could add it for consistency.


dybvig:
The change in representation your proposal lays out doesn't actually limit the lengths of vectors, since the high-order bit of the type/length field was previously always zero.  It also doesn't actually require mutators to do a separate mutability check, since they can mask in and check the mutable bit when they check the type.

It does limit the lengths of strings and bytevectors, and the main question is whether this is acceptable on 32-bit machines.  The prospect doesn't bother me as much as it once would have, because we have the 64-bit option.  How much longer will we care about 32-bit machines, anyway?

I would indeed want to extend the immutability option to fxvectors for consistency, which would be easy.  Ideally, I would also want to extend it to pairs, which would not be easy following the path you've proposed, since there's no place in the pair itself to store a mutable bit.

Here is a different proposal that supports an immutability option for all types of objects, including pairs; while it doesn't reduce lengths, it does waste space, particularly for smaller objects, on 32-bit machines:

- align objects on 16-byte rather than just 8-byte boundaries,

- use the extra primary tag bit as a mutable flag, and

- add an `immutable` procedure, which for an input value _x_ returns a value _y_ that is like _x_ except that, if the object to which _x_ refers can be mutated directly through _x_, it cannot be mutated directly through _y_.

That way mutability would be a capability of the pointer rather than an inherent property of the object itself, and an object remains mutable through the original pointer.  Is this sufficient for Racket's purposes, or can Racket be modified to make it sufficient?

On 64-bit machines, most objects (all but flonums) are already aligned on 16-byte boundaries, so the hit here is again significant only on 32-bit machines.

Aligning pairs on 16-bytes boundaries on 32-bit machines is a particularly big waste, since it effectively doubles the size of a list.  If we're willing to steal an extra primary code for pairs, say from symbols, we could leave pairs aligned on 8-byte boundaries on 32-bit machines and still be able to distinguish mutable from immutable pairs.  For example, we could give pairs primary type codes `#b110` (immutable) and `#b111` (mutable), so that after isolating the type bits, pair? would check to see if the result is 6 or greater, and mutable-pair? would check to see if the result is exactly 7.

Closures and flonums could remain aligned on 8-byte boundaries on 32-bit machines, since they are already immutable.

Of course, this doesn't reduce the waste for other small objects, like boxes, ratnums, small strings, small vectors, small bytevectors, and small record instances.

dybvig:
Never mind my alternative proposal---it adds unacceptable overhead to object field accesses, which would have to mask off the mutable bit.  Of course I didn't think of that until walking away from the keyboard and letting my mind clear.

mflatt:
Just to sync, I've updated the demo commit to include immutable fxvectors and fold the mutability test into the type test for strings, bytevectors, and fxvectors. I bet there's a better way to handle the vector tests, but I didn't find it. So, this commit still
 
 * halves the maximum size of a string, bytevector, or fxvector;

 * adds a bitwise "and" to vector-length and the range check for vector-ref; and

 * adds a range check in `make-string`, `make-bytevector`, and `make-fxvector.

Just to clarify to anyone else reading this on why I haven't tried to do anything with pairs: Racket's mutable pair is a separate datatype from its immutable pair, so Racket can just not expose `set-car!` and `set-cdr!`. The problem with Racket vectors, etc., is that operations like `vector-ref` are supposed to work on mutable and immutable vectors alike, and we'd like to have `vector-ref`, `string-ref`, `bytevector-u8-ref`, bytevectoru8-set!`, etc., run as fast as possible. Immutable boxes are also a problem, but also not much use, so I'm happy to do something slow with them.

dybvig:
The thing to do with vectors is to handle them more like strings, bytevectors, and fxvectors, i.e., put the immutable bit near the low end with the type tag.  There's no advantage to putting the immutable bit at the top, and it makes the mask larger, resulting in larger instructions, especially on 64-bit machines.  Vectors were being treated differently just to take advantage of the fact that the length could be a plain fixnum, which will no longer be the case.

To make this feasible, I'm working on a set of changes now intended to reduce the special handling of vectors to no more than a few cases, which will be flagged by meta-asserts.  Assuming I catch all of the cases, it should be relatively straightforward to treat vectors more like the others.

Immutable boxes would be easy to support with a bit in the tag field.  Pairs are the troublemakers.

dybvig:
With [this commit](https://github.com/cisco/ChezScheme/commit/564542d32bbae6b33cef808613238d5a4a2a8ee2), I've weeded out the special handling of vectors in the compiler and run-time system, commonized more of the string/fxvector/bytevector/vector code in the compiler, and tested with alternative type-vector, mask-vector, and vector-length-offset settings (`#b100`, `#b111`, 3 on a 32-bit build; `#b1000`, `#b1111`, 4) on a 64-bit build.  All seems to work, but I reverted the settings for vectors back to the fixnum type, mask, and offset, since that gives us shorter instruction sequences for some vector operations---for now.  For separate mutable and immutable vectors, you'll want to use the top type bit as the immutable bit (or you'll get a failing idiot check on start-up about vector type/length fields not looking like fixnums), so just what I listed above for immutable vectors, and for mutable vectors, use type `#b000` on 32-bit builds and `#b0000` on 64-bit builds.

mflatt:
Sync'ed to 564542d32bbae6b33cef808613238d5a4a2a8ee2, and changed the public interface to `vector->immutable-vector`, etc., instead of `vector-set-immutable!`.

Issues and questions:

 * Since `(make-vector 0)` always returns the same vector, `vector->immutable-vector` doesn't yet work on the empty vector. Is there a better approach than adding a new hand-coded primitive to allocate a fresh 0-sized vector?

 * To support immutable boxes, is the right idea to add a type tag `#b10001110` (to go along with `#b00001110`) and use a suitable mask for testing for boxes?

 * The new commit adds support for immutable values to fasl, but writing is disabled because bootstrapping will need an extra step. I think I know how to do that and end up with suitable bootfiles, but is the disabled approach the right idea?

dybvig:
If you're happy with the more restricted interface that doesn't alter objects in place, that's fine with me, but it does prevent one from defining a cyclic structure using mutation then marking all of the pieces immutable.  Also, if the public Scheme feature copies the object, I'd prefer the public C feature do so as well, especially since we might come up with an implementation that doesn't involve mutation someday.

Is it appropriate to mark the empty vector, bytevector, string, and fxvector immutable from the start?

To the question about boxes: "yes".  Try to find a bit within the first 8, however, so the tag check remains a byte comparison.

I don't understand the problem you have with bootstrapping.  Please explain further.

I believe `maximum-vector-length` should still be the same as `most-positive-fixnum`; if so, you appear to have gone too far on changes related to that.

`copy()` in gc.c can now put immutable vectors (and boxes) into `space_pure`.  Woe be unto the programmer who decides `#3%vector-set!` is a useful way to work around immutability.

Kent

mflatt:
Good point about constructing cyclic structures. Racket provides `make-reader-graph` as a primitive, so that the runtime system can construct cyclic immutable data. Is it better to keep the more restrictive interface (for future implementation flexibility) and add something like `make-reader-graph` to Chez, or is it better to keep the imperative interface after all?

On marking the empty string, etc. as immutable: Although the empty values are inherently immutable, making `...-immutable?` return true for them complicates the contract for functions like `make-string`, where the result should be mutable, or `string-copy!` (I forgot to add a check there, so far) where the destination string should be mutable and it makes sense to pass an empty string with a range of 0 characters to copy. On the whole, it worked better in Racket to keep the "mutable" empty values separate from the immutable ones.

Bootstrapping: when I uncomment the uses of `vector-immutable?`, etc., in "fasl.ss", then I got unbound-variable errors. I thought it was a bootstrapping issue, but maybe I'm doing something else wrong.

dybvig:
I'd be tempted to add `make-reader-graph` or in this one case allow the Racket implementation of `make-reader-graph` access the imperative routines in from the system module, rather than commit to a public imperative interface.

Adapting the language to the contract system seems rather backward, but okay.  Rather than allocating a fresh immutable empty vector (etc.) each time, I would create one at startup, like the mutable empty vector so that `eq?` returns `#t` when passed immutable empty vectors, just as it does for mutable empty vectors.

To cure the unbound variable errors, add stub definitions to patch.ss, e.g., `(define (vector-immutable? x) #f)`.   Back out the definitions once bootstrapped.

dybvig:
I just looked at the code a bit and have a few small suggestions.

When `(fx>= n 0)` is followed by `(fx<= n ---)`, this can be replaced by the single test `($fxu< n (fx+ --- 1))` if `(fx+ --- 1)` is a fixnum.  so the two tests at the start of `make-bytevector` in bytevector.ss:
```
(unless (and (fixnum? n) (fx>= n 0)) ($oops who "~s is not a nonnegative fixnum" n))
(unless (fx<= n (constant maximum-bytevector-length)) ($oops who "~s is too large" n)) 
```
can be replaced with:
```
(meta-assert (< (constant maximum-bytevector-length) (constant most-positive-fixnum)))
(unless (and (fixnum? n) ($fxu< n (fx+ (constant maximum-bytevector-length) 1)))
  ($oops who "invalid bytevector length ~s" n))
```

(vector-immutable? vector) (etc.)
  - forces `(and (vector? v) (not (vector-immutable? v)))` or the equivalent before doing `vector-set!`.
  - can we replace it with `(mutable-vector? x)`, which returns #t iff x is a vector and x is mutable?
  - if we keep `vector-immutable?`, `np-simplify-if` should optimize `(and (vector? v) (not (vector-immutable? v)))` pattern as it does, e.g., `(and (port? p) (binary-port? p))` and we should use that pattern in our checks, e.g., in bytevector-copy and `bytevector-fill!`.

It looks like bytevector-s64-native-ref got the set! treatment by mistake.

The new `always-immutable-flag` constant doesn't seem to be used.

I wouldn't mind a single "not an immutable XXX" condition message in place of the dual "not an XXX" and "immutable" messages being produced in various places, and it would simplify the code.  I don't mind the way it is, either, if that's preferable to you.

It's difficult for me to be sure from the github view, but it looks like cond might not be formatted consistently with other code (one space indentation instead of two before clauses)

mflatt:
Thanks for the comments. I'll look into those.

Meanwhile, new issues with the latest commit:

 * I added the immutable empty vector to `S_G` and made it accessible
   to `vector->immutable-vector` via a "(cs)null_immutable_vector"
   foreign procedure, while "fasl.c" accesses
   `S_G.null_immutable_vector` directly. Is there a better or more
   canonical way?

 * To make `bytevector=?` treat immutable and mutable bytevectors as
   equal, I had to introduce a new `len2` temporary. I don't know how
   much that matters; I think I couldn't inline the expression for
   `len2` in the `if` test because that position has a more restricted
   grammar than the right-hand side of `set!, but I'm not certain.

 * What would be the process of adding a `set-box!` library entry? I
   get a "missing no headroom libspec" if I just comment the `set-box!`
   definition in "library.ss". This doesn't matter for the commit, but
   I'm curious about the concept that I'm missing.

To be clear, when I referred to "contracts" before, I didn't mean the
implementation of a contract system like `racket/contract`, but instead
the way that programmers reason about contracts. For example, if a
programmer decides to implement `string-reverse!`, should it accept the
empty string? If so, and if the implementer includes defensive checks,
the implementer must be careful to check for the empty string in
addition to whether the string is immutable. If not, clients must be
careful to never pass an empty string. The contracts are easier to
reason about when mutability is independent of length (whether or not
the contracts are enforced or not at a given time through a contract
system).

mflatt:
@dybvig The latest commit follows your suggestions, including changing the error messages. I added both `mutable-vector?` and `immutable-vector?`, etc., to replace `vector-immutable?`, etc.

dybvig:
Instead of publishing the immmutable empty vector (etc.) via a foreign procedure, I would put it into a new immutable-empty-vector slot into the thread context (look for `fxlength-bv` in cmacros.ss and `FXLENGTHBV` in scheme.c and gc.c) where it can be accessed efficiently via `($tc-field 'immutable-empty-vector ($tc))`.  `faslin` can also reference it from there using IMMUTABLEEMPTYVECTOR(tc), so there's no need to put it into `S_G` as well.

Or what if `mutable-vector?` and `immutable-vector?` both return true for the empty vector?

Adding an extra local to `bytevector=?` should be fine and seems necessary.  To be on the safe side, we could look at the generated code (make library.asm) and see if the compiler has spilled something we'd rather it not and prevent it from doing so, if necessary, by using specific registers.

Library entries must be declared in cmacros.ss.  Look for `unbox` and add a similar line for `set-box!`.

mflatt:
Ok, that change is done. Unless I'm missing something, I don't have any pending issues at this point.

Making `mutable-vector?` and `immutable-vector?` both return true for the empty vector could make sense. I didn't pursue that, though, because it doesn't fit with Racket's existing API, where not-immutable => mutable.

dybvig:
Okay, having heard no objections, I'll proceed with the merge.  Expect a few minor changes as I go over the code and check the compiler output.  I'll let you know before I change anything major.

dybvig:
Using the command-line instructions to merge while adding some of my own modifications seems to have left github unaware that the branch was merged, so I'm closing this pull request explicitly.

PR #160, opened 2017-04-08
Fix reduction for map with optimization level 3

gus-massa:
I'm still not sure that this is an error, and I'm still not sure that this is the correct fix.  Anyway ... 

The result of

    (begin (#3%map / '(5 4 3 2 1 0)) 'something)

is `'something`, but I expect to see an error in the last `(/ 0)`.

The problem is in the reduction for `#3%map` when it is called only for effects. The reduction test that the procedure is _pure_, but I think that it must be _ivory_. 

If this is correct, I can make the same changes for `for-each` because it has the same problem.

The PR is splitted in two commits, but I expect to squash them before they are merged.

### DIFF
--- a/LOG
+++ b/LOG
@@ -444,3 +444,9 @@
   $close-resurrected-mutexes&conditions, $keep-live primitive added
     externs.h, prim5.c, thread.c, 4.ms, thread.ms, release_notes.stex,
     7.ss, cpnanopass.ss, np-languages.ss, primdata.ss, prims.ss
+- fix reduction for map and for-each with optimization level 3
+  to drop the expression, check that procedure has the correct
+  arity and is discardable or unsafe.
+  Also add a simplification for for-each with empty lists
+  with optimization level 2.
+    cp0.ss, 4.ms, primdata.ss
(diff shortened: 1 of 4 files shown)


dybvig:
Thanks.  It is a bug, and the fix looks good.

Expanding on your ivory? observation, we're also missing an `arity-okay?` check for the primitive.  For example, `(lambda (ls) (#3%map cons ls) ls)` improperly reduces to `(lambda (ls) ls)`.  Also, `for-each` has the same bugs.

Would you be willing to update your pull request to address these additional issues?  If not, I can address them after pulling your changes.

gus-massa:
I changed the criteria for the procedure from `(_pure_ and (unsafe or unrestricted))` to `(_discard_ and (unsafe or unrestricted))`. It’s more similar to the original code, and it’s weaker than `ivory`. I think that `pure` was too strict and avoid the optimization of some obvious good cases. I also added the `arity-okay?` check. 

I also fixed `for-each` in a similar way, and I added a reduction in `cp02` for the case where the lists are all empty, like `(for-each / ‘() ‘() ‘())`. The indentation is a little weird because it uses `null-rec?` that is a local function.

I didn’t add to `for-each` an especial case for expressions with explicit lists like `(for-each + (list 1 2) (list 3 4))`. I saw that `map` has a special case for this kind of expressions, but I am still not sure haw this will interact with the other special case that only `for-each` has with `cp03`. I’ll try to understand this and if I find something useful I’ll send a PR next week.

The bootfiles are in a separate commit, but I can rebase+squash them if necessary.

dybvig:
This looks good, except the LOG entry mentions cp0.ms where it should mention 4.ms.  Please update the LOG and either remove or regenerate the boot files so I can do a clean pull.

gus-massa:
I fixed the LOG and commit message cp0.ms -> 4.ms .

I also removed the commit with the boot files.

Issue #91, opened 2016-08-18
"cond-expand" needed for easier porting other libs to chez


ChaosEternal:
with inclusion of "cond-expand" in Chez, will reduce the effort of porting 3rd party libraries to Chez.
and seems it is standardized in r7rs

akeep:
Chez Scheme has a syntax-level ``cond`` expression called ``meta-cond`` that allows for conditional expansion of code.  The test expression in ``meta-cond`` is just a general Scheme expression that can be evaluated at expansion time, for instance the [example](http://cisco.github.io/ChezScheme/csug9.4/syntax.html#./syntax:s35) from the documentation uses ``(= (optimize-level) 3)``.

I could imagine providing this as functionality as ``cond-expand``, but looking at the R7RS documentation, it looks like the test expression for ``cond-expand`` is much more restricted (though it sounds like it could be generalized on an implementation specific basis).  However, Chez Scheme does not currently expose any "feature identifiers" or a way to determine what libraries are available.  Would you need this latter functionality for it to be useful for you?

ChaosEternal:
I really want r7rs compatible cond-expand

akeep:
The library side of this is interesting, since we look for and load libraries based on filenames, but don't necessarily know if a library is available unless is imported, it is possible that ``cond-expand`` would have the side-effect of loading the library into the library manager.

jltaylor-us:
Since this issue is still open, here's a couple of things I stumbled across today that indicate the r7rs-small "specification" of `cond-expand` is a hot mess.  It's not possible to have an implementation compatible with a spec that's nonsense.
http://comp.lang.scheme.narkive.com/3N1291hB/r7rs-missing-library-export-in-the-base-library
https://groups.google.com/forum/#!topic/scheme-reports-wg1/sMNF8k0h3So

mnieper:
R7RS's `cond-expand` appearing in the body of libraries and programs is, in fact, not specified well.

However, R7RS's `cond-expand` appearing in library declarations is without these problems and has clear semantics. For practical purposes, it should be enough to implement (and use) only this form of `cond-expand`.

ceving:
R7RS's definition of `cond-expand` is somehow broken.

You can do something like

    (cond-expand (r7rs) ...)

But according to R7RS `cond-expand` itself requires, that you first define:

    (import (scheme base))

But that is already R7RS syntax.

[...]

PR #161, opened 2017-04-09
Add cond-expand from SRFI-0.

bnoordhuis:
Motivation: no cond-expand makes it pretty much impossible to port libraries to Chez Scheme.

I tried to follow the house style as much as possible, apologies if I got it wrong.  Thanks for Chez Scheme, it's amazing!

Closes out #91.

### DIFF
--- a/s/syntax.ss
+++ b/s/syntax.ss
@@ -8931,6 +8931,31 @@
   (lambda (x)
     (syntax-error x "misplaced aux keyword")))
 
+;;; cond-expand from SRFI-0: https://srfi.schemers.org/srfi-0/srfi-0.html
+(define-syntax cond-expand
+  (lambda (x)
+    (syntax-case x (and else not or)
+      [(_) (syntax-error x "unfulfilled")]
+      [(_ [else body ...]) #'(begin body ...)]
+      [(_ [(and) body ...] more ...) #'(begin body ...)]
+      [(_ [(and test rest ...) body ...] more ...)
+       #'(cond-expand
+           (test (cond-expand ([and rest ...] body ...) more ...))
+           more ...)]
+      [(_ [(or) body ...] more ...) #'(cond-expand more ...)]
+      [(_ [(or test rest ...) body ...] more ...)
+       #'(cond-expand
+           (test (begin body ...))
+           (else (cond-expand ([or rest ...] body ...) more ...)))]
+      [(_ [(not test) body ...] more ...)
+       #'(cond-expand
+           (test (cond-expand more ...))
+           (else body ...))]
+      [(_ [name body ...] more ...)
+       (if (eq? 'chezscheme (syntax->datum #'name))
+           #'(begin body ...)
+           #'(cond-expand more ...))])))
+
 ;;; (define-record name pname (field ...))
 ;;; (define-record name pname (field ...)
 ;;;   ((field init) ...))
--- a/LOG
+++ b/LOG
@@ -444,3 +444,6 @@
   $close-resurrected-mutexes&conditions, $keep-live primitive added
     externs.h, prim5.c, thread.c, 4.ms, thread.ms, release_notes.stex,
     7.ss, cpnanopass.ss, np-languages.ss, primdata.ss, prims.ss
+- added cond-expand from SRFI-0
+    pretty.ss, primdata.ss, syntax.ss,
+    8.ms, syntax.stex
--- a/csug/syntax.stex
+++ b/csug/syntax.stex
@@ -1448,6 +1448,23 @@ below.
   [else (safe-frob x)])
 \endschemedisplay
 
+\section{Feature-based conditional expansion\label{SECTSYNTAXCONDEXPAND}}
+
+Expansion-time feature detection can be done via \scheme{cond-expand},
+which is similar to \scheme{meta-cond} but is portable across a wide
+range of Scheme implementations.
+Chez Scheme defines a single feature: \scheme{chezscheme}
+
+\schemedisplay
+(cond-expand
+  [chezscheme (display "Chez Scheme")]
+  [chicken (display "Chicken Scheme")]
+  [guile (display "Guile Scheme")]
+  [else (display "Some other Scheme")])
+\endschemedisplay
+
+See also: \hyperlink{https://srfi.schemers.org/srfi-0/srfi-0.html}{SRFI-0}
+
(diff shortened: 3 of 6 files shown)


s/syntax.ss
```diff
@@ -8931,6 +8931,31 @@
...
+      [(_ [(or test rest ...) body ...] more ...)
+       #'(cond-expand
+           (test (begin body ...))
+           (else (cond-expand ([or rest ...] body ...) more ...)))]
+      [(_ [(not test) body ...] more ...)
+       #'(cond-expand
+           (test (cond-expand more ...))
+           (else body ...))]
+      [(_ [name body ...] more ...)
+       (if (eq? 'chez (syntax->datum #'name))
```
bnoordhuis: This isn't very elegant but I couldn't get it to work by adding `chez` to the syntax-case's literals and adding a pattern, I presume because the symbol doesn't get exported.  If there is a better way to do this, I'll be happy to update the pull request.

dybvig:
Thanks for the pull request.  I have some questions and issues to raise before proceeding.

1. Your note says this closes out [#91](https://github.com/cisco/ChezScheme/issues/91).  The requester in that case was looking for R7RS `cond-expand` rather than SRFI-0 `cond-expand`.  Are there any relevant differences?

2. Is it necessary or even sufficient to build `cond-expand` into Chez Scheme?  Presumably, a library or top-level program using`cond-expand` would have to import it from `(chezscheme)` or another non-standard library.  If you have to modify an importing library or top-level program to do that, you might as well modify it to import from an external library that defines a Chez-Scheme-specific `cond-expand`.

3. Does the fact that Chez Scheme looks for libraries first under the extension "chezscheme.sls" already give you what you need?  If nothing else, you can put a Chez-Scheme-specific version of `cond-expand` in "cond-expand.chezscheme.sls" and a generic version in "cond-expand.sls".  Importing libraries can then import `cond-expand` from `(cond-expand)` and remain portable.

4. Informal usage aside, `chez` doesn't make sense on its own.  Please consider an alternative, such as `chez-scheme` or `chezscheme`.

bnoordhuis:
Hi, thanks for your comments.

> 1. Your note says this closes out #91. The requester in that case was looking for R7RS cond-expand rather than SRFI-0 cond-expand. Are there any relevant differences?

R7RS stipulates two feature ids that I wasn't sure whether to implement:

1. `r7rs` should evaluate to true only when the runtime is R7RS-compliant.  I don't think this applies to Chez Scheme, in which case evaluating to false is the right thing to do.

2. `(library id)` should evaluate to true when library _id_ is available.  The exact meaning of 'library' seems somewhat underspecified and implementation-specific.  FWIW, I've never seen it used in the wild.

My unscientific sampling suggests that 99% of cond-expand forms are just to detect the Scheme runtime so I figured I'd start out simple.  I can add test cases that explicitly check that the aforementioned features evaluate to false if you want.

> 2. Is it necessary or even sufficient to build cond-expand into Chez Scheme? [...]
> 3. Does the fact that Chez Scheme looks for libraries first under the extension "chezscheme.sls" already give you what you need?

I don't think that would be practical.  Most code I've seen simply assumes that cond-expand is available because that's how it works in other Scheme environments.

I think of it as the Scheme equivalent of `#ifdef __GNUC__`.  Having to jump through additional hoops to make it available would defy its purpose.

The linked lalr-scm pull request is a good example of cond-expand in a single-file library.  Your chezscheme.sls suggestion would destroy the single-file aspect and make porting needlessly complicated.  It would also have to be duplicated in every ported project.

> 4. Informal usage aside, chez doesn't make sense on its own.  Please consider an alternative, such as chez-scheme or chezscheme.

I can change that if you insist but it's similar to how other runtimes announce themselves (e.g., `chibi`, `chicken`, `gauche`, `guile`, `kawa`, etc.)  `chez-scheme` or `chezscheme` would look rather unidiomatic.

dybvig:
No identifiers are visible in an R6RS library or top-level program unless they are imported from somewhere.  After looking at lalr-scm, it seems you mean a different notion of library, i.e., unstructured code loaded into the interactive top-level via the `load` procedure.  Instead of supporting this less structured, less reliable, and less efficient packaging mechanism, I'd rather we push toward adoption of R6RS libraries and encourage other implementations to support analogues of the chezscheme.sls extension for implementation-dependent code.

`chezscheme` or `chez-scheme` is more consistent than `chez` for most if not all of the examples you list:  Guile => `guile`, Gauche => `guache`, Kawa => `kawa`, Chez Scheme => `chezscheme`.  I'm not sure about Chibi or Chicken; their web pages refer to them both with and without "Scheme" as part of the name, so `chibi` and `chicken` are as good as `chibischeme` and `chickenscheme`.  Chez Scheme is never properly referred to as "Chez", though people sometimes do so informally.

bnoordhuis:
I've renamed the identifier to `chezscheme` and fixed the conflict in LOG.

> Instead of supporting this less structured, less reliable, and less efficient packaging mechanism, I'd rather we push toward adoption of R6RS libraries and encourage other implementations to support analogues of the chezscheme.sls extension for implementation-dependent code.

That seems like an uphill battle if there ever was one.  I think you ought to go for pragmatism over idealism here: having an out-of-the-box cond-expand makes Chez Scheme more accessible to programmers coming from other Schemes.  That's an unequivocally good thing, IMO.

> chezscheme or chez-scheme is more consistent than chez for most if not all of the examples you list: Guile => guile, Gauche => guache, Kawa => kawa, Chez Scheme => chezscheme. 

As mentioned, I've renamed it but I disagree with this line of reasoning: the others are consistently referred to as `<name> Scheme` too, if for no other reason than to disambiguate between the runtime and the word it's named after.  `chezscheme` is simply going to look odd but that's your prerogative.

svenha:
Hi all.

Ben, thanks for pushing the very popular SRFI-0 (yes, cond-expand is somewhat ugly, but widespread in many code bases I have seen).
Most Scheme implementations provide a function to set additional feature identifiers. Is this implemented in your pull request, too?

Sven

bnoordhuis:
Hi Sven.  No, I implemented just the bare essentials.  I figured I'd start out small for easy reviewing and to make it less contentious.

akeep:
Since this is implemented as a single macro, it doesn't seem necessary to include this in the base of Chez Scheme.  It is worth noting that there is a package of Chez Scheme compatible SRFIs in https://github.com/arcfide/chez-srfi, which includes SRFI-0.

PR #167, opened 2017-04-20
Move unused terminals in L4.5 to later languages.

LeifAndersen:
L4.5 has several terminals that it does not use. Later languages,
however, do use them. So I moved the introduction site for those
terminals to the languages where they are first used.

### DIFF
--- a/s/np-languages.ss
+++ b/s/np-languages.ss
@@ -357,9 +357,7 @@
  ; introducing mvlet, and mvcall
   (define-language L4.5 (extends L4)
     (terminals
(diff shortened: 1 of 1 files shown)


LeifAndersen:
To be clear, I don't know why the terminals are introduced in L4.5 rather than the languages where they are used. Perhaps its historical, or perhaps its just for readability. Or perhaps its something else entirely that I am missing.

akeep:
@LeifAndersen these changes look good, please add a brief description of the change to the LOG file.

Thanks.

PR #168, opened 2017-04-21
Ephemerons

mflatt:
Here's a first cut (no documentation or tests) to add ephemerons:

  * `(make-ephemeron key value)` makes an ephemeron that holds `value` only as long as `key` is reachable, where `value` doesn't count as reachable from the ephemeron until `key` is otherwise determined to be reachable

 * `(ephemeron-value eph)` returns either the value or `#!bwp`, where a value goes to `#!bwp` when the key is unreachable, even if the value is otherwise still reachable

 * `(ephemeron-key eph)` returns either the key or `#!bwp`; the key and value go to `#!bwp` at the same time

 * `(ephemeron? v)` reports whether v is an ephemeron

Racket does not provide an `ephemeron-key` operation, and I doubt that it's useful, but it seems harmless. The other functions match Racket, which explains the `make-` on the front of `make-ephemeron`, except that Racket uses `#f` in place of `#!bwp`.

The implementation includes a new type tag for ephemerons and a new space (since the sweep action is special).

The GC change avoids quadratic-time ephemeron processing by registering each discovered ephemeron with the segment that contains its key, if the key has not been determined to be reachable already. The second part of https://gist.github.com/mflatt/1fd3015539f9a571fee429d59799e347 sets up an example that would exhibit quadratic-time behavior, otherwise.

Triggering ephemerons requires that copy() check the old-space segment for each object. This extra check on every copy() is not measurable in my tests so far. Where copy() used to take a space-kind argument, it now takes a space-info argument — which was always computed to get the space-kind argument, so I think that part of the change has no effect by itself.

An ephemeron is allocated with room for two extra pointers that are used at GC time: one for a list of a all discovered ephemerons, and one for a segment-specific list to trigger. Having the extra pointer space in the object is convenient, but I wonder whether it's better to allocate those administrative lists separately during collection.


dybvig:
Thanks for working on this.  If successful, it might be a useful substitute for weak pairs in weak eq hashables, which probably should probably have ephemeron-like behavior.

To that end, do you think it would be possible to make ephemerons a kind of pair, like weak pairs?  Extra fields should be okay, if needed.  gc support for mutable ephemerons could be tricky.

Incidentally, I just glanced at the first couple of diffs, and I noticed that faslin uses INITCAR and INITCDR to initialize fields of an ephemeron.  I don't think that will work (unless ephemerons become pairs) so I'm guessing it hasn't been tested.

I'll look closer at the code when I get a chance.

mflatt:
I initially considered making ephemerons like weak pairs, with the key and value as being `car` and `cdr`, respectively. That didn't seem useful in the same way as weak pairs, though, since you generally wouldn't want to put a chain in the value part of an ephemeron. Maybe you have a different arrangement in mind, with the value in the `car` part, something strongly held in the `cdr` part, and the key in an additional field?

I also had in mind that Racket distinguishes ephemerons from other datatypes, and so having a Chez-level empheron respond #t to `pair?` is inconvenient. That's a Racket-specific issue, though.

You're right, of course, that I didn't test the fasl part, yet.

dybvig:
Hashtables in Chez Scheme store the key and value in the car and cdr of a pair.  In this case, the cdr is not used to chain list elements.  For weak eq hashtables, the pair is a weak pair.  We could instead have the pair be an ephemeron, at least optionally, if ephemerons are pairs (needed for `hashtable-cell`).

Weak pairs are pairs, but they can be distinguished via the weak-pair? predicate.  We could do the same for ephemerons.  weak-pair? a slow predicate, however, since it has to index into the segment table.  It also involves a foreign procedure call, though we could fix that.

mflatt:
I forgot that Chez Scheme hashables expose the implementation's use of pairs though `hashtable-cell`, so now I understand the motivation for making ephemerons a subtype of pairs.

I'll revise the PR to make the additions just `ephemeron-cons` and `ephemeron-pair?`.

I don't immediately see an obstacle to making ephemerons mutable, except that a kind of race condition with the GC can leave you with a half-`#!bwp`ed pair. Did you have different problem in mind?

(To distinguish Racket-level ephemerons from pairs, I guess Racket will just introduce an extra record wrapper – the same as for weak boxes implemented with weak pairs.)

dybvig:
Okay, after thinking about the ephemeron gc problem a bit, I expected to see something like this in the code:

A. When sweeping an ephemeron in sweep() or sweep_space(), two cases:

  1. If the key is not oldspace or is oldspace and has been forwarded, relocate both the key and the value.

  2. Otherwise, associate the ephemeron with the key in a key->ephemeron table (implemented, say, as a hashtable)

B. After copying any non-flonum object in copy(), relocate the keys and values of all ephemerons associated with the object in the key-ephemeron table, and remove the associations from the table.

C. At the end of collection, "bwp" the key and value of any empheron present in the key-ephemeron table.

I'm probably missing something I might spot if I actually tried to implement it, but I think I see this basic structure in your code.  It looks like your code treats the segment-info structure as a coarse form of hashtable key, in effect, and that (perhaps because of this) it handles emphemerons in batches at the end of each sweep_space loop.  It also seems like the table might contain emphemerons for which the key has already been forwarded, and for this you need the trigger-next = #t flag.  Is that about right?

Using the segment-info structure in this way is clever, but I wonder if it is sufficient to eliminate all pathological behavior.  The copy() check for the presence of a list in si is less overhead than a hashtable lookup, but we might be able to make a hashtable lookup reasonably inexpensive in the common cases (once we figure out what those are).

dybvig:
Nobody should be mutating a pair during collection unless it's locked and being mutated from C code, in which case I'm not too excited about it.   The real problem is dealing with wrong-way pointers in `sweep_dirty` and an analogue of `resweep_dirty_weak_pairs`, though we might get away without the latter.

mflatt:
The revised commit changes the API to `ephemeron-cons` and `ephemeron?`.

The GC implementation is essentially as before. You're right that the
segment-info structure is acting as a coarse hash table, and so on. I
believe the worst-case complexity here is proportional to the square of
the number of objects M in a segment --- but I also expect that
worst-case time to be both unlikely and less than the total number of
objects, and using a per-segment trigger list minimizes the constant
factor on the number of copied objects. In any case, I'm happy to defer
to your judgment on the GC-time data structures and performance.

I don't see the need for something like `resweep_dirty_weak_pairs`,
since the pending list plays a similar role, and ephemeron handling in
`sweep_dirty` seems straightforward, but maybe it's beyond the frontier
of what I've understood.

The code at https://gist.github.com/mflatt/1fd3015539f9a571fee429d59799e347
is a step closer to a test suite, including a small initial test for fasl writing and reading.

mflatt:
With this third commit to add tests and docs, the pull request is meant to be complete.

csug/smgmt.stex
```diff
@@ -418,6 +427,89 @@ dropped, but makes no guarantees about when this will occur.
...
+(define x (cons 'a 'b))
+(define p (weak-cons x x)) ; \var{not an ephemeron pair}
+(set! x '*)
+(collect)
+(car p) ;=> (a . b)
+(cdr p) ;=> (a . b)
+\endschemedisplay
+
+\noindent
+As for weak pairs, the last two expressions of the middle example
```
jltaylor-us: I think this reads better as "As _with_ weak pairs..."

csug/smgmt.stex
```diff
@@ -418,6 +427,89 @@ dropped, but makes no guarantees about when this will occur.
...
+\entryheader
+\formdef{ephemeron-pair?}{\categoryprocedure}{(ephemeron-pair? \var{obj})}
+\returns \scheme{#t} if obj is a ephemeron pair, \scheme{#f} otherwise
+\listlibraries
+\endentryheader
+
+\schemedisplay
+(ephemeron-pair? (ephemeron-cons 'a 'b)) ;=> #t
+(ephemeron-pair? (cons 'a 'b)) ;=> #f
+(ephemeron-pair? (weaj-cons 'a 'b)) ;=> #f
```
jltaylor-us: typo in `weak-cons`

mflatt: Thanks!

ovenpasta:
Great.

weak-cons not weaj-cons ;)

c/gc.c
```diff
@@ -185,7 +190,19 @@ static IBOOL search_locked(ptr p) {
 
 #define locked(p) (sorted_locked_objects != FIX(0) && search_locked(p))
 
-static ptr copy(pp, pps) ptr pp; ISPC pps; {
+static void check_trigger_ephemerons(si) seginfo *si; {
+  /* Registering ephemerons to recheck at the gradularity of a segment
```
gwatt: granularity?

mflatt: Fixed - thanks!

dybvig:
After spending a lot of time over the past couple of weeks trying to benchmark the system w/ and w/o support for ephemerons, I've come to the conclusion there isn't enough difference to matter.  Or there is and the benchmarks aren't consistently showing it.  I was initially getting statistically significant increases in gc time and overall cpu time of around 5% and 2%, but they have disappeared with my latest timings on a different machine.  C'est la benchmarking vie.

I'm interested in trying the simpler implementation I outlined in an earlier note, just to see what that costs, but I don't have time to run the experiment.  Maybe I will sometime after the current implementation has been pulled.

Here are some suggestions for the current implementation:

- Use FORCEINLINE (defined in segment.h, which is included into gc.c by the include of system.h) for check_trigger_ephemerons, lest some C compiler think the better of it.

- Declare all of the static functions on top, as is the convention.  I think the only ones missing right now are check_pending_ephemeron and check_trigger_ephemerons.  I don't know if FORCEINLINE will cause problems on the latter.

- Have sweep and sweep_space place ephemerons whose keys have not yet been reached directly onto the trigger ephemeron list of the key's segment and also onto the global trigger_ephemerons list.  (check_dirty_ephemeron already does this.)  Then eliminate the pending_ephemerons variable and uses of it.  (Perhaps I've missed something and this won't work or isn't a good idea.)

- Have clear_trigger_ephemerons identity ephemerons to be bwp'd by checking the key's space rather then by checking to see whether the trigger_next field is #t.

- Externally link ephemerons on the segment and global trigger lists, and eliminate both extra fields from the ephemeron representation.  This also obviates the special case for ephemerons in size_object.

- Have trigger_ephemerons be a list of lists of ephemerons rather than a simple list of ephemerons to avoid the list traversal in add_trigger_ephemerons_to_repending.

I can attempt any or all of these after doing the pull, if you prefer, but please update LOG, update the boot files, and revert patch.ss so I can do a clean pull, since I haven't figured out how not to take credit for the changes when doing a command-line pull.

mflatt:
I made the first two changes (the easy ones), rebased, and included a
commit for bootfiles. I can squash the commits, especially the
non-bootfile ones, if you'd prefer.

I didn't make the third change. I think the indirection through
`pending_ephemerons` makes the GC perform less work in the common case
--- but probably not enough to matter. As one example, when running the
Racket expander to load `racket/base`, ephemerons are used for weak
hashtables (as now built-in):
```
 ephemerons found         183,173
 dirty ephemerons found:    6,752

 ephemeron checks:        191,617 [i.e., how many times a key is checked]

 triggers installed:       81,434
 triggered lists:             192
 triggered ephemerons:      1,500 [i.e., ephemerons in those lists)
```
If I remove the indirection through `pending_ephemerons`:
```
 ephemerons found         183,220
 dirty ephemerons found:    6,576

 ephemeron checks:        311,903

 triggers installed:      202,554
 triggered lists:          41,077
 triggered ephemerons:     81,030
```
So, an ephemeron can be handled immediately about 1/2 of the time, and
it's rare for a key to become accessible only after following ephemeron
values. As a result, the indirection avoids almost all trigger work. If
adding an emphemeron to `pending_ephemerons` requires allocation, then
maybe it's worth a check for the half that can be handled right away,
but keeping the indirection for the rest.

Then again, these are all tiny numbers relative to the overall run:
```
    643 collections
    8.641620000s elapsed cpu time, including 1.805889000s collecting
    8.663434000s elapsed real time, including 1.811093000s collecting
    5438378944 bytes allocated, including 5347986480 bytes reclaimed
```
The gist stress test intentionally creates chains through ephemerons,
and then `pending_ephemerons` has no effect (as expected).

mflatt:
Forgot to revert "patch.ss", now done in the bootfiles commit.

dybvig:
<i>I didn't make the third change. I think the indirection through pending_ephemerons makes the GC perform less work in the common case ...</i>

Fair enough.  Is there a reason why `check_dirty_ephemeron` should not also put the ephemeron on `pending_ephemerons` rather than on the trigger lists?

mflatt:
My limited understanding of the context for `check_dirty_ephemeron` was that it needs to decide on a `youngest` result immediately. The more eager check lets it sometimes pick a more precise `youngest`.

mflatt:
Oh, if the eager check fails, the ephemeron could be put in the pending list instead of added as a trigger. So, yes, I think I made `check_dirty_ephemeron` too much like `check_ephemeron`.

dybvig:
That's what I was thinking.  I'm testing the change now and will commit it unless I run into a problem.

PR #169, opened 2017-04-22
more optimizations for map and for-each with explicit list

gus-massa:
Extend the reductions for `map` and `for-each` when the arguments are
explicit lists like `(list 1 2 3 ...)` or `'(1 2 3 ...)`.
.  cp0.ss 4.ms

This has many tweaks here and there, but they are too much interrelated to split this naturaly in a few smaller commits. 

--- The idea is to extend the reduction for expressions like

    (map + (list 1 2) (list 3 4))

to expressions like

    (map + (list 1 2) '(3 4))

I have to change the order of some optimizations. In expressions like the former, cp0 first try to inline the lists and if that failt then it tries to inline the implementation of `map` for short list with up to 4 items.

--- Also, when `map` tries to detect a short list, the optimization will be able to calculate the lenght of list like `(list 1 2 3)`, instead of only list like `'(1 2 3)`. This extend the reductions for expressions like

    (map + (begin (something) '(1 2 3)) another-list)

to expressions like

    (map + (begin (something) (list 1 2 3)) another-list)

--- I made similar changes to `for-each`. In particular for-each had a reduction for quoted short lists that is extended to list defined with list. 

--- And I added a reduction for expressions like

    (for-each something (list 1 2 3 ...) '(1 2 3 ...))

I just used the internal function `inline-lists` with a flag to signal if it is used in a map or a for-each. In a `for-each` it is not necesary to calculate all the `cdr` initially, but using the implementation that is used for `map` avoids creating the list, and in many times the values can be propagated.


dybvig:
Thanks.  These optimizations look sound and useful.  I suggest a few adjustments:

1. Add a length check for the `inline-lists` quoted-list case to avoid turning a small amount of code into code proportional to the size of a long list.  The code uses 4 as the maximum size for this purpose elsewhere.  You could relax this if a call to `list` is also present, though it's probably not worth the effort.

2. The safe-assert condition on line 3730 is guaranteed only if n isn't the result of a call to `list` without arguments, e.g., it isn't guaranteed if the list arguments to for-each are `(list)` and `e`.  Now, `(list)` should have been reduced to `(quote ())` by the cp0 handler for `list` before it arrives here, but since cp0 isn't guaranteed to reach a fixed point, we should count only on what the local cases guarantee.  A simple way around this would be to check for `(list)` as well as `null-rec?` in the first cond clause or to have `null-rec?` check for `(list)` with an appropriate comment about why we have it do so.

3. It would be nice to verify that side effects are preserved and aren't improperly interleaved when a constant list is involved by adding a mat similar to the one on line 1192 of 4.ms.  It looks like we have no such tests for `for-each`, so adding similar tests for it would be nice.

4. cp0.ss indentation: add a space at the front of lines 3595-3600 and 3602-3603, and remove the extra space at the front of line 3609.  You might as well remove the (preexisting) extra space at the front of lines 3624-3626.  (Actually, you can eliminate the `begin` wrapper and remove three spaces from the remaining lines.)

gus-massa:
4 - indentation: Fixed

2 - `safe-assert`/`(list)`: I added the code in `null-rec?`. I hope the comment is fine.

3 - more test: I added a few combinations with `map` applied to expressions with side effects with `(list …)` and `‘(…)`. 

I also added similar mats with `for-each`. I used a `lambda` with a side effect to ensure that the side effects of the arguments are run before the side effect of the lambda, and to avoid too smart optimizations that remove the whole application. I’m not testing the result of the `for-each`.

But I’m not sure if the application order in a `for-each` is guaranteed in Chez Scheme. For example, is this possible?

    (for-each display ‘(1 2 3 4)) ; ==> 2 1 4 3

One possible fix is to add in each mat all the possible results, so they will pass in possible any application order of the `lambda`. Another possibility is to remove the side effect. And a third possibility is to test both behaviors.

1 - check long quoted lists: I added a check to avoid the expansion when all list are quoted and they have more than 4 elements. I don’t like the idea of the 4 element threshold too much, because this transformation may enable more reductions than the other cases. In the other cases the code must still build the list so it’s not possible to propagate the values at compile time. Anyway, there are a few bad examples without the threshold like

    (let ([l ‘(1 2 3 ... 100)])
      (for-each display l)
      (for-each display l)
      (for-each display l)
      (for-each display l)
      (display l))

dybvig:
`for-each` does guarantee left-to-right order of evaluation---it's one of its primary charms.   `map`, however, does not.

It's appropriate that you do not check the `for-each` return value.  Most Chez Scheme primitives that return an unspecified value actually return void; `for-each` doesn't always, because it usually returns the value of the last call to maintain proper treatment of tail recursion.

gus-massa:
Do you think I have to fix something or you are planning to merge this as is? The tests already check the official behavior.

dybvig:
Since receiving clarification of the `for-each` left-to-right application requirement, have you checked to make sure the new optimizations preserve the order and the new tests verify it?  If so, I think we're ready to go, though I'll give it a once over before pulling.  Please remove the boot files from the pull request---I'll create and push them myself.

gus-massa:
I was sure that the optimization didn’t change the order of the applications. Anyway, I have read it again and double checked just in case. It is ok.

The tests assumed no reorder of the application too. (I asked because I was unsure if the tests were too strict.) They are ok too.

dybvig:
One of the mats was failing due to an order-of-evaluation problem, so I've reverted cp0.ss for now.  I see the problem and will patch it up soon.

gus-massa:
I looked at the fix and now I understand the error. I made an easy example to see it:

    (#3%for-each (lambda (x y) (display '!))
                 (list (display 'a) (display 'A))
                 (list (display 'b) (display 'B)))

Bad output: `Bba!A!`

Correct output: `aAbB!!`, `AabB!!`, `BbAa!!`, (8 posibilities)

PR #170, opened 2017-04-28
Add CC option to configure

mohd-akram:


akeep:
This change looks good, but I have a couple questions before I merge this change.

First, I noticed when zlib is compiled, it is still using the default system compiler before CC is not passed along.  I noticed we also do not send along other options to zlib, and maybe we should, but on the other hand zlib's use of autoconfig may already be sufficient, and I'm assuming in your testing you didn't run into any trouble.  So, the question is, do you want for the CC option to be passed into the zlib build process?

Second, I noticed you did not change the compile line in the mats Mf-* makefiles.  These files use 'cc' instead of 'gcc', so as long as there is a C compiler available at cc these can be used, however, it is possible the same C compiler should be used there.  That said, we've never used gcc specifically there, and we've not run into problems.  Did you try testing this out?

-andy:)

mohd-akram:
The motivation for this pull request was to upstream some [changes](https://github.com/macports/macports-ports/tree/master/lang/chez-scheme/files) that were required for MacPorts. There, I used the zlib provided by MacPorts through patching and perhaps this can be an option for ChezScheme as well (eg. `ZLIB_DIR`). Similarly, with nanopass I used `CHEZSCHEMELIBDIRS` to point to a MacPorts-provided nanopass.

Therefore, I agree that zlib should handle its own compilation. The makefiles in `mats` didn't seem to take any options (`Mf-config`) so I didn't modify them.

akeep:
Okay. sounds good.  I'll go ahead and merge the change.

PR #171, opened 2017-05-02
support Windows build on Bash/WSL

mflatt:
The future of Unix tools on Windows is likely Bash on Ubuntu on Windows (a.k.a "Bash/WSL"), so this patch adapts the configure script and makefiles to support it – at least for the latest version, where Windows executables can be run from Bash/WSL.

Just running `configure` should create a Linux build in Bash/WSL, but this patch makes `configure` set up a Windows build when the `OS` environment variable is set to `Windows_NT`.

Visual Studio is found and set up essentially as before, but backslashes in makefiles are treated differently in Cygwin and Bash/WSL, so some parts are moved to ".bat" files. The makefiles now use "cmd.exe" to launch batch files or otherwise switch to Windows-command world. Finally, Bash/WSL needs "scheme.exe" instead of just "scheme" to run the built executable.

### DIFF
--- a/BUILDING
+++ b/BUILDING
@@ -167,15 +167,18 @@ The make file supports several targets:
 
 WINDOWS
 
-Building Chez Scheme under Windows with Cygwin follows the
-instructions above except that make install is not supported:
+Building Chez Scheme under Windows with Cygwin or Bash/WSL follows the
+instructions above, except that 'make install' is not supported, and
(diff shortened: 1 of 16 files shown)


ProfessorX:
Commended. <ThumbsUp>

burgerrg:
Matthew, thanks for doing this!

PR #176, opened 2017-05-30
add bytevector-compress and bytevector-decompress

mflatt:
Not a high priority, but would be nice to have if it seems like a good idea: Expose zlib's compress() and uncompress() API as `bytevector-compress` and `bytevector-decompress`.

This is useful to Racket, because compiled code is written via `compile-to-port` to a bytevector. With this commit, the resulting bytevector can be compressed and decompressed without going through a file.

### DIFF
--- a/LOG
+++ b/LOG
@@ -523,3 +523,7 @@
     c/Mf-*
 - Suppress warnings from implicit fallthrough in case labels.
     Mf-{a6,arm32,i3,ppc,ta6,ti3,tpp32}le
+- added bytevector-compress and bytevector-uncompress
+    bytevector.ss, primdata.ss, new-io.c, prim5.c, externs.h,
+    objects.stex, release_notes.stex,
+    bytevector.ms, root-experr*
--- a/c/externs.h
+++ b/c/externs.h
@@ -197,6 +197,12 @@ extern ptr S_get_fd_length PROTO((ptr file, IBOOL gzflag));
 extern ptr S_set_fd_length PROTO((ptr file, ptr length, IBOOL gzflag));
 extern void S_new_io_init PROTO((void));
 
+extern uptr S_bytevector_compress_size PROTO((iptr s_count));
+extern ptr S_bytevector_compress PROTO((ptr dest_bv, iptr d_start, iptr d_count,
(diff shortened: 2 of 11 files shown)


dybvig:
The destination-length computation in `bytevector-compress` seems at best zlib-version-specific.  Can the `compressBound` function be used in place of the computation?  Did you consider recording the length with the compressed data so `bytevector-uncompress` need not take the length as an argument?  It seems like a reasonably small amount of overhead to include a 64-bit length at the front of the compressed bytevector.

mflatt:
I think adding a size to the start of the result of `bytevector-compress` is a good idea. I was originally trying to more directly reflect the underlying function, but having the size built-in would certainly work better in my one use case so far.

The `compressBound` function looks handy, and I'll switch to that. Oddly, the documentation that I found for `compress` gives the formula in the code instead of pointing to `compressBound` (which is in the same set of docs).

mflatt:
Updated as suggested.

dybvig:
Looks pretty good.   A few minor comments:

- It might be nice for `bytevector-uncompress` to verify that the embedded length is valid using `is_valid_zlib_length`.   I'm thinking of the case where a compressed bytevector is created on one type of machine and uncompressed on another with a smaller uLong size.  (I like that you've already hardcoded the size and endianness to facilitate cross-machine-type compression/decompression.)

- It might also be nice for `bytevector-uncompress` to complain if the length returned by `$bytevector-uncompress` isn't the same as the recorded uncompressed length, rather than truncating in the case where the length is shorter.

- It would be nice to get the source bytevector into the error messages / condition irritants list as an aid to debugging.

- Instead of defining `uncompressed-length-length` to be 8, it would be better to use `(ftype-sizeof integer-64)` or `(ftype-sizeof iptr)`.

- In the CSUG description of `bytevector-compress` I would stick the word `new` in the returns clause and remove the first paragraph of the description.  I would remove `, for example,` from the second paragraph.  In both new CSUG entries, the first `\var{bytevector}` in the returns clause should just be `bytevector`.

I can make these changes after pulling, if you prefer.

mflatt:
Ok - done!

PR #177, opened 2017-05-30
add file-position-object and eager source-line counting

mflatt:
This commit adds a way to store line and column information explicitly, instead of reconstructing it from a position by re-reading the file. Specifically a file-position object – which combines a position, line, and column – can be used in place of a position integer in functions like `make-source-object`.

The intent is to add line and column options without affecting existing code. It's not a completely backward-compatible change, because `source-object-bfp` and `source-object-efp` functions can now return file-position objects instead of integers, but they won't show up unless some new function is used.

When `locate-source` receives a file-position object in place of a position integer, then it doesn't try to consult a file in the filesystem, and just returns the file-position object's line and column. 

The changed include a new `record-source-lines-and-columns` that causes `compile-file` and other tools to record line and column information instead of just positions. That works by collecting position-to-line information at the same time as computing the file checksum, and the position-to-line table is kept as part of the source-file descriptor. The `make-source-object` function maps a position to a file-position object when the table is present. A `source-file-descriptor-lines-done!` function drops the table, and it's called automatically by `read-datum/annotations`. Maybe there's a better way to do all that. In fact, for my purposes, it's not so important to have something like `record-source-lines-and-columns` at all, but I wanted to try to make the new functionality generally available, and I do use it a little in my current Racket prototype.

Finally, inspectors for code and continuations support a new `'source-object` message to return the source object directly, instead of sending it through `locate-source`.


rain-1:
Thanks a lot! This is a feature I have been craving. I tried to build this but it seemed to get into an infinite loop.

mflatt:
@rain-1 You'll need to use a regular Chez Scheme build to bootstrap this one. Here are some instructions (based on instructs from Kent) that I've been meaning to turn into a PR as a patch `BUILDING`:

```
BOOSTRAPPING VIA CROSS COMPILATION

To build a change that is not compatible with existng Chez Scheme
builds, you will need to build in cross-compilation mode. Start with a
build from unadulterated sources in $ORIG on machine type $M. Then, in
the modified source tree:

  ./configure
  (cd $M/c && make)
  (cd $M/s && make allx)
```

One reason I've never gotten around to a PR is that it may be better to add a target to the main `Makefile`, and I'm undecided on that myself.

dybvig:
I'm tempted to make this fully backward-incompatible by having every source object record line and column information in addition to file-position information.  This would add overhead to the reader, which I don't think is a big deal since it doesn't tend to be a bottleneck on compilation, at least, and it would increase the size of every source object.  We could probably mitigate the latter with some careful encoding.  For example, we could combine bfp, beginning line, and beginning column in a single fixnum when they are small enough, and we could do the same for efp, ending line, and ending column except represent them as offsets from the bfp, starting line, and starting column to reduce the chance we'd have to resort to a larger encoding.  If this is too much of a compatibility hit we could make the line and column information optional.

dybvig:
_If this is too much of a compatibility hit we could make the line and column information optional._

I guess that's what you've already done. :)  Though I would be tempted to require line and column for both bfp and efp or for neither, have source-object-bfp and source-object-efp return simple positions, and add source-object-{b,e}{l,c} procedures that return beginning and ending line and column, all or none of which can be false.   Source objects with line and column information could be represented by a subtype of the existing source objects.

mflatt:
Any of those choices work for me, although I like least the choice of requiring both start and end to have lines and columns, because an ending line and column doesn't seem useful to me. The Racket API is position+line+column+span (and an ending position works well enough in place of a span).

If we don't have the requirement of keeping old things the same size, then I think adding a starting line and column to the `source` record type is the way to go. The current implementation causes my example ".so" files to grow by about 5% (when keeping inspector information), and I expect that adding two fields to `source-object` would be more compact.

Any of the choices is also an easy change, if you're comfortable with the extra argument to `make-source-file-descriptor` and the way `source-file-descriptor-lines-done!` works.

dybvig:
I've been looking over the source-location PR and have to admit `record-source-lines-and-columns` and all it entails aren't really to my liking.  In the interface, I don't like the rather indirect interaction of the `record-source-lines-and-columns` parameter, selected calls to `make-source-file-descriptor`, and `make-source object`, nor do I like the need to prune the table explicitly.  In the implementation, putting the table in the source-file descriptor (sfd) rather than, say, associating it with the sfd via a weak hashtable, is problematic because it's too easy for the table to get written along with the sfd to an object file, as will currently happen with `compile-file` and `compile-library`, which read, compile, and write the object file one expression at a time.  Also, line breaks can be determined properly in general only via a transcoded text port, so `make-source-file-descriptor` needs to do an additional pass (separate from the checksum computation) through the file's contents to create the table.

The existing (pre-PR) implementation has two major benefits, which are low read-time overhead and low storage overhead, requiring as it does only the maintenance and storage of the bfp and efp.  It has one major issue, which is large overhead when multiple calls to `locate-source` are made for the same file, since `locate-source` opens and scans the file twice on each call.  It also has a couple of minor issues: the file might not be available by the time `locate-source` is called, and the current transcoder might have changed.

Setting aside the `record-source-lines-and-columns` parameter and all it entails, your solution potentially addresses both the major and minor issues, since it allows programs such as the reader to compute the line and column information in the same pass as it computes the file positions.  On the other hand, doing so loses the major benefits of low read-time and storage overhead.

What would you think about having `locate-source` create and cache (via a weak hashtable) the position-to-line table instead as a way of avoiding multiple rereadings of the source file?  Because sfds might last indefinitely via pointers from compiled code, we could have each maximum-generation collection clear the cache of any tables that haven't been used since the last maximum-generation collection.  This would address the major issue without adding read-time or storage overhead, and it's a completely transparent change with no new primitives or incompatibilities.  It would not address the minor issues.

If you really need to be able to determine and store lines and columns at read time in some cases, e.g., you're reading something other than a file and can't use a similar cache, or you need to be able to report line and column information for files that aren't present at `locate-source` time, we could allow bfp and efp to be represented as integers or records, as the PR does, and also modify `locate-source` to create and cache the position-to-line tables to handle cases where the information isn't supplied.

In either case, I would leave out `record-source-lines-and-columns` and the related changes.

mflatt:
Here's another direction (not yet carefully tested) for your consideration:

 * Keeps file-location objects, so space is the same as before without lines and columns.

 * Gets rid of `record-source-lines-and-columns`, `source-file-descriptor-lines-done!`, and the tasteless and problematic state in source-file objects.

 * Adds `current-locate-source` and `current-make-syntax-object`, which the system uses in place of calling `locate-source` and `make-syntax-object` directly, and whose initial values are `locate-source` and `make-syntax-object`.

This implementation does not add caching for `locate-source`. I don't know how to do that and remain consistent with the idea that `locate-source` tells you whether the source file has changed.

The `current-locate-source` parameter lets a program disable the system's on-demand computation of lines and files. I find that control appealing, because I worry about system operations trolling through the filesystem in a final application. Meanwhile, `current-locate-source` could be set at compile time to cache files being compiled, or `current-make-syntax-object` could be set to compute line and column information eagerly while the reader constructs annotations.

dybvig:
Though I like the simplicity of the current mechanism, and I have some qualms about the potential performance hit to the reader, I can go along with allowing the optional use of file-position records in place of simple file positions in source objects as well as adding and using the current-make-source-object parameter (which I believe you mean rather than current-make-syntax-object).

I still think locate-source (actually $locate-source) should handle caching internally, and there are better ways to prevent trolling through the filesystem at run time, so I'd prefer to add caching to $locate-source and not add the current-locate-source parameter.  More on my rational in my responses to your comments below.

_This implementation does not add caching for locate-source. I don't know how to do that and remain consistent with the idea that locate-source tells you whether the source file has changed._

I don't share your concern with this.  There's already a window between when the file is opened and the two scans (to compute the CRC and locate the line and column) are complete, during which the file might change on disk.  If we prune the table fairly quickly, the window is not that much larger.  Also note the check is gone completely if the line and column are computed at compile time since by design the updated locate-source doesn't bother to find the file and check the CRC.

_I worry about system operations trolling through the filesystem in a final application._

You make it sound so nefarious. :)  Seriously, since this most commonly happens only when the application errors out without having set an appropriate exception handler, how worrisome is it?  Also, via current-make-source-object, one can ensure that all source-object positions are file position records, in which case $locate-source won't do the trolling anyway.  One can also limit trolling by setting source-directories and library-directories before calling anything that might end up in $locate-source.

mflatt:
On `locate-source` and caching: I had imagined that the current setup was designed to handle the case that you you compile a file, start debugging it (without doing enough work to trigger a GC), modify the file, and then use the inspector to view source information. If you're ok with the inspector reporting a bad line and column in that case, then it's fine with me.

On the filesystem and final applications: Maybe the issue here is "final application". I had in mind one that could use `eval`, which could trigger the compilation of a `define-values` form, which could trigger the use of `locate-source` (on a source that was recorded during the compilation of the application and through `syntax->datum` applied to the evaluated form). I could arrange for Racket to never include `define-values` in its output, but I worry that some other form that Racket uses may change to use `locate-source` in the future. Racket could also avoid source objects that have only positions, but since Racket allows source locations at its own layer to just have positions, it's clumsy at best to encode that in a file-position object. Our users have had trouble with the interaction of slow filesystems and unexpected filesystem operations, and we get bug reports when the distribution contains or accesses paths that it shouldn't. A parameter to disable filesystem access in `locate-source` would avoid complicated questions in a simple and reliable way.

mflatt:
The new commit adds caching in `locate-source`. I'm not sure the way I connected cache clearing to a maximum-generation GC is the way you had in mind.

Here's a program I tried for checking the performance effect of `current-make-source-object` on the reader: https://gist.github.com/mflatt/36b7f5f927b50460ea4369d83e5d56f2 . On my machine, that program runs in about 7.7 seconds with the current Chez Scheme and about 7.9 seconds with this PR – so a slowdown on the order of 3%. Uncommenting out the parameter adjustment (which means that every source-object constructor goes through an indirection with argument validation) makes it run in about 8.3 seconds, but that kind of penalty happens only when `current-make-source-object` is adjusted. I'm hoping that `read-datum` is generally used to read programs, where expansion and compilation time will be much more than reader time, so that 3% will be ok. Still, it's the least-important part of the PR for my purposes.

mflatt:
Well, there are test failures in "6.ms". I think the two tests fail because the new cache records a relative path for an `sfd` based on one search, and then the same relative path is used after the directory changes... so, more work there, but only if this is an acceptable direction overall.

mflatt:
The new commit makes the cache record state used to find a file, and it discards the cache entry if the state doesn't match. With that repair, the tests pass.

dybvig:
Having to ask for the current directory among the parameters to check if the cache is still valid is a big hit, depending on the O/S.  It makes me wonder if we aren't better off changing the interface to make caching explicit, perhaps along the lines you suggested earlier, or perhaps by changing the interface to locate-source.  For example, `make-source-locator` might take an sfd and return a source locator (i.e., a representation of the cache), and `locate-source` might optionally accept a source locator in place of an sfd.

mflatt:
If I understand the source-file locator suggestion, then I don't think it works. Annotations contain source objects that become arguments to `locate-source`. A source object shouldn't contain a source-file locator, because that would just take us back to the problem of having a line table in an annotation.

My reasoning on the potential cost of `current-directory` was that it's cheaper than reading a file and would at least resolve the current potential quadratic behavior. And there's the `current-locate-source` parameter if an application needs something different.

Another thought on the cost of `current-make-source-object`: Is it worth reserving a tc slot for the parameter, which I imagine would make lookup even faster? Or is the cost more likely in the indirection through `$make-source-object`?

dybvig:
I think it works, and it doesn't require source objects to contain a source-file locator, but it does push the responsibility of maintaining the cache onto the `locate-source` caller.  That was the effect I was going for, because it doesn't change the behavior of `locate-source`, and the caller should know better its tolerance for changes that might affect the search.  On the other hand, it is inconvenient, and a way to achieve a similar effect would be an optional _okay-to-cache?_ argument to `locate-source` to say it's okay to cache (unconditionally).  For most applications, this would be the desired behavior.

Incidentally, checking `current-directory` and the other parameters is insufficient as well as expensive, because it doesn't catch changes in the filesystem that effectively relocate source files.  `current-directory` can also fail due to lack of permissions or even fail to return due to cycles in the file system, though such situations are rare.

Although it's impossible to make the cache truly transparent due to various race conditions, we could come pretty close by recording the file modification time and having `$locate-source` go through the motions of locating the source file using the same algorithm as `$open-source-file` but stop short of opening the file and computing the checksum and instead compare the file modification time.   I prefer the optional _okay-to-cache?_ argument to locate-source, however.

dybvig:
Using a tc parameter for `current-make-source-object` might help, but the replacement of inline allocation with an anonymous procedure call is probably the bigger hit.

mflatt:
I'm still missing something. The particular case I have in mind is the call to `locate-source` in the expansion of `let-values`. I can see how `let-values` expansion could reasonably pass a true value of `ok-to-cache?`, but I don't see where a locator object would live for `let-values` to use.

dybvig:
Right.  That's where it gets inconvenient.  The `let-values` macro, or perhaps the expander operating on its behalf, would have to maintain a mapping from sfds to locator objects.  That's why I prefer the _okay-to-cache?_ flag.

mflatt:
The new commit adds a `use-cache?` argument to `locate-source`.

dybvig:
Thanks.  Is it necessary to represent the line explicitly in the position->line table?  That is, isn't the vector index the same as the line number, so can't the table just be a vector of positions, so that v[0] is the position of line 0, v[1] the position of line 1, etc.?

mflatt:
Good point! I'll adjust the table representation.

mflatt:
I squashed the commits to make the overall diff easier to see (and I can revert if squashing is unhelpful), but there's still a "patch.ss" change to revert.

dybvig:
I like what you did with `$make-source-object`.  Maybe we should take it a step further and combine it with your idea for using a tc parameter:

- new tc parameter: `$current-mso` initially #f
- `(current-make-source-object)` returns `(or ($current-mso) make-source-object)`
- `(current-make-source-object p)` sets `$current-mso` to `(if (eq? p make-source-object) #f p)` after verifying p is a procedure
- `$make-source-object` becomes a macro defined in read.ss that, if `($current-mso)` is #f, calls `%make-source`; otherwise calls `($current-mso)`

That should eliminate most of the read overhead when `current-make-source-object` has its default value.

mflatt:
Here's another try at the API. Instead of adding file-position objects, this version adds optional beginning-file-line and beginning-file-column components to a source object. This approach is more compatible with existing code, avoids a new "file-position object" concept, and simplifies source-object construction using a line and column. It doesn't support an ending line and column.

Internally, the variant with line and column components is a subtype of the one without. New `source-object-bfl` and `source-object-bfl` accessors each get a value of a new component, returning #f if the value is not present. [Possible issue: copying/adjusting a source object would be a little tedious, since `make-source-object` doesn't accept #f as a line or column.]

To help the compiler, etc., use the new optional components, the function bound to `current-locate-source` does not accept the same arguments as `locate-source`. [Possible issue: maybe `current-locate-source` should have a different name.] Instead, the parameter's function accepts a source object (not an sfd), a boolean for `start?` to indicate whether a start or end position is wanted, and a boolean for `cache?`. If `start?` is true and the source object has sfl and sfc components, the default function for `current-locate-source` returns those components with the sfd path without searching; otherwise, `locate-source` is used. This worked out, because all existing calls to `(current-locate-source)` have a source object on hand.

This commit also implements `$current-mso` as suggested.

The current commit is set up to simplify cross-compilation. There are several `UNPATCHME`s to clean up after bootstrapping.

dybvig:
I think I'd prefer to have `(current-locate-source)` take exactly the same arguments as `locate-source` and to push the logic that looks for the existence of the source-object start line+column into the caller.  If that doesn't seem reasonable, I agree `current-locate-source` is misnamed, and there ought to be a name for its default value that one can call directly or use to reset the parameter.

mflatt:
I added `locate-source-object-source` and renamed the parameter to `current-locate-source-object-source`, although those names seem long.

Having the parameter's procedure take a source object is intended to allow control over the way that recorded information and the filesystem are used, since there are many reasonable choices that depend on the application. For example, an application might want to use a recorded line and column but still perform a search for the file path.

dybvig:
Here are some possible shorter alternatives to locate-source-object-source:

- source-object-location
- source-location
- determine-source-location
- determine-location

I would be tempted to leave out the use-cache? argument and have the default procedure pass locate-source #t for the argument.

mflatt:
Leaving out the `use-cache?` argument would mean that the inspector uses the cache, too, which would be a change; some tests in "6.ms" (around line 2600) would fail, I think. In that case, the tests would have to change, or the inspector would have to avoid `current-locate-source-object-source`. I recommend keeping `use-cache?`, but the behavior of that corner doesn't matter to me, so I'll go with whatever you prefer.

I don't have strong opinions on the name for [`current-`]`locate-source-object-source`. FWIW, `current-source-`[`object-`]`location` seems misleading as a replacement for `current-locate-source-object-source` (since it sounds like a parameter that holds a source location, instead of a function to find a location), and I like keeping `locate` in the name to connect it to `locate-source`. But I'd be happy to change to whatever name seems best to you.

dybvig:
Agreed.   I'm okay with the current names.

dybvig:
I have reviewed the latest code changes in detail, and I have a few final suggestions.

(1) The `current-mso` thread-context field needs to be relocated by `sweep_thread` (gc.c), so add:
```
relocate(&CURRENTMSO(tc))
```
after
```
relocate(&SFD(tc))
```

(2) When called from `$display-condition` rather than `$make-source-oops`, `$display-condition` should probably pass `(current-locate-source-object-source)` `#f` for the `use-cache?` argument.  The `$display-condition` prefix? argument already distinguishes the calls from `display-condition` and `$make-source-oops`, though before it is also used to determine the value of `use-cache?`, it should probably be renamed.

(3) I would change `source-object-bfl` and `source-object-bfc` to `source-object-line` and `source-object-col` (or `source-object-column`).  I don't think anyone would be surprised that these are the beginning line and column.

(4) I would rewrite the loop in the read.ss definition of `binary-search` from:
```
(let loop ([lo 0] [hi (vector-length table)])
  (let* ([mid (fxsra (fx+ lo hi) 1)]
         [pos (vector-ref table mid)])
    (cond
      [(fx= (fx+ 1 lo) hi)
       (values name
               hi
               (fx+ 1 (fx- fp pos)))]
      [(< fp pos) (loop lo mid)]
      [else (loop mid hi)])))
```
to
```
(let loop ([lo 0] [hi (vector-length table)])
  (if (fx= (fx+ 1 lo) hi)
      (values name
              hi
              (fx+ 1 (fx- fp (vector-ref table lo))))
      (let ([mid (fxsra (fx+ lo hi) 1)])
        (if (< fp (vector-ref table mid))
            (loop lo mid)
            (loop mid hi)))))
```
both to avoid the computation of mid in the base case and to avoid implicit reliance on the fact that mid is the same as lo in that case.

(5) The optional third argument to `$source-file-descriptor` in the syntax.ss definition of `include` can be omitted, since `#t` is the default value, and this an unnecessary change to the code.

mflatt:
Thanks for the repairs and other improvements! I think the latest commit gets them all. For (2), I opted for an additional argument, even though the new argument and `prefix?` are correlated.

If it's otherwise ok, let me know if you'd like the commits squashed and/or bootfiles added with `UNPATCHME`s removed.

dybvig:
I have just one fix to the fixes:  the semicolon after `relocate(&CURRENTMSO(tc))` is unnecessary, if harmless.

Otherwise, this looks good, so please go ahead and squash, create bootfiles, and remove the patches.

mflatt:
Done

Issue #178, opened 2017-05-31
Errors when building on FreeBSD


krrrcks:
I followed #7 and produced the boot files for 'a6fb' on my 64bit Linux machine. After that I copied them to my FreeBSD machine which is "FreeBSD 11.0-RELEASE-p8" running as a VM. After `configure` I started the build process by `gmake` and when compiling `prim5.c` I got these error messages:

```
gcc  -Wpointer-arith -Wextra -Werror -O  -c -DX86_64 -I../boot/a6fb -I../zlib -I/usr/local/include -I/usr/X11R6/include prim5.c
prim5.c: In function 's_iconv_from_string':
prim5.c:1984:22: error: passing argument 2 of 'libiconv' from incompatible pointer type [-Werror=incompatible-pointer-types]
   ICONV((iconv_t)cd, (ICONV_INBUF_TYPE)&inbuf, &inbytesleft, &outbuf, &outbytesleft);
                      ^
In file included from prim5.c:1897:0:
/usr/local/include/iconv.h:85:15: note: expected 'char **' but argument is of type 'const char **'
 extern size_t iconv (iconv_t cd,  char* * inbuf, size_t *inbytesleft, char* * outbuf, size_t *outbytesleft);
               ^
prim5.c: In function 's_iconv_to_string':
prim5.c:2015:22: error: passing argument 2 of 'libiconv' from incompatible pointer type [-Werror=incompatible-pointer-types]
   ICONV((iconv_t)cd, (ICONV_INBUF_TYPE)&inbuf, &inbytesleft, &outbuf, &outbytesleft);
                      ^
In file included from prim5.c:1897:0:
/usr/local/include/iconv.h:85:15: note: expected 'char **' but argument is of type 'const char **'
 extern size_t iconv (iconv_t cd,  char* * inbuf, size_t *inbytesleft, char* * outbuf, size_t *outbytesleft);
               ^
cc1: all warnings being treated as errors
gmake[2]: *** [Makefile:30: prim5.o] Error 1
gmake[1]: *** [Makefile:20: build] Error 2
gmake: *** [Makefile:19: build] Error 2
```
This seams to be different to #143, but as @gwatt mentioned in https://github.com/cisco/ChezScheme/issues/143#issuecomment-284274115 he faced the same problem when building on FreeBSD.

As I am not a C programmer I have no clue on how to fix it or provide a PR that smooths these things.

bnoordhuis:
[...]

PR #180, opened 2017-06-02
Fix for building on FreeBSD (issue #178)

krrrcks:


akeep:
Thanks for the update.  It looks like we've got some work to do on the unit tests on FreeBSD still.  I did notice, it looks like the libc.so is now libc.so.7, I should have a simple patch for that ready soon.

PR #181, opened 2017-06-06
Fix fxbit-set? signature

gus-massa:
The return value is marked as `fixnum`, but it should be `boolean`.

(I'm not sure if it is necesary to recalculate the bootfiles.)


dybvig:
Good change, thanks.  Generally speaking, it's not necessary to commit new boot files if the changes don't affect compilation.  For example, a change in an error message or a change in the definition of a primitive in prims.ss would not require new boot files.  A change to primdata.ss would typically affect compilation, but the output type is not presently used for anything, if I'm not mistaken.

PR #183, opened 2017-06-14
Disable warning for implicit fallthrough

gwatt:
gcc 7 has a new warning: `-Wimplicit-fallthrough`. When coupled with `-Wextra -Werror`, this warning breaks the build.

### DIFF
--- a/BUILDING
+++ b/BUILDING
@@ -167,15 +167,18 @@ The make file supports several targets:
 
 WINDOWS
 
-Building Chez Scheme under Windows with Cygwin follows the
-instructions above except that make install is not supported:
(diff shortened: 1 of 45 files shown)


dybvig:
This looks good to me, though with the pull of #170 the make files are out of date, and my attempt to fix the conflicts lost the record of your commits.  Would you please update the files so I can do a clean pull?

gwatt:
@dybvig done.
I dropped the changes to non-linux makefiles. As this warning is only in gcc7, and pre 4.4 gcc will also issue a warning for unknown flags, I decided to not include it for all architectures that might possible use gcc.

PR #184, opened 2017-06-17
Drop the arguments of void in cp0

gus-massa:
This reduces 

    (#3%void x (vector 1 2) (newline) y) ==> (begin (newline) (void))

This can be useful to mark `x` and `y` as unused and avoid allocating the `vector`.

I was not sure if it’s necessary to add tests, and in which file to put them.

I enabled it only for the optimization level `3` because otherwise it would reduce

    (#2%void (values 1 2)) ==> (void)

### DIFF
--- a/LOG
+++ b/LOG
@@ -514,3 +514,6 @@
 - support Windows build on Bash/WSL
     BUILDING, configure, workarea, c/vs.bat (new), mats/vs.bat (new),
     c/Mf-*nt, mats/Mf-*, s/Mf-base
+- drop the arguments of void in cp0
+    cp0.ss
+    primdata.ss
--- a/s/cp0.ss
+++ b/s/cp0.ss
@@ -1943,6 +1943,11 @@
                      (residualize-seq '() (list who e) ctxt)
                      true-rec)]))
 
+      (define-inline 3 (void)
+        [args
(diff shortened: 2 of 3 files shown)


dybvig:
Thanks for the pull request, but `void` doesn't take any arguments, so it doesn't make sense for the optimizer to discard them.  Are you proposing that `void` should be modified generally to accept any number of arguments and ignore them?  If so, more would need to be changed than cp0, and before additional changes are made, I'd like to know why it's desirable.  `values` can already be used for this purpose in many cases, and of course there's always `begin` or `(lambda r (void))`.

Incidentally, the result of an argument expression evaluating to other than one value is unspecified, and the compiler often chooses not to verify that an argument expression evaluates to exactly one value when no further processing of the value is required, e.g., when the value is unused or when the entire call evaluates to the value, as in `(cons* (values 2 3))`, so if the optimization made sense at optimization level 3, it would make sense at other optimization levels.

gus-massa:
I was confused because Racket has an extended version of `void` that just ignores its arguments. It's sometime handy in the RELP to ignore the result of an expression.

    (void (time (+ 1 2 3)))

Anyway, it's not very incredible useful, so I think it's better to make no change.

PR #188, opened 2017-06-29
Don't mark pariah as `simple?`

gus-massa:
I'm not sure about this, and I'm not sure if this is the correct fix ...

If I run 

    (expand/optimize 
     '(if (zero? (random 37))
        (pariah (display "green"))
        (void)))

I get 

    (if (#2%zero? (#2%random 37))
        (#2%display "green")
        (#2%void))

but I expect something like

    (if (#2%zero? (#2%random 37))
        (begin
          (pariah (void))
          (#2%display "green"))
        (#2%void))

The problem is that `(pariah)` is marked as `simple?` and it's removed from the `begin` sequence.

This change breaks the reduction of

    (expand/optimize 
     '(if (zero? (random 37))
       (pariah (void))
       (void)))

### DIFF
--- a/LOG
+++ b/LOG
@@ -529,3 +529,6 @@
     bytevector.ms, root-experr*
 - fixed typo in S_abnormal_exit
     schsig.c
+- don't remove the pariah form in the cp0 pass
+    cp0.ss,
+    misc.ms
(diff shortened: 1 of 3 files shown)


dybvig:
This change makes sense to me; otherwise, the `pariah` form is worthless unless cp0 is disabled.  Please fix any mats that break as a result of the change, if any, and (particularly if there aren't any that break) add new tests that verify the change has the desired effect.  Please also add a LOG entry.  Don't bother rebuilding boot files, however.

gus-massa:
Done. I added a test.

PR #191, opened 2017-07-16
fix `(define-values () ....)` to expand to a definition

mflatt:
Expanding to a definition means that it doesn't interrupt a definition
sequence.


dybvig:
Looks good, thanks.

PR #192, opened 2017-07-17
Add a compilation pass to cp0 to reduce expression using types

gus-massa:
cc: @mflatt 

This is still work in progress, so the files still have some unnecessary comments and rough parts, but it's already big and I think I need some feedback before continuing. 

In general, this changes are useful with optimization-level 2, because most of the type checks are skipped in level 3.

The idea was to reduce most common idioms, for example:

    (lambda (x) (vector-set! x 0 0) (vector? x)) 
      ==> (lambda (x) (vector-set! x 0 0) #t)

    (lambda (x) (when (vector? x) (vector? x))) 
      ==> (lambda (x) (when (vector? x) #t))

    (lambda (x) (when (vector? x) (box? x))) 
      ==> (lambda (x) (when (vector? x) #f))

    (lambda (x) (unless (vector? x) (error 'e "")) (vector? x))
      ==> (lambda (x) (unless (vector? x) (error 'e "")) #t)

More examples of the reductions are in the first part of cptypes.ms

The main part of the code is in cptypes.ss. It has an explanation of the parameters of the main function at the beginning of the file.

(Some auxiliary functions are copied from cp0.ss.  There is also an hamt implementation in hamt.ss. It's necessary to get fast lookup of the discovered types.)

----

The first commit is just to recalculate automatically the bootfiles in travis-ci. It's handy for testing while writing the initial versions, but my idea is to remove/revert it in the final version. 

The second commit adds the signature information available in primdata.ss to the `primref`. so the other parts of the code can use it. This step is tricky, so I left it in an isolated commit, but my plan is to squash it in the final version. (The third commit has the bootfiles.)

The last commit is the main part of the implementation.

### DIFF
--- a/mats/4.ms
+++ b/mats/4.ms
@@ -1304,9 +1304,9 @@
               (list 'b 'u 'y)
               (list 'c 'v 'z))))
     '(#2%list
-       (#2%string->symbol (#2%string-append "a" "b" "c"))
-       (#2%string->symbol (#2%string-append "t" "u" "v"))
-       (#2%string->symbol (#2%string-append "x" "y" "z"))))
+       (#3%string->symbol (#3%string-append "a" "b" "c"))
+       (#3%string->symbol (#3%string-append "t" "u" "v"))
+       (#3%string->symbol (#3%string-append "x" "y" "z"))))
   (equivalent-expansion?
     (parameterize ([optimize-level 3] [enable-cp0 #t] [#%$suppress-primitive-inlining #f])
       (expand/optimize
--- a/mats/Mf-base
+++ b/mats/Mf-base
@@ -125,7 +125,7 @@ ecpf = $(defaultecpf)
 
 # set of mats to run
 mats = primvars 3 4 5_1 5_2 5_3 5_4 5_5 bytevector thread\
- misc cp0 5_6 5_7 5_8 6 io format 7 record hash enum 8 fx fl cfl foreign\
+ misc cp0 cptypes 5_6 5_7 5_8 6 io format 7 record hash enum 8 fx fl cfl foreign\
  ftype unix windows examples ieee date exceptions oop
 
 Examples = ../examples
--- a/mats/cp0.ms
+++ b/mats/cp0.ms
@@ -815,13 +815,13 @@
            (lambda (r)
              (emit-word! 2953052161)
              (emit-word! 3766812992)
-             (emit-word! (#2%+ 3766747136 (#2%ash r 0))))))
+             (emit-word! (#3%+ 3766747136 (#2%ash r 0))))))
       (syntax-case x ($primitive)
         [(set! test
            (lambda (r1)
              (ew1! 2953052161)
              (ew2! 3766812992)
-             (ew3! (#2%+ 3766747136 (#2%ash r2 0)))))
+             (ew3! (#3%+ 3766747136 (#2%ash r2 0)))))
          (eq? #'r1 #'r2)])))
  ; verify optimization of (if e s s) => (begin e s)
   (equivalent-expansion?
@@ -859,14 +859,14 @@
       (expand/optimize
         '(lambda (x y) (if (not (or (fx< x y) (fx> y x))) x y))))
     '(lambda (x.0 y.1)
-       (if (if (#2%fx< x.0 y.1) #t (#2%fx> y.1 x.0))
+       (if (if (#2%fx< x.0 y.1) #t (#3%fx> y.1 x.0))
            y.1
            x.0)))
   (equivalent-expansion?
     (parameterize ([#%$suppress-primitive-inlining #f] [run-cp0 (lambda (cp0 x) (cp0 x))] [optimize-level 2])
       (expand/optimize
         '(lambda (x y) (if (or (fx< x y) (fx> y x)) x y))))
-    '(lambda (x y) (if (if (#2%fx< x y) #t (#2%fx> y x)) x y)))
+    '(lambda (x y) (if (if (#2%fx< x y) #t (#3%fx> y x)) x y)))
   (equivalent-expansion?
     (parameterize ([#%$suppress-primitive-inlining #f] [run-cp0 (lambda (cp0 x) (cp0 x))] [optimize-level 2])
       (expand/optimize
--- a/mats/cptypes.ms
+++ b/mats/cptypes.ms
@@ -1,0 +1,518 @@
+;;; cptypes.ms
+;;; Copyright 1984-2017 Cisco Systems, Inc.
+;;; 
+;;; Licensed under the Apache License, Version 2.0 (the "License");
+;;; you may not use this file except in compliance with the License.
+;;; You may obtain a copy of the License at
+;;; 
+;;; http://www.apache.org/licenses/LICENSE-2.0
+;;; 
+;;; Unless required by applicable law or agreed to in writing, software
+;;; distributed under the License is distributed on an "AS IS" BASIS,
+;;; WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+;;; See the License for the specific language governing permissions and
+;;; limitations under the License.
+
+(define-syntax cptypes-equivalent-expansion?
+  (syntax-rules ()
+    [(_ x y)
+     (equivalent-expansion?
+      (parameterize ([enable-cp0 #t]
+                     [#%$suppress-primitive-inlining #f]
+                     #;[optimize-level (max (optimize-level) 2)])
+        (expand/optimize x))
+      (parameterize ([enable-cp0 #t]
+                     [#%$suppress-primitive-inlining #f]
+                     #;[optimize-level (max (optimize-level) 2)])
+        (expand/optimize y)))]))
+
+(mat cptypes-handcoded
+  (cptypes-equivalent-expansion?
+    '(vector? (vector))  ;actually reduced by folding, not cptypes 
+    #t)
+  (cptypes-equivalent-expansion?
+    '(vector? (vector 1 2 3)) 
+    #t)
+  (cptypes-equivalent-expansion?
+    '(vector? (box 1)) 
+    #f)
+  (cptypes-equivalent-expansion?
+    '(box? (vector 1 2 3)) 
+    #f)
+  (cptypes-equivalent-expansion?
+    '(box? (box 1)) 
+    #t)
+  (cptypes-equivalent-expansion?
+    '(pair? (cons 1 2)) 
+    #t)
+  (cptypes-equivalent-expansion?
+    '(lambda (x) (vector-set! x 0 0) (vector? x))
+    '(lambda (x) (vector-set! x 0 0) #t))
+  (cptypes-equivalent-expansion?
+    '(lambda (x) (vector-set! x 0 0) (box? x))
+    '(lambda (x) (vector-set! x 0 0) #f))
+  (cptypes-equivalent-expansion?
+    '(lambda (x y) (vector-set! x 0 0) (set! y (vector? x)))
+    '(lambda (x y) (vector-set! x 0 0) (set! y #t)))
+  (cptypes-equivalent-expansion?
+    '(lambda (x y) (set! y (vector-ref x 0)) (list (vector? x) y))
+    '(lambda (x y) (set! y (vector-ref x 0)) (list #t y)))
+  (cptypes-equivalent-expansion?
+    '(lambda (x) (vector-set! x 0 0) (let ([y (random 7)]) (list (vector? x) y y)))
+    '(lambda (x) (vector-set! x 0 0) (let ([y (random 7)]) (list #t y y))))
+  (cptypes-equivalent-expansion?
+    '(lambda (x) (vector-set! x 0 0) (let ([y (vector? x)]) (list (random 7) y y)))
+    '(lambda (x) (vector-set! x 0 0) (let ([y #t]) (list (random 7) y y))))
+  (cptypes-equivalent-expansion?
+    '(lambda (x) (let ([y (vector-ref x 0)]) (list (vector? x) y y)))
+    '(lambda (x) (let ([y (vector-ref x 0)]) (list #t y y))))
+  (cptypes-equivalent-expansion?
+    '(lambda (x) (let ([y (vector-ref x 0)])
+                   (let ([z (vector? x)])
+                     (list y y z z))))
+    '(lambda (x) (let ([y (vector-ref x 0)])
+                   (let ([z #t])
+                     (list y y z z)))))
+  (cptypes-equivalent-expansion?
+    '(lambda (x) (let ([y (vector-ref x 0)]) (display (list y y))) (vector? x))
+    '(lambda (x) (let ([y (vector-ref x 0)]) (display (list y y))) #t))
+  (cptypes-equivalent-expansion?
+    '(lambda (x) (let ([y (random 7)]) (display (list (vector-ref x 0) y y))) (vector? x))
+    '(lambda (x) (let ([y (random 7)]) (display (list (vector-ref x 0) y y))) #t))
+  (cptypes-equivalent-expansion?
+    '(let ([y (vector 1 2 3)]) (display (list (vector? y) y y)))
+    '(let ([y (vector 1 2 3)]) (display (list #t y y))))
+  (cptypes-equivalent-expansion?
+    '(let ([y (vector 1 2 3)]) (display (list y y)) (vector? y))
+    '(let ([y (vector 1 2 3)]) (display (list y y)) #t))
+  (cptypes-equivalent-expansion?
+    '(vector? (let ([y (vector 1 2 3)]) (display (list y y)) y))
+    '(begin (let ([y (vector 1 2 3)]) (display (list y y)) y) #t))
+  (cptypes-equivalent-expansion?
+    '(vector? (let ([y (vector 1 2 3)]) (display (list y y)) (vector 4 5 6)))
+    '(begin (let ([y (vector 1 2 3)]) (display (list y y)) (vector 4 5 6)) #t))
+  (cptypes-equivalent-expansion?
+    '(lambda (x) (when (null? x) (display x)))
+    '(lambda (x) (when (null? x) (display '()))))
+  (cptypes-equivalent-expansion?
+    '(lambda (x) (when (vector? x) (eq? x 'vector?)))
+    '(lambda (x) (when (vector? x) #f)))
+  (cptypes-equivalent-expansion?
+    '(lambda (x) (when (vector? x) (pair? x)))
+    '(lambda (x) (when (vector? x) #f)))
+  (cptypes-equivalent-expansion?
+    '(lambda (x) (when (vector? x) (vector? x)))
+    '(lambda (x) (when (vector? x) #t)))
+  (cptypes-equivalent-expansion?
+    '(lambda (x) (when (procedure? x) (procedure? x)))
+    '(lambda (x) (when (procedure? x) #t)))
+  (cptypes-equivalent-expansion?
+    '(lambda (f) (f) (procedure? f))
+    '(lambda (f) (f) #t))
+  (cptypes-equivalent-expansion?
+    '(lambda (x)
+       (vector-set! x 0 0)
+       (let loop ([n 1000])
+         (unless (zero? n)
+           (display (vector? x))
+           (loop (- n 1)))))
+    '(lambda (x)
+       (vector-set! x 0 0)
+       (let loop ([n 1000])
+         (unless (zero? n)
+           (display #t)
+           (loop (- n 1))))))
+  (cptypes-equivalent-expansion?
+    '(lambda (x)
+       (let loop ([n 1000])
+         (unless (zero? n)
+           (vector-set! x 0 n)
+           (loop (- n 1))))
+       (vector? x))
+    '(lambda (x)
+       (let loop ([n 1000])
+         (unless (zero? n)
+           (vector-set! x 0 n)
+           (loop (- n 1))))
+       (vector? x)))
+  (cptypes-equivalent-expansion?
+    '(begin (error 'who "msg") 1) ;could be reduced in cp0
+    '(begin (error 'who "msg") 2))
+  (cptypes-equivalent-expansion?
+    '(lambda (x) (vector-set! x) 1)
+    '(lambda (x) (vector-set! x) 2))
+  (cptypes-equivalent-expansion?
+    '(lambda (x) (#2%-) 1)
+    '(lambda (x) (#2%-) 2))
+  (cptypes-equivalent-expansion?
+    '(lambda (x) (#2%make-vector x 0 7) 1)
+    '(lambda (x) (#2%make-vector x 0 7) 2))
+  (cptypes-equivalent-expansion?
+    '(lambda (x) (vector-set! x 0 0) (set-box! x 0) 1)
+    '(lambda (x) (vector-set! x 0 0) (set-box! x 0) 2))
+  (cptypes-equivalent-expansion?
+    '(lambda (x) (vector-set! (box 5) 0 0) 1)
+    '(lambda (x) (vector-set! (box 5) 0 0) 2))
+  (cptypes-equivalent-expansion?
+    '(lambda (x) (#2%odd? x) (real? x))
+    '(lambda (x) (#2%odd? x) #t))
+  (cptypes-equivalent-expansion?
+    '(lambda (x) (vector-set! x 0 0) (#2%odd? x) 1)
+    '(lambda (x) (vector-set! x 0 0) (#2%odd? x) 2))
+)
+
+(mat cptypes-type-if
+  (cptypes-equivalent-expansion?
+    '(lambda (x) (if (vector-ref x 0) (newline) (void)) (vector? x))
+    '(lambda (x) (if (vector-ref x 0) (newline) (void)) #t))
+  (cptypes-equivalent-expansion?
+    '(lambda (x) (if (vector-ref x 0) (vector? x) (void)))
+    '(lambda (x) (if (vector-ref x 0) #t (void))))
+  (cptypes-equivalent-expansion?
+    '(lambda (x) (if (vector-ref x 0) (void) (vector? x)))
+    '(lambda (x) (if (vector-ref x 0) (void) #t)))
+  (cptypes-equivalent-expansion?
+    '(lambda (x) (if (zero? (vector-ref x 0)) (newline) (void)) (vector? x))
+    '(lambda (x) (if (zero? (vector-ref x 0)) (newline) (void)) #t))
+  (not (cptypes-equivalent-expansion?
+         '(lambda (x) (if (zero? (random 2)) (vector-set! x 0 0) (void)) (vector? x))
+         '(lambda (x) (if (zero? (random 2)) (vector-set! x 0 0) (void)) #t)))
+  (not (cptypes-equivalent-expansion?
+         '(lambda (x) (if (zero? (random 2)) (void) (vector-set! x 0 0)) (vector? x))
+         '(lambda (x) (if (zero? (random 2)) (void) (vector-set! x 0 0)) #t)))
+  (cptypes-equivalent-expansion?
+    '(lambda (x) (vector-set! x 0 0) (if x (newline) (void)))
+    '(lambda (x) (vector-set! x 0 0) (if #t (newline) (void))))
+  (cptypes-equivalent-expansion?
+    '(lambda (x) (vector-set! x 0 0) (if (vector? x) (newline) (void)))
+    '(lambda (x) (vector-set! x 0 0) (if #t (newline) (void))))
+  (cptypes-equivalent-expansion?
+    '(lambda (x) (when (vector? x) (if x (newline) (void))))
+    '(lambda (x) (when (vector? x) (if #t (newline) (void)))))
+  (not (cptypes-equivalent-expansion?
+         '(lambda (x) (when (boolean? x) (if x (newline) (void))))
+         '(lambda (x) (when (boolean? x) (if #t (newline) (void))))))
+  (cptypes-equivalent-expansion?
+    '(lambda (x) (vector-set! x 0 0) (if (zero? (random 2)) (vector? x) (void)))
+    '(lambda (x) (vector-set! x 0 0) (if (zero? (random 2)) #t (void))))
+  (cptypes-equivalent-expansion?
+    '(lambda (x) (vector-set! x 0 0) (if (zero? (random 2)) (void) (vector? x)))
+    '(lambda (x) (vector-set! x 0 0) (if (zero? (random 2)) (void) #t)))
+  (cptypes-equivalent-expansion?
+    '(lambda (x) (if (vector? x) (vector? x) (void)))
+    '(lambda (x) (if (vector? x) #t (void))))
+  (not (cptypes-equivalent-expansion?
+         '(lambda (x) (if (vector? x) (void) (vector? x)))
+         '(lambda (x) (if (vector? x) (void) #t))))
+  (cptypes-equivalent-expansion?
+    '(lambda (x y) (if (vector? x) (if (vector? y) (list (vector? x) (vector? y)) (void)) (void)))
+    '(lambda (x y) (if (vector? x) (if (vector? y) (list #t #t) (void)) (void))))
+  (cptypes-equivalent-expansion?
+    '(lambda (x y) (if (and (vector? x) (vector? y)) (list (vector? x) (vector? y)) (void)))
+    '(lambda (x y) (if (and (vector? x) (vector? y)) (list #t #t) (void))))
+  (not (cptypes-equivalent-expansion?
+         '(lambda (x y) (if (or (vector? x) (vector? y)) (vector? x) (void)))
+         '(lambda (x y) (if (or (vector? x) (vector? y)) #t (void)))))
+  (not (cptypes-equivalent-expansion?
+         '(lambda (x y) (if (or (vector? x) (vector? y)) (vector? y) (void)))
+         '(lambda (x y) (if (or (vector? x) (vector? y)) #t (void)))))
+  (cptypes-equivalent-expansion?
+    '(lambda (x y) (if (if (vector? x) (vector? y) #f) (list (vector? x) (vector? y)) (void)))
+    '(lambda (x y) (if (if (vector? x) (vector? y) #f) (list #t #t) (void))))
+  (cptypes-equivalent-expansion?
+    '(lambda (x y) (if (if (vector? x) (vector? y) #t) (void) (vector? x)))
+    '(lambda (x y) (if (if (vector? x) (vector? y) #t) (void) #t)))
+  (cptypes-equivalent-expansion?
+    '(lambda (t) (let ([x (if t (begin (newline) #f) #f)]) (number? x)))
+    '(lambda (t) (let ([x (if t (begin (newline) #f) #f)]) #f)))
+  (cptypes-equivalent-expansion?
+    '(lambda (t) (let ([x (if t 1 2)]) (fixnum? x)))
+    '(lambda (t) (let ([x (if t 1 2)]) #t)))
+  (cptypes-equivalent-expansion?
+    '(lambda (t) (let ([x (if t 1 2.0)]) (number? x)))
+    '(lambda (t) (let ([x (if t 1 2.0)]) #t)))
+  (cptypes-equivalent-expansion?
+    '(if (error 'who "msg") (display 1) (display 2))
+    '(if (error 'who "msg") (display -1) (display -2)))
+  (cptypes-equivalent-expansion?
+    '(begin (if (error 'who "msg") (display 1) (display 2)) (display 3))
+    '(begin (if (error 'who "msg") (display 1) (display 2)) (display -3)))
+  (cptypes-equivalent-expansion?
+    '(begin (if (box? (box 0)) (error 'who "msg") (void)) (display 1))
+    '(begin (if (box? (box 0)) (error 'who "msg") (void)) (display -1)))
+  (not (cptypes-equivalent-expansion?
+         '(begin (if (box? (box 0)) (void) (error 'who "msg")) (display 1))
+         '(begin (if (box? (box 0)) (void) (error 'who "msg")) (display -1))))
+  (cptypes-equivalent-expansion?
+    '(lambda (x) (if (zero? (random 2)) (vector-set! x 0 0) (error 'who "msg")) (vector? x))
+    '(lambda (x) (if (zero? (random 2)) (vector-set! x 0 0) (error 'who "msg")) #t))
+  (cptypes-equivalent-expansion?
+    '(lambda (x) (if (zero? (random 2)) (error 'who "msg") (vector-set! x 0 0)) (vector? x))
+    '(lambda (x) (if (zero? (random 2)) (error 'who "msg") (vector-set! x 0 0)) #t))
+  (cptypes-equivalent-expansion?
+    '(begin (if (zero? (random 2)) (error 'who "msg") (error 'who "other")) (display 1))
+    '(begin (if (zero? (random 2)) (error 'who "msg") (error 'who "other")) (display -1)))
+  (cptypes-equivalent-expansion?
+    '(lambda (x y) (if y (vector-set! x 0 0) (vector-set! x 0 1)) (vector? x))
+    '(lambda (x y) (if y (vector-set! x 0 0) (vector-set! x 0 1)) #t))
+  (not (cptypes-equivalent-expansion?
+         '(lambda (x y) (if y (void) (vector-set! x 0 0)) (vector? x))
+         '(lambda (x y) (if y (void) (vector-set! x 0 0)) #t)))
+  (not (cptypes-equivalent-expansion?
+         '(lambda (x y) (if y (vector-set! x 0 0) (void)) (vector? x))
+         '(lambda (x y) (if y (vector-set! x 0 0) (void)) #t)))
+  (cptypes-equivalent-expansion?
+    '(lambda (x y) (if (if y (vector? x) (error 'who "msg")) (vector? x) (void)))
+    '(lambda (x y) (if (if y (vector? x) (error 'who "msg")) #t (void))))
+  (cptypes-equivalent-expansion?
+    '(lambda (x y) (if (if y (error 'who "msg") (vector? x)) (vector? x) (void)))
+    '(lambda (x y) (if (if y (error 'who "msg") (vector? x)) #t (void))))
+  (not (cptypes-equivalent-expansion?
+         '(lambda (x y) (if (if y (vector? x) (error 'who "msg")) (void) (vector? x)))
+         '(lambda (x y) (if (if y (vector? x) (error 'who "msg")) (void) #t))))
+  (not (cptypes-equivalent-expansion?
+         '(lambda (x y) (if (if y (error 'who "msg") (vector? x)) (void) (vector? x)))
+         '(lambda (x y) (if (if y (error 'who "msg") (vector? x)) (void) #t))))
+  (cptypes-equivalent-expansion?
+    '(lambda (t) (vector? (if t (vector 1) (vector 2))))
+    '(lambda (t) (if t (vector 1) (vector 2)) #t))
+  (cptypes-equivalent-expansion?
+    '(number? (if t 1 2.0))
+    '(begin (if t 1 2.0) #t))
+  (cptypes-equivalent-expansion?
+    '(lambda (t) (fixnum? (if t 1 2)))
+    '(lambda (t) (if t 1 2.0) #t))
+  (cptypes-equivalent-expansion?
+    '(lambda (t) (boolean? (if t #t #f)))
+    '(lambda (t) (if t #t #f) #t))
+  (cptypes-equivalent-expansion?
+    '(lambda (t) ((lambda (x) (if x #t #f)) (if t (vector 1) (box 1))))
+    '(lambda (t) (if t (vector 1) (box 1)) #t))
+  (cptypes-equivalent-expansion?
+    '(lambda (t)(not (if t (vector 1) (box 1))))
+    '(lambda (t) (if t (vector 1) (box 1)) #f))
+  (cptypes-equivalent-expansion?
+     '(lambda (x y z f)
+       (let ([t (if x (vector 1) (box 1))])
+         (if (if y t z) (f t 1) (f t 2))))
+     '(lambda (x y z f)
+       (let ([t (if x (vector 1) (box 1))])
+         (if (if y #t z) (f t 1) (f t 2)))))
+  (not (cptypes-equivalent-expansion?
+          '(lambda (x y z f)
+            (let ([t (vector? x)])
+              (if (if y t z) (f t 1) (f t 2))))
+          '(lambda (x y z f)
+            (let ([t (vector? x)])
+              (if (if y #t z) (f t 1) (f t 2))))))
+  (not (cptypes-equivalent-expansion?
+          '(lambda (x y z f)
+            (let ([t (vector? x)])
+              (if (if y t z) (f t 1) (f t 2))))
+          '(lambda (x y z f)
+            (let ([t (vector? x)])
+              (if (if y #f z) (f t 1) (f t 2))))))
+)
+
+(define (test-chain/preamble/self preamble check-self? l)
+  (let loop ([l l])
+    (if (null? l)
+        #t
+        (and (or (not check-self?)
+                 (cptypes-equivalent-expansion?
+                   `(let ()
+                      ,preamble
+                      (lambda (x) (when (,(car l) x) (,(car l) x))))
+                   `(let ()
+                      ,preamble
+                      (lambda (x) (when (,(car l) x) #t)))))
+             (let loop ([t (cdr l)])
+               (if (null? t)
+                   #t
+                   (and (cptypes-equivalent-expansion?
+                          `(let ()
+                             ,preamble
+                             (lambda (x) (when (,(car l) x) (,(car t) x))))
+                          `(let ()
+                             ,preamble
+                             (lambda (x) (when (,(car l) x) #t))))
+                        (not (cptypes-equivalent-expansion?
+                               `(let ()
+                                  ,preamble
+                                  (lambda (x) (when (,(car t) x) (,(car l) x))))
+                               `(let ()
+                                  ,preamble
+                                  (lambda (x) (when (,(car t) x) #t)))))
+                        (loop (cdr t)))))
+             (loop (cdr l))))))
+
+(define (test-chain l)
+  (test-chain/preamble/self '(void) #t l))
+
+(define (test-chain* l)
+  (test-chain/preamble/self '(void) #f l))
+
+(define (test-chain/preamble preamble l)
+  (test-chain/preamble/self preamble #t l))
+
+(define (test-chain*/preamble l)
+  (test-chain/preamble/self preamble #f l))
+
+(define (test-disjoint/preamble/self preamble check-self? l)
+  (let loop ([l l])
+    (if (null? l)
+        #t
+       (and (or (not check-self?)
+                (cptypes-equivalent-expansion?
+                  `(let ()
+                     ,preamble
+                     (lambda (x) (when (,(car l) x) (,(car l) x))))
+                  `(let ()
+                     ,preamble
+                     (lambda (x) (when (,(car l) x) #t)))))
+            (let loop ([t (cdr l)])
+              (if (null? t)
+                  #t
+                  (and (cptypes-equivalent-expansion?
+                         `(let ()
+                            ,preamble
+                            (lambda (x) (when (,(car l) x) (,(car t) x))))
+                         `(let ()
+                            ,preamble
+                            (lambda (x) (when (,(car l) x) #f))))
+                       (cptypes-equivalent-expansion?
+                         `(let ()
+                           ,preamble
+                           (lambda (x) (when (,(car t) x) (,(car l) x))))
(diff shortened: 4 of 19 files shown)


dybvig:
Michael Adams created a type-recovery algorithm and implemented it for Chez Scheme as part of his dissertation research, with help from Andy Keep and others, but I never took the time to put it into the release stream, which would have involved fleshing out the primdata information, among other things.  Before incorporating another solution, it would be nice to give that work a close look.  [This paper](https://michaeldadams.org/papers/fast_flow_sensitive_cfa/paper-2011-oopsla.pdf) describes the algorithm.

gus-massa:
I read the preprint only twice, so I hope I didn't misunderstand too much.

There are a lot of small differences that can be ported in either way. For example, his implementation reduces `#2%car` to `#3%car` when posible, but I didn't add that. Also, he is using a lattice of types with a very small height, but I'm using 4 or 5 steps in the numeric tower.

The main difference is that he is using a graph to handle cycles, like `(let loop ([n 100]) ... )` and I just give up and don't try anything smart with cycles, so all the analysis can be done in one pass.

The other difference is that I'm always copying the info from each environment to the surrounding environment. This can lead to quadratic run time in cases like

    (lamdba (x1 ... xn)
      (let ([r1 (random)])
        ...
           (let ([r2 (random)])
             (display (list (unbox x1) ... (unbox xn)))))
      (display (list (box? x1) ... (box xn))))

He is using an abstract analysis of the flow of the types an structure similar to a skiplist to avoiding to update the types in each step, and the types just teletransport to the correct spot. So this is analyzed in O(n log(n)) instead of O(n^2).

As an informal benchmark, I'm comparing the runtime between the master branch and my branch. With something like 

cd a6nt/s; time make allx o=3

my branch, o=3:   28.6 sec
master,  o=3:   26.4 sec

(with o=2, it runs twice, so you must divide the time by 2 to compare)
my branch, o=2: 148.8 sec   (~74.4s sec)
master, o=2: 170.6 sec   (~85.3s sec)

I'm "cheating" a little because most of the times I'm not saving the type of a variable when it is a `procedure?`.  (My code is correct (I hope), but it miss some optimization opportunities, but not too much in usual code.) If I add all the procedures, the code has to track too many variables and the compilation gets ¿50%? slower. He is saving all the info for all procedures, with more details, so this can have an important effect.

gus-massa:
Sorry for the long delay. I updated the PR. This is a short description of the commits:

* fix enumerate signature
A small fix, otherwise the PR doesn’t compile.
---
* add signatures field to primref record
* Add cptypes pass to cp0 to reduce expression using types
These are the main parts of the old PR. See the description in the initial post.
---
* revert to version that can update primref record (skip)
* recalculate the bootfiles automatically in travis-ci (skip)
* bootfiles (signatures)
These are some commits to try to test the intermediate versions in Travis without updating the bootfiles with each change. These commits should be deleted of the final version.
---
* Fix hashing of prelex
* fxmap version of cptypes
* change fxmap api to hide internal structure
These commits change the hamt by an intmap/fxmap made by @97jaz. In particular the new structure can merge two similar intmap in `O(d log(n))`, where `d` is the number of differences and `n` is the total amount of variables. Also, the `prelex` have a new field that is a counter, that is used as the hash key for the map.
---
* remove some unnecessary pred-env merges
* update comments and fix exact-integer? and result type of ìf`
* merge types after if
* rename merge -> union/intersection
* refactor pref-union/intersect
These are some assorted changes, mostly to avoid unnecessary merges in the `pred-env` to try to avoid quadratic behavior in long chains of nested `let`’s. They also provide more information about the return type of an `if`. (They can probably be squashed in a single commit, but some of them are mixed with the previous commits and they are difficult to reorder.)
---
* upper and lower bound for types
* cache for primref->result/argument-predicate
* add safeongoodargs flag to primref
Try to replace some applications of the primitives with the unsafe version when the types of all the arguments are well enough. This changes `(car <pair>)` to `(#3%car <pair>)` but it doesn’t change `(vector-ref <vector> <n>)` to `(#3%vector-ref <vector> <n>)` because it’s difficult to verify if `<n>` is in the correct interval. There are surprisingly many primitives where all the arguments can be checked.
---
---
About the thesis of Michael Adams: The implementation is quite different but most of the results are equivalent. (Something similar to a list of features is in mats/cptypes.ss) My PR uses nanopass instead of a custom graph. 
My version currently passes the checks in Travis. The runtime of the o=3 version increase only a little and the runtime of the o=2 version is smaller. It can be merged after checking a few points that I’m not 100% sure that are correct and fixing a few/many style problems. 
 IIUC the main features that are missing in my version are:
* Types for loops: I don’t track the types in recursive functions like
```
    (let loop ([n 0])
      (when (< n 100)
        (loop (+ n 1))) ; <- obviously n is a number
```
I think this can be added to my current implementation later. It’s not straightforward but I have a vague idea about the implementation, but I have not written a single line of it.
* [fixed] Quadratic behavior with too many `let`’s, for example in 
```
    (let ([x1 (read)])
      (let ([xn (read)])
        (+ x1 xn)
      ) ; <- merge the types
    ) ; <- merge the types
```
In my first version the extended version of this code has quadratic behavior because the types of each `x1` and `xn` should be merged after each `let`. With the new version I just keep the dead variables in the environment. The new hashing is fast and these additional variables don’t make the code slower.
* Quadratic behavior with too many `if`’s, for example in 
```
    (if x1 
      (if x2 1 2)
      (if x3 3 4)
    ) ; <- merge the types discovered in each branch 
```
This example is not quadratic, and I’m still not sure if something like this can be quadratic in the current version. But if in the future I add support for loops and function calls, it will more easy to find quadratic examples.
The thesis has a chapter to fix the problem of the nested `if`. I think it’s very interesting but the solution is difficult to integrate with my version. In particular, IIUC in the thesis the position of each `if` is important, but I try to reduce the if’s as soon as possible (for example if `x2` is a number, the second `if` is reduced to the constant `1`). I think that long chains of `if` are not so common, so my version is good enough without this feature.
* [half fixed] Avoid unnecessary type checks
```
    (car <pair>) ==> (#3%car <pair>) [fixed]
    (vector-ref <vector> <n>) ==> (#3%vector-ref <vector> <n>) [not fixed]
    (bitwise-copy-bit <sint> <uint> <bit>) ==> (bitwise-copy-bit <sint> <uint> <bit>) [not fixed]
```
I added a reduction for the cases where it is possible to check all the arguments and replace the application with the unsafe version. This is possible with `car`, `cdr`, and many (~200) primitives, but it’s not possible with `vector-ref` and most vector primitives. It’s possible in the future to add some support for more general cases, in particular the operations on vectors.
A special case are the primitives like `bitwise-copy-bit`. They are marked as safeongoodarguments, but the type system I currently use can’t track the variables with enough precision, so they are not changed in the current versions, but it’s possible to extend the type system in the future.

samth:
Note that large uses of `match` (at least in Racket) can generate quite deeply nested `if` expressions. I'm not sure if that's likely to be a problem with the issue you mentioned.

gus-massa:
I´ll take a look at the expansion `match` in Racket, but I expect them not to be so big to cause a problem.

I think that it's easier to explain the problem with an example. In the current version of my PR this code:

    (lambda (t)
      (unless (symbol? t)
        (error 'e "msg"))
      (let ([x1 (read)]
            [x2 (read)]
            [xn (read)])
        (if (eq? t 'v1)
            (display (+ 1 x1 x2 xn))
            (if (eq? t 'v2)
                (display (+ 2 x1 x2 xn))
                (if (eq? t 'vm)
                    (display (+ -1 x1 x2 xn))
                    (display (+ -2 x1 x2 xn)))))
        (list (number? x1) (number? x2) (number? xn))))

The optimization step for this code is `O(n*m*log(N))`, where `n` is the number of auxiliary variables, `m` is the number of conditions and `N` is the total amount of variables (not shown in the example). But the length of the code is `O(n*m)` because each variable must be used in (almost) each branch condition.  If a variable is used only in one single branch in a long chain of `if`s, after the first type merge it will dispersar and all the the other merges will not see the variable.

So this simpler code code is optimized in a time essentially `O(n+m)` (ignoring something like `log(N)`)

        (if (eq? t 'v1)
            0
            (if (eq? t 'v2)
                1
                (if (eq? t 'vm)
                    (+ -1 x1 x2 xn)
                    (+ -2 x1 x2 xn))))))


I'm (slightly) worried than in a future version, I (or someone) will add some tracking of types inside function calls. Something like:

    (lambda (t)
      (unless (symbol? t)
        (error 'e "msg"))
      (let ([x1 (read)]
            [x2 (read)]
            [xn (read)])
        (let ([sum (lambda (z) (dislay (+ z x1 x2 xn)))]) ; let's assume it is not inined
          (if (eq? t 'v1)
              (sum 1)
              (if (eq? t 'v2)
                  (sum 2)
                  (if (eq? t 'vm)
                      (sum -1)
                      (sum -2))))
          (list (number? x1) (number? x2) (number? xn))))

In this case, if the optimizer is smart enough to realize that after calling `sum` all the captured variables `x1`, `x2`, `xn` are integers, then the merge of the types will be `O(n*m*log(N))` but the code will be of length `O(n+m)` only because sum is not inlined. 

This is not a problem in my current version, because it just ignore function calls. Only the primitives may mark the variables.

97jaz:
The complexity of the merge algorithm is given in Okasaki and Gill as worst case O(m+n).

akeep:
Hey @gus-massa,

Thanks for all your work on this!

It looks like this pull request is not quite ready to go though, because when I run the Chez Scheme unit tests I end up with a variety of failures.  I would note that you should not need to change the ``.travis.yml`` file to get things working, if your boot files are up to date.

Two types of errors I saw
--------------------------

I saw two types of unit test failures when running the ``mats``. 

First, there are a few places where the output of ``cp0`` is tested and the ``$cptypes`` pass improves on the output by using unsafe primitives in place of the type-checking version.  These (if you think they are correct) you should go ahead and update.  As part of the review process, these changes will be looked at as well.

Second, there are places where the changes in the PR are leading to some errors in the ``mats``.  The ``sint-list->bytevector`` procedure is incorrectly optimized when compiling the Chez Scheme libraries, and this is leading a handful to tests related to it to fail, and there are places where a specific error was expected and there is now an ``invalid memory reference`` error.

I spent some time looking into the ``sint-list->bytevector`` problem, and I was able to isolate it to an erroneous optimization that is getting rid of a loop in the case of a list of 16-bit signed integers.  The original ``bytevector.ss`` file is a little big to operate on, so I managed to pair it down to a smaller test that demonstrates the problem:

```scheme
(define $s16-native-list->bytevector
  (lambda (ls)
    (let ((v (make-bytevector (fx* (length ls) 2))))
      (let loop ((ls ls) (i 0))
        (unless (null? ls)
          (let ((k (car ls)))
            (unless (and (fixnum? k) (fx<= -32768 k 32767)) (errorf #f "invalid value ~s" k))
            (bytevector-s16-native-set! v i k))
          (loop (cdr ls) (fx+ i 2))))
      v)))
```

When I run ``expand/optimize`` on this with the current Chez Scheme from the git-repo, 2d1d3ad, I get the following:

```scheme
(begin
  (set! $s16-native-list->bytevector
    (lambda (ls.0)
      (let ((v.1 (#2%make-bytevector (#2%fx* 2 (#2%length ls.0)))))
        ((letrec ((loop.2 (lambda (ls.3 i.4)
                            (if (#2%null? ls.3)
                                (#2%void)
                                (begin
                                  (let ((k.5 (#2%car ls.3)))
                                    (if (if (#2%fixnum? k.5)
                                            (#2%fx<= -32768 k.5 32767)
                                            #f)
                                        (#2%void)
                                        (#2%errorf #f "invalid value ~s" k.5))
                                    (#2%bytevector-s16-native-set! v.1 i.4 k.5))
                                  (loop.2 (#2%cdr ls.3) (#2%fx+ 2 i.4)))))))
           loop.2)
          ls.0 0)
        v.1)))
  (#2%void))
```

However, when I run ``expand/optimize`` on this with the current Chez Scheme patched with your changes, it (incorrectly) optimizes away the loop:

```scheme
(begin
  (set! $s16-native-list->bytevector
    (lambda (ls.0)
      (let ((v.1 (#2%make-bytevector (#2%fx* 2 (#2%length ls.0)))))
        (if (#2%null? ls.0)
            (#2%void)
            (let ((k.2 (#2%car ls.0)))
              (if (if (#2%fixnum? k.2) (#3%fx<= -32768 k.2 32767) #f)
                  (#2%void)
                  (#2%errorf #f "invalid value ~s" k.2))
              (#2%bytevector-s16-native-set! v.1 0 k.2)))
        v.1)))
  (#2%void))
```

I also verified it was ``cptypes`` making the change by switching up ``run-cp0`` with the following:

```scheme
(run-cp0
  (let ()
    (define (echo msg x)
      (display msg)
      (newline)
      (pretty-print (#%$uncprep x))
      x)
    (lambda (cp0 x)
      (let* ((x (#%$cp0 (echo "original input" x)))
             (x (#%$cptypes (echo "output of cp0 (1)" x)))
             (x (#%$cpletrec (echo "output of cptypes (1)" x)))
             (x (#%$cp0 (echo "output of cpletrec (1)" x)))
             (x (#%$cptypes (echo "output of cp0 (2)" x)))
             (x (#%$cpletrec (echo "output of cptypes (2)" x))))
        (cp0 (echo "output of cpletrec (2)" x))))))
```

This printed out (amongst other things):

```scheme
---
output of cp0 (1)
(begin
  (set! $s16-native-list->bytevector
    (lambda (ls.6)
      (let ((v.7 (#2%make-bytevector (#2%fx* 2 (#2%length ls.6)))))
        ((letrec ((loop.8 (lambda (ls.9 i.10)
                            (if (#2%null? ls.9)
                                (#2%void)
                                (begin
                                  (let ((k.11 (#2%car ls.9)))
                                    (if (if (#2%fixnum? k.11)
                                            (#2%fx<= -32768 k.11 32767)
                                            #f)
                                        (#2%void)
                                        (#2%errorf #f "invalid value ~s" k.11))
                                    (#2%bytevector-s16-native-set! v.7 i.10 k.11))
                                  (loop.8 (#2%cdr ls.9) (#2%fx+ 2 i.10)))))))
           loop.8)
          ls.6 0)
        v.7)))
  (#2%void))
output of cptypes (1)
(begin
  (set! $s16-native-list->bytevector
    (lambda (ls.6)
      (let ((v.7 (#2%make-bytevector (#2%fx* 2 (#2%length ls.6)))))
        ((letrec ((loop.8 (lambda (ls.9 i.10)
                            (if (#2%null? ls.9)
                                (#2%void)
                                (let ([k.11 (#2%car ls.9)])
                                  (if (if (#2%fixnum? k.11)
                                          (#3%fx<= -32768 k.11 32767)
                                          #f)
                                      (#2%void)
                                      (#2%errorf #f "invalid value ~s" k.11))
                                  (#2%bytevector-s16-native-set! v.7 i.10 k.11))))))
           loop.8)
          ls.6 0)
        v.7)))
  (#2%void))
---
```

You'll notice the recursive call to ``loop.8`` was eliminated.  My suspicion (though I've not tracked this through the ``cptypes`` code) is that the something with the merge points around the ``(unless (and (fixnum? k) (fx<= -32768 k 32767)) (errorf ---))`` is merging to ``bottom`` when it should not be, or the ``pred-implies?`` is incorrectly identifying the result as bottom, which is causing it not to generate the `loop`` call in the output.  I could be wrong though, there could be something more subtle going on here, and it is worth noting that similar shaped code for the 32-bit signed integer does not manifest the same problem.

A bug like this might also be what is causing the ``invalid memory`` exceptions I'm seeing.

A note on style
----------------

I noticed it looks like you are using a ``box`` to represent an input/output parameter.  Instead of using boxes I recommend using multiple-value returns.  The advantage is that you avoid allocation, since (in the worst case) multiple-value returns are handled similar to a function call, and in the best case, it is possible for the multiple-value consumer to be pushed with in the multiple-value producer (see: [An Efficient Implementation of Multiple Return Values in Scheme](https://www.cs.indiana.edu/~dyb/pubs/mrvs.pdf) by Ashley & Dybvig).

A note on asymptotic behavior
--------------------------------

Like @samth, I'm a little concerned about the potential for _O(n^2)_ behavior, traditionally we've tried to keep optimizations to _O(n log n)_, only allowing for _O(n^2)_ passes where it was required for correctness.  That said, if we can get this into shape, we can do some testing to see how deeply nested the ``if`` expressions have to get before we really start to see performance degradation in the compiler, and in the worst case, we can always cut off the analysis if we discover we are going down a path that is going to lead to bad performance.  In practice, as you've said, we may find we rarely hit the limit.

Let me know if you'd like help on this, and thanks again for the work you've already put in!

gus-massa:
Thanks for the minimized example. The problem with `bytevector-[u/s]16-native-set!` was that it has a wrong signature. (The 32 bits versions is correct.) So cptypes detects that the variable `k` is a `fixnum?` and it uses the wrong signature to detect that `k` must be a `symbol?`, so it expect that the `bytevector-u16-native-set!` will generate an error, so cptypes drops the rest of the expression.

I submitted a fix in a different PR https://github.com/cisco/ChezScheme/pull/240 because this looks isolated enough and simple enough to be merged soon.

I'll reply to the rest of your comment later.

gus-massa:
I'm working mostly in Windows, but I never could fix the problems with `iconv.dll`, so if I run all the test I get a lot of errors. I'm using a VM with Linux now, so I can see the actual errors.

I found a few more (~5) errors with the signatures, that were causing explicit error or invalid memory errors. (I'll submit the PR soon.) I also found a a few places were I marked the primitive erroneously  as `safeongoodars`. I also fixed the checks of the expanded code. (I'll update this PR in a week after I fix most of the problems.) 

I still have to fix the errors with records. @mflatt found an error that is probably related to records, but he has no an small/easy example. Some of the messages of the errors in the mats look similar, so I hope that fixing the mats will fix it.

About Style (box -> values): OK, I'll change it

About asymptotic behavior: I'll try to think the details later. I'll try to fix the other problems before.

akeep:
That all sounds good, thanks again for all of your hard work on this!

As the first real consumer of the signature data in ``primdata.ss``, I'm afraid you've ended up being the tester for that data.

I'm not a windows user, but you might consider putting in an issue about the problems you've run into running the tests there, as this is something that should be possible, and there are windows users in the Chez Scheme community that might be able to help out.

burgerrg:
@gus-massa, try https://github.com/win-iconv/win-iconv to make an iconv.dll that works on Windows. I use this for x86 and x64.

gus-massa:
@burgerrg  I tried now to compile win-iconv, but I got an error (Perhaps because I'm using Cygwin instead of MinGW?) Anyway, I downloaded the compiled version from https://mlocati.github.io/articles/gettext-iconv-windows.html and I copied the file libiconv-2.dll to ChezScheme\a6nt\bin\a6nt and it works.

I still get some errors in the master branch in Windows. I'll try to classify them and send a pull request or a bug report if I can't fix it.

@akeep In my "secret" branch, I fixed all the errors that the test in "allx" can detect. But I want to make a few more tweaks before updating the PR. I'm optimistic about the problems with the quadratic time, I think it's solvable with a few modifications to the current structures. I also tried to change the box to values, but some parts are painful. For example to look at all the arguments of a function: 

    ;--- with box ---
    [(call ,preinfo ,pr ,e* ...)
     (let* ([r* (map (lambda (e) (box #f)) e*)]
            [t* (map (lambda (e) (box (unbox types))) e*)]
            [e* (map (lambda (e r t) (cptypes e 'value r t #f #f))
                     e* r* t*)])
       (set-box! types (something-with (unbox types) e* r*))
       (for-each (lambda (t) (set-box! types (pred-env-intersect (unbox types) (unbox t)))) t*)
       `(call ,preinfo ,pr ,e* ...))]

and 

    ;--- with box ---
    [(call ,preinfo ,pr ,e* ...)
     (let* ([e/r/t*  (map (lambda (e)
                            (let-values ([(e r t t-t f-t)
                                          (cptypes e 'value types)])
                              (list e r t)))
                          e*)]
            [e* (map car e/r/t*)]
            [r* (map cadr e/r/t*)]
            [t* (map caddr e/r/t*)])
       (let ([types (something-with types e* r*)])
         (values `(call ,preinfo ,pr ,e* ...)
                 #f
                 (fold-left (lambda (types t) (pred-env-intersect types t)) types t*)
                 #f #f)))]

akeep:
@gus-massa I agree that looks pretty hairy, but I would approach this a little differently.

First, instead of using map and generating a list of lists, then mapping over the lists, I'd probably consider either building a loop to do this in place, or I'd write a multi-valued map macro that allowed for specification for the base-case null list values.  In this case though, the nanopass framework can help out, because it can already create the multi-valued map for you:

```scheme
[(call ,preinfo ,pr ,[e* r* t* t-t* f-t*])
 (let ([types (something-with types e* r*)])
   (values `(call ,preinfo ,pr ,e* ...)
           #f
           (fold-left (lambda (types t) (pred-env-interset types t)) types t*)
           #f #f))]
```

If you take a peak at ``cpnanopas.ss``, you'll also see places where we've written custom loops, or extracted that into a function when we were going to use that several times.

gus-massa:
After another long delay ... Fixing the quadratic time was possible, but it was harder than I expected.

I squashed most of the commits and the fix in a few big commits. Short description:

---
* add `signatures` field to primref record  …
* Version to upate primref record during rebase: step 1 (skip)
* Version to upate primref record during rebase: step 2 (skip)
* Version to upate primref record during rebase: step 3 (skip)
* recalculate the bootfiles automatically in travis-ci (skip)
* bootfiles (signatures)

Add the `signatures` to the `primref`. This step is tricky, so I left the intermediate steps to make a future rebase easier, but most of these commits should be removed in the final version.

---
* Hashing of prelex for cptypes
* Add fxmap for cptypes
* Add more operations to fxmap
* test for fxmap-{remove,reset}/base (skip)

Implementation of an intmap that will hold the types recovered during the cptype pass. 

---
* Remove special case for (#2%map p '()) in cp0 [fix in master]  …
* Add cptypes pass to cp0 to reduce expression using types  …

This is the main part of the PR. This includes the previous version, some improvements and many fixes.

---
* add safeongoodargs flag to primref  …

Reductions to replace a primitive by the unsafe version. I left the safeongoodargs separated because it has a lot of instances to be reviewed, so perhaps it's better to merge it later.


---
---

About the previous comments:

### box -> values
done

### fix tests
done

### records
I changed like the 90% of the implementation for records. I think that now it is correct.

### quadratic time
The main idea is that the types are stored in a immutable intmap/fxmap that remembers how much changes were used to create it. So each time it is necessary to merge two intmaps it is possible to select the smaller one and add to the biggest one only the recent differences. The main advantage of this intmaps/fxmaps is that the difference operation can be very efficient.

Some construction like `let` use a special operation to forget the temporal variables. This operation not only removes the variable but it also tries to reconstruct the inmap before the temporal variable was added. [In particular, with this reconstruction some trees that look `equal?` are `eq?`, like in the additional test. Most of the times only a part of the tree is reconstructed.] This helps the following parts of the code to compare two inmaps faster, because the iterations can skip the common parts using `eq?` without comparing the subtrees. 

The analysis of the `if`  is more tricky, because it's necessary to use a different strategy when the `test` part is longer than both branches. Also, if one one of the branches is longer than the `test` it's important to use the intmap of the smaller branch for the iterations and compare using lookups in the intmap of the biggest branch. (The "length" is actually calculated the number of changes in the relevant intmap, that is bounded by the actual length of the code.)

This pass runs in O(N log(N) (log(N)+R)) where N is the total length of the code and R is the depth of the record hierarchy, because to compare two records the program must walk from a rtd to the parents until it finds the other rtd or the base rtd. In practice R is small, I tested with the Chez Scheme compilation and I got R=3, and I expect that most programs don't have rtd with parent  with parent  with parent  with parent  with parent. So the effective bound is O(N log^2(N)).

[It is possible to fix the problem of the records with too many ancestors, but in that case I think it's better to change it in another time.]

akeep:
This all sounds good.  It will take me a few days to get some time to look through and test all these changes, but I'm looking forward to going through it. Thanks for all the work 👍

akeep:
Hi @gus-massa,

Sorry it has taken me so long to get back to you.  I found a few bugs in testing and ended up fixing them myself as I was reviewing and experimenting a bit with the code.  Unfortunately, I was unable to push the changes directly into this pull request, so I have them up on a GitHub desktop created branch called [pr/192](https://github.com/cisco/ChezScheme/tree/pr/192).

Before I get into the list of the things that I changed (and where), there were a few things that I did not do, that are minor changes we still might want to consider.

In the ``fxmap`` code, it looks like ``$empty`` is probably a singleton and we might get away with an ever so slightly faster ``eq?`` check using the singleton instead of using ``$empty?``  (It is even possible to force the ``$empty`` record protocol to produce a singleton, but that is likely overkill.

In the ``cptypes`` code:
---------------------------
We might consider making $record/rtd $record/ref lists into records... it is both more space efficient (2 words for the record instead of 4 for 2 pairs) and more performance efficient when checking for membership.

I think there are still a handful of places where we can avoid doing multiple ``car``/``cadr``/etc. calls on the same object.  Less important for ``car``/``cdr``, but for ``cadr``, etc. slightly more important since we're walking multiple pointers repeatedly doing that.

The primitive call handling is a bit complicated, and I'm wondering if it is worth it to follow the ``cp0`` example of creating handlers for the primitives.  There may be some code we can communize in here, and I think it might clean things up a bit, though we are sort of on the borderline of needing more complicated handling.

On the performance side, I suspect there is still quadratic behavior lurking in here somewhere, but with the changes you've made I suspect it would be very difficult to construct the program that exposed this.  Part of the reason I say this is that the merging of ``fxmap``s has a linear worst case, and the handling of call arguments also does a linear walk over the items bound in the types maps when the ``pred-env-rebase`` function is called.

That said, I'm happy enough with where the performance is on this.  I think you've done a lot of good things to curtail the cases where we might run into trouble, and it doesn't seem to impact the compile time much if at all on my mac, though I've not done any formal benchmarking of it.

So, onto the changes I've made on my branch:

Removed ``counter`` field from ``prelex``, using the ``operand`` field instead to provide the index into the ``fxmap``.  This follows other uses within the compiler where we use the ``operand`` field as a little place for state that is used within a single pass.  This has a few advantages.  First, it keeps the record a little smaller.  Second, it means that the ``prelex`` numbering can start from ``0`` for each compilation unit, which should help keep the numbers for the ``fxmap`` a  bit smaller in longer running sessions with multiple calls to the compiler.  Finally, it avoids adding to the burden of the ``tc-mutex``, since the ``counter`` is now a local variable bound within the pass and it is safe for us to set the ``prelex``es, since only the instance of the pass holding this block of code has a handle on it.  As part of this change ``prelex-counter`` is now defined in ``cptypes`` and the ``operand`` is cleared after the variables go out of scope.
```
  base-lang.ss, cptypes.ss
```

Fixed the ``highest-set-bit`` function in ``fxmap`` so that it will work in the 32-bit versions of Chez Scheme.  The ``fxsrl`` by 32 raises an exception in the 32-bit compiler, and was leading to tests to fail in 32-bit mode.
```
  fxmap.ss
```

Restructured ``predicate-implies?`` so that it uses committed choice instead of uncommitted choice in comparing ``x`` and ``y``.  Basically, this means, instead of doing:
```
(or (and (predicate-1? x) (predicate-1? y) ---)
     (and (predicate-2? x) (predicate-2? y) ---)
     ...)
```
we now do:
```
(cond
  [(predicate-1? x) (and (predicate-1? y) ---)]
  [(predicate-2? x) (and (predicate-2? y) ---)]
  ...)
```
This avoids running predicates on ``x`` that we know will fail because an earlier predicate matches, generally getting out of the ``predicate-implies?`` faster.  This did require a little restructuring, because in some cases ``x`` was dominant and in other cases ``y`` was dominant.  This is now restructured with ``y`` dominate, after the ``eq?`` and ``x 'bottom`` check.
```
  cptypes.ss
```

Replaced ``let-values`` calls with cata-morphism syntax, including removal of ``map`` calls that built up a list of values that then needed to be separated out with ``(map car ...)`` ``(map cadr ...)`` etc. calls.  This avoid building up structures we don't need, since the nanopass framework will generate a mutltivalued ``map`` for these situations.
```
  cptypes.ss
```

The ``if`` clause in ``cptypes/raw`` now uses ``types1`` (the result of the recursive call
on ``e1``) in place of the incoming ``types`` value when processing the ``e2`` or ``e3`` expressions in the cases where ``e1`` is known statically to produce either a false or non-false value.
```
  cptypes.ss
```

Fixed a bug with directly-applied variable-arity lambda.  The original code marked all directly-applied variable arity lambda's as producing bottom, because it was checking for the interface to be equal to the number of arguments.  However, variable arity functions are represented with a negative number.  For instance, the original code would transform the expression:
```
(begin
  ((lambda (a . b) (set! t (cons* b a t))) 'a 'b 'c)
  t)
```
to
```
((lambda (a . b) (set! t (cons* b a t))) 'a 'b 'c)
```
anticipating that the call would raise an error, however, it is a perfectly valid (if some what unusual) expression.  I ended up adding tests that run ``cptypes`` without ``cp0`` to get this type of expression to ``cptypes``, so we can test it.  The code also chooses the correct clause from a ``case-lambda`` with multiple clauses.
```
  cptypes.ss
```

Fixed ``make-time``, ``time-second-set!``, and ``time-second`` to indicate that second can be an ``exact-integer``, since it is not really restricted to the fixnum range (and in fact we even test this fact in the mats on 32-bit machines).
```
  primdata.ss
```


Changed check of ``prelex-was-assigned`` (which is not reliably set on the input to any given pass) with ``prelex-assigned``, which should always have an accurate, if conservative, value in it.
```
  cptypes.ss
```

Added ``enable-type-recovery`` parameter to allow the type recover to be turned on and off (defaults to on), and added ``cptype`` to the ``cp0`` not run path that runs ``cpletrec``, so that ``cptypes`` can be run independent of ``cp0``.  This is helpful for testing and allows us to benefit from type recovery, even in cases where we do not want ``cp0`` to do anything that would re-arrange the input code
```
  compile.ss, front.ss, primdata.ss
```

Stylistic changes, mostly for consistency with other parts of the compiler, though I'm not married to these changes if you'd really prefer to keep things the way the are.
1. clauses of define-record type now use parenthesis instead of square brackets.
2. indented by 2 spaces where things were only indented by one space
3. define, let, define-pass, nanopass pass productions clauses, now use
   parenthesis for outer markers instead of square brackets.
```
  fxmap.ss, cptypes.ss
 ```

Sorry for the changes being written to a branch, there is probably a better way for us to collaborate on this code, but sadly I'm not as experienced with the GitHub tooling as I probably should be and didn't want to spend the whole evening searching for a better solution.

Thanks again for all your hard work on this, let me know what you think and how you'd like to proceed.

-andy:)

**Comment by akeep** [comment]
One thing I almost forgot, I have not yet updated the documentation to add the ``enable-type-reccovery`` to CSUG, but we'll need to do that as well before we add pull in these changes.

**Comment by gus-massa** [comment]
There are three lines in cptypes.ss that have a `#;CHECK` reminder/warning. These are parts that I think that are correct, but I'm not sure. (For example, the clause for `immutable-list` in `cptypes/raw`.) I think that you already looked at the code and thought it was fine, but can you take another look at them so I remove the reminder/warning?

**Comment by akeep** [comment]
> ## About the ideas for changes:
> * change ($empty? x) -> (eq? x <singleton>)
>
> OK. I'm only worry about the possibility in the future of reusing fxmap.ss in another pass (like cp0) and getting a different "singleton". This would be a problem is the fxmaps are somehow stored and shared. This problem is too hypothetical, so it can be solved with a comment in the file with a warning for the future.

Actually, you are already treating this as a singleton in places, where you return ``empty-fxmap`` in ``fxmap-remoe`` and ``fxmap-remove/base`` and when you initialize ``pred-env-empty``, it is treated as potentially different only in ``fxmap-for-each/diff``

> There are three lines in cptypes.ss that have a #;CHECK reminder/warning. These are parts that I think that are correct, but I'm not sure. (For example, the clause for immutable-list in cptypes/raw.) I think that you already looked at the code and thought it was fine, but can you take another look at them so I remove the reminder/warning?

The ``immutable-list`` and ``cte-optimization-lock`` look good to me.  ``immutable-list`` is a special case that allows for ``cp0`` to look inside a list it knows was constructed during expansion of some internal syntax when we know there is no way it could be modified, so we can treat it as constant without doing any escape analysis on it.

The ``predicate-implies?`` check is a good question.  Obviously, ``eq?`` is the safest check here, though I think ``eqv?`` is also safe enough.  In theory, I believe that these are quoted constants which should not mutated, though Chez Scheme does not prevent this mutation.  That said, I think I'd rather stick with ``eqv?`` right now, unless we find some places where we think this is being overly conservative.  If we do loosen it up, we may also want to consider checking for record equality, which r6rs ``equal?`` does not do.

**Comment by gus-massa** [comment]
> The `predicate-implies?` check is a good question.  ...

I think it's better to use `eqv?` because `predicate-implies?` is used in `predicate-implies-not?`. So we get these reductions

    (eq? 3.3 3.3) => #t      [unespecified]
    (eq? 3.3 7.7) => #f
    (eqv? 3.3 3.3) => #t
    (eqv? 3.3 7.7) => #f

The first one is officialy unespecified, but using `eqv?` in `predicate-implies?` makes cptypes reduce it to `#t` allways. Using `eq?` would make the it sometimes be reduced to `#f`, because the result would depend on some of the internal details, for example if both `3.3` are constant propagated from the same origin, or if they are calculated independently, or if there was a round trip to fasl. 

Something similar applies to the third one, but the result of the third one must be `#t` , so comparing internaly with `eq?` would be a problem.

----

I also found another small optimization of the program in cptypes. I removed the wrapper around `cptypes/raw` and now the recursion calls `cptypes` directly. The wrapper was there to replace `t-types` and `f-types` with the value of `types` when one of them was `#f` (that means the there are no additional predicates for the `#t` or `#f` branches). I now take care of that case inside the clause of `if` when it is necesary, instead of in every recursive call. So it must save a few milliseconds here and there. (I also had to make a trivial tiny modification in the clause of `let/let*/letrec`, but I have to debug the code for a while until I notice that I must fix it too.) [preview ](https://github.com/gus-massa/ChezScheme/commit/c98e69c234da63433abbb6554e60402c4a18fdeb)

**Comment by akeep** [comment]
Sounds good.  If the recursion is now within ``cptypes`` and the body expression is just a call to ``Expr``, then all of the cata-morphisms can have the ``cptypes :`` removed.  This will cause ``Expr`` to recur to ``Expr`` directly for the cata-morphisms.  It probably won't have a measurable impact on performance, but it does simplify the code a bit.

I agree on the ``eqv?``, I think getting the numeric equivalence is worthwhile.

**Comment by akeep** [comment]
Cool.  I'll can rebuild the boot files after merging this in to a local copy of master.

We should get the LOG and CSUG changes pulled in too if we can.  I can take a look at this over the weekend, and hopefully get things put together.

Just so I understand, it sounds like the things you checked in on this pull request are up to date, with the exception of the LOG and CSUG changes.  As far as the travis changes, I've actually re-done how travis checks things on master and it should be doing the checking correctly now, though it will still run afoul of the changes that invalid the boot files.

**Comment by akeep** [comment]
It took me a bit more effort than I expected to get the boot files built, but I managed to get these built and tested.

Unfortunately, in testing I ran into what looks like it might be an unexpected error.  I was testing on `a6osx`, `i3osx`, `ti3osx`, and `ta6osx`, and ran into a strange bug on one of the runs of `ta6osx`.

I'm getting the error: `pred-env-union/base: ` when running one of the threaded tests.  My suspicion is that something is getting fouled up with all the threads running (there is a call to `eval` in the thread test, which is invoking the `cptypes` pass:

```scheme
219     (define (pred-env-union/from from base types new-base)
220       ; Calculate the union of types and from, and intersect it with new-base
221       ; Iterate over the difference of from and base.
222       (let ([ret new-base])
223         (fxmap-for-each/diff (lambda (key x y)
224                                (let ([z (fxmap-ref types key #f)])
225                                  ;x-> from
226                                  ;y-> base
227                                  ;z-> types
228                                  (set! ret (pred-env-add/key ret key (pred-union x z)))))
229                              (lambda (key x)
230                                (let ([z (fxmap-ref types key #f)])
231                                  ;x-> from
232                                  ;z-> types
233                                  (set! ret (pred-env-add/key ret key (pred-union x z)))))
234                              (lambda (key x) (error 'pred-env-union/base "") (void))
235                              from
236                               base)
237           ret))
```

It looks like line `234`, in this test is getting something unexpected, but I have not yet figured out why, nor do I have a good way to reproduce this quite yet.

The fact that this only showed up in one machine type of the two that run the thread tests and then only in one of the twelve runs of this tells me that something intermittent (and well buried) is going on in this code, so it might take some effort to ferret it out.

Just FYI, we should update this error message to use `$oops` (or `$impoops` if you think this should be impossible) with a `who` argument like `compiler-internal` and a little bit more meaningful error message to give us some kind of indication as to what might have failed.  We try to stick to using `$oops` or `$impoops` inside the compiler, though I noticed a few `error` or `error` calls have leaked in from place to place.

The actual error I saw was in the following summary from a `ta6osx` run, the three errors are all in a row and the second one is almost certainly the result of the first, though the third one may or may not be related:

```
% cat ../mats/summary
-------- o=0 --------
-------- o=3 --------
-------- o=0 cp0=t cl=3 --------
-------- o=3 cp0=t cl=3 --------
-------- o=0 spi=t p=t --------
-------- o=3 spi=t p=t --------
-------- o=0 eval=interpret cl=6 --------
-------- o=3 eval=interpret cl=6 --------
-------- o=0 cp0=t eval=interpret --------
-------- o=3 cp0=t eval=interpret --------
0a1,3
> thread.mo:Error in mat thread clause 57: "pred-env-union/base: " at line 982, char 3 of thread.ms
> thread.mo:Error in mat thread clause 58: "extra threads running (#<thread 913> #<thread 912> #<thread 911> #<thread 910> #<thread 909> #<thread 908> ...)" at line 1035, char 3 of thread.ms
> thread.mo:Bug in mat thread clause 59 at line 1036, char 3 of thread.ms
-------- o=0 ehc=t eoc=f cl=9 --------
-------- o=3 ehc=t eval=interpret --------
```

akeep:
Oh, almost forgot.  I'll continue to poke at this a bit, but any insight you can provide would be helpful.

Thanks!

gus-massa:
[I still have problem with my main computer, so I have to guess a little ...]

One important part of  `(pred-env-union/from from base types new-base)` is it must be possible to construct `from` starting with `base` and using a few `add` operations. Sometimes the process is more complicated, but the operations like `remove/base` ensure that there is an alternative path that uses only `add` operations.

The error you found in the test is triggered because `base` has a binding that is not in `from`. This is impossible if you use only `add` operations (`add` actually intersects the new type with the old value, it's not a simple overwrite, it never cleans a binding). I think the `error` procedure should be changed to `$impoops` because in a bug free universe it is impossible to reach this part of the program.

I still don't understand where the problem is, but something like the following example should cause a problem. Each variable has a friendly name and also a number that is the counter that is temporally assigned to the `prelex-operand` field. The intmap uses only the counters, not the name or the prelex itself. I assume that there is a rogue prelex, that is shared between two threads, and the different threads reuse the numbers.

    (lambda (t1 x2 y3)
      (when (box? shared4)    ;  <-- the number 4 was assigned by another thread
        (if t1
            (let ([temp4 (read)])
              (display (list (unbox temp4) (unbox temp4)))
              )    ;  <-- here temp4 is removed, but this "removes" shared4 too
            (display (list (unbox x2) (unbox y3))
              )    ;  <-- enough recovered types to make this the long branch
            )    ;  <-- here the types are merged, the missing shared4 is a problem
        ))

The error will be raised only if the numbers in each tread are luckily aligned. In other cases there may be silent errors. This code is not similar to the code in [thread.ms#L979](https://github.com/cisco/ChezScheme/blob/master/mats/thread.ms#L979) so I can't guarantee this is the problem. My guess is that the imported identifiers in the examples are the shared prelex. This is probably not a problem in `cp0` because they don't get any value overwritten in the `prelex-operand` field.

akeep:
I believe the only place more than one compiler session should have its hands on the same prelex is as part of the `$cte-optimization-info` related to a symbol or as part of the box in the `cte-optimization-loc`.  `cp0`, depends on this being true for more than just the `prelex-operand` info, it is also modifying the `prelex-name`, which is uses to replace incoming `prelex` records with new `prelex` record.  `cpnanopass` also relies on this when it replaces the `prelex` with `uvar` records.  So, if we can have multiple compiler sessions running simultaneously with their hands on the same `prelex` I would expect these places to also be a problem, though the nature of race conditions is that they might not show up very often, so it is possible we've just not had quite the right conditions to expose this.

All of that said, we should be able to look into this and check for it by putting in some auxiliary information in the `cptypes` pass to identify a `prelex` we are seeing for the first time that already has its `prelex-operand` set.  We've certainly had cases where we were allowing a single `prelex` record to be used at more then one binding form within the system, which led to some problems.

One other thing worth noting on this is that `prelex` records are created as part of the expander process, and each compiler session runs its own expander, so it is creating new `prelex` records for the code it produces.  In order for multiple sessions to have a handle on the same `prelex` it would need to be the case that output from a single expander session got sent to multiple compiler instances.  The only way I know of that this can happen is from a code fragment being inlined from the `cte-optimization-loc` box (aka the cross-library optimization information) from another library, which I don't think we should be seeing here, but that is definitely worth looking in to.

It might be a few of days before I get a chance to spend much time on this again, but hopefully we can get a reproducible, which will make me more confident in our fix :)

One question I did have in the `fxmap` code is why you decided to have the merge function run by side-effect instead of constructing the output functionally?  Was this just to try to optimize around the O(m + n) worst case of the merge function?

gus-massa:
I made a small debugging change to `prelex-counter` function to remember the `prelex` that it has modified adding the `counter`, and so detect any `prelex` numbered by another thread and raise an error. It is in my branch [17-5-Types-Pass-Error](https://github.com/gus-massa/ChezScheme/commits/17-5-Types-Pass-Error), with two auxiliary commits to make travis-ci happy.

With the two auxiliary commits [all the tests pass](https://travis-ci.org/gus-massa/ChezScheme/builds/370800404), but with the main commit I get an [error exactly in all the threaded versions](https://travis-ci.org/gus-massa/ChezScheme/builds/370817470). (This change is more sensitive than the test error you found, because it will catch any bad `prelex`, instead of catching only the cases where the numbering and recovered types are correctly aligned.) (I couldn't find a way to see the error in travis-ci, `grep` is eating the output.)

I hope it's helpful.

----

About the side effects instead of the functional approach in the merge: To avoid the `O(n+m)` worst case I need that:
- the `counter`s of the `prelex` are almost continuous (or at lest that they don't increase exponentially) so the big trees look like trees instead of linked lists
- the branches that look `equal?` are actualy `eq?`, so if one tree is constructed from nother tree and they are almost equal then the common parts can be detected with `eq?` and skipped in block, without looking at each leaf.

So I need some merge operations that use tree (or four?) `intmaps`. There was a functional merge written by @97jaz , but it was not enough because it uses two trees and the result is a combination of the two trees. But I sometime must paste the results into a third tree to ensure that the common part between the result and the third tree is `eq?` and not just `equal?`. 

Once I thought about adding a functional version that calls the configurable function with an additional argument that acts like an accumulator, but I didn't understand the code well enough. (Now I think I can, but the code is still tricky.) (In case this is not clear, in the equivalent code with `list`s, it's like replacing a combination of a `for-each` and a `set!` by a `foldr`, the `lambda` needs an additional argument.)

akeep:
Sorry that I've been a little slow getting back to this, work was a bit hectic and it has taken me a little bit of time to get back to this.

Just to make sure I understand, the differences currently in this PR are with your changes that seem to run without error across the board, but if I take a look at your branch: [17-5-Types-Pass-Error](https://github.com/gus-massa/ChezScheme/commits/17-5-Types-Pass-Error) it will fail reliably in the threaded tests.

Is your suspicion that the changes in the current PR still have the bug, but have chased it into hiding?

I've made some changes in a local copy of this to match recent commits, merged master into this set of changes, and I'm currently running all the tests in `a6osx`, `i3osx`, `ta6osx`, and `ti3osx` on my mac.

So far this seems to be running along fine, but if we think there are still bugs hiding at the intersection of the `prelex-counter` code changes and the threaded versions, I can try to dig into that a bit (though I'm not sure if I'll have a chance to finish it up this week, and I'm traveling over the long weekend).

Thanks for continuing to push this along, I feel like we're really close to getting this in.

gus-massa:
I think we agree, and the current version of this PR has a problem with the `prelex`s that are somehow shared between threads.

In [17-5-Types-Pass-Error](https://github.com/gus-massa/ChezScheme/commits/17-5-Types-Pass-Error) I just added an additional checks so any strange `prelex` will raise an error. It doesn't change the result of the program and it should never be triggered, but it is actually triggered reliably in all the threaded compilations in travis.

akeep:
I had a little time today to experiment with the [17-5-Types-Pass-Error](https://github.com/gus-massa/ChezScheme/commits/17-5-Types-Pass-Error) branch and you are not testing what you think you are testing.

The problem with the test is that the counter and hash table are effectively being allocated only once, instead of each time cptypes is run.  Since cptypes is run multiple times it is actually the same thread that is seeing the same prelex across multiple invocations of cptypes.  When I looked at the test my first thought was that I wanted to have proof that it was, in fact, different threads that were being the same prelex, so instead of storing just the index in the prelex-operand flag, I had it store a pair of the count and the thread-id (from `get-thread-id`).  When I ran this, I had the error message include the thread id as part of the result.  What I saw was that it was the same thread id that had both assigned the prelex and was now re-assigning the prelex.

The updated `prelex-counter` code:

```scheme
  (define prelex-counter
    (let ()
      (define count 0)
      (define known (make-eq-hashtable))
      (lambda (x)
        (when (prelex-operand x)
          (let ([k (eq-hashtable-ref known (car (prelex-operand x)) #f)])
            (if k
              (unless (eq? x k)
                (newline)
                (printf "matting !!! [cptypes] duplicated prelex counter with ~s and ~s [~s (on thread: ~s)] ~s (on thread: ~s)" x k (car (prelex-operand x)) (cdr (prelex-operand x)) count (get-thread-id))
                (newline) 
                ($impoops 'cptypes "duplicated prelex counter with ~s and ~s [~s (on thread: ~s)] ~s (on thread: ~s)" x k (car (prelex-operand x)) (cdr (prelex-operand x)) count (get-thread-id)))
              (begin
                (newline)
                (printf "matting !!! [cptypes] unexpected prelex counter with ~s [~s (on thread: ~s)] ~s (on thread: ~s)" x (car (prelex-operand x)) (cdr (prelex-operand x)) count (get-thread-id))
                (newline)
                ($impoops 'cptypes "unexpected prelex counter with ~s [~s (on thread: ~s)] ~s (on thread: ~s)" x (car (prelex-operand x)) (cdr (prelex-operand x)) count (get-thread-id))))))
        (or (let ([p (prelex-operand x)])
              (and p (car p)))
            (let ([c count])
              (set! count (fx+ count 1))
              (prelex-operand-set! x (cons c (get-thread-id)))
              (eq-hashtable-set! known c x)
              c)))))
```

One run of compiling the `thread.mo` mats for instance showed (amongst other some cascading thread errors):

```
Error in mat thread clause 57: "cptypes: duplicated prelex counter with ignore and ignore [3631 (on thread: 4141)] 3986 (on thread: 4141)" at line 982, char 3 of thread.ms
Error in mat thread clause 59: "cptypes: duplicated prelex counter with tmp and x [5177 (on thread: 4161)] 5295 (on thread: 4161)" at line 1036, char 3 of thread.ms
```

In both cases the setter and the detection were on the same thread (`4141` in the first case `4161` in the second case).

One way to avoid this problem is to move the creation of the hash table and the counter within the pass (probably I should have done this when I changed `prelex-counter` in the first place, since the intention was for count to start at 0 for each case).  I ended up moving the ``define-pass`` line up above the `prelex-counter` definition along with all of the other function definitions in the code.  This required a bit of re-giggering of the code since `pred-env-empty` was no longer in scope.

You can see the update and the diff in this [gist](https://gist.github.com/akeep/37db1369a87bfdf7cd55aaba2619446c).

So, I'm going to see if I'm going to pull some of the changes in the gist into my current copy of this and give it a try running the full Chez Scheme bully tests to see if I can get the error to occur, if it survives the bully tests I'm going to be tempted to merge this, though I may also ask for a second opinion on where we ended up.

Thanks again!

gus-massa:
I still don't understand why this cause the unexpected `prelex`, but I'll try to read this again in the morning.

Meanwhile, I tried to reread the current code one more time and I found a small error, and I pushed a fix. I'd normally squash it into my old commit, but since you are already working on top of it it may cause some problems. It's a tiny change, so it's even possible to copy it by hand to your current version.

IIUC this bug I found is unrelated, but sometimes optimization bugs reappear in strange places. I tried to build a test case, but it's difficult because I think that most of the times the expression is reduced by cp0.

akeep:
@gus-massa I think you should be able to test this by disabling cp0, since we now run the cptypes pass as part of the block with cpletrec so that it runs regardless of if cp0 is on or not.  (Actually, I made this change for exactly this reason.)

I've also discussed this pass a bit with @dybvig, partially to verify my assumptions around how we are using the prelex in the cptypes pass.  As part of our conversation he asked if you had tried writing this up in a paper.  We both agree that this can be helpful in checking the reasoning on the path, and I think it would help @dybvig and I to better understand exactly what is going on in the pass.  I feel like I have a decent general understanding, but there are still some parts, particularly around the type environment maintenance, that I don't fully grok yet.

As I mentioned I'm traveling this weekend (in fact I was supposed to be on a plane right now, but we're currently delayed do to maintenance issues with the plane), so it will probably be Wednesday evening before I get much time to look at this again.

gus-massa:
I have some notes, I'll give them another read and send them. They are mostly extended comments of the code, and cover mostly the use of the fxmap and prelex, to be sure that there are no quadratic part hiding in the corner cases. 

But the notes don't discuss the easy parts, like the shotcircuiting of `(begin <bottom> e2)` or `(if <bottom> e2 e3)`. Also IIRC the type lattice is not covered, because I think it has no tricky parts, most of the details of the conversions are in the big comment at the beginning of the file.

gus-massa:
I just pushed a small fix of the error messages, in particular it uses `$impoops` instead of `error` because it should never happen. I'd squash this commit too, but as you have the fixed the thread problems in a not public branch, it may be easier to just copy these changes there (perhaps with a better text).

gus-massa:
I made a few small updates. I squashed my two minor last commits with a few new changes. If you have already made some changes I can try to rearrange them with this changes.

### Fix cptypes in multi-thread version

This is essentially the [gist](https://gist.github.com/akeep/37db1369a87bfdf7cd55aaba2619446c) made by @akeep . I removed the error checks and the debug messages.

With this change many of the auxiliary functions are defined inside the pass. I'm not sure if this is a good idea, because I'm not sure if they will be slower than a version of the functions that are defined outside the pass.

Also, IIUC the cache of the types of the signatures in [L551](https://github.com/gus-massa/ChezScheme/blob/639f32cfc6f462fe9492d13b6fd246cb6be1df3f/s/cptypes.ss#L551) and [L605](https://github.com/gus-massa/ChezScheme/blob/639f32cfc6f462fe9492d13b6fd246cb6be1df3f/s/cptypes.ss#L605) will get a different instance in each time the pass is used, so it will be slower than having a common cache for all the instances that the pass is called. (Are hash-table thread safe?)

### Fix call case with rest argument in cptypes 

This has the two small fixes, and a special case to make cptypes detect that `(list x y z)` is a `pair`, i.e. it is not `null`. This is useful because cp0 may reduce 

`((lambda (x . rest) ...) a b c)` ==> `((lambda (x rest) ...) a (list b c))`

and this make the type detection of `rest` slightly more consistent.

I also added some test here. More on this in the next commit.

### more cptypes

To make the test pass, I had to add a call to cptypes in two files. 

First I made the calls look at the `(enable-type-recovery)` parameter so they can be disabled.

Second I added a call to cptypes when cp0 is not called, to make it consistent with other compilation ways. The problem is that after using this in the test I don't like it.

I made two version of the test. 

* One disables cp0 and use the fact that cptypes will be called anyway

* The other uses `(run-cp0 (lambda (cp0 x) (#3$cptypes x)))`

I think that it's more intuitive that disabling cp0 will disable all (or most) of the optimization, including cptypes. It's not intuitive that cptypes is called anyway. My guess is that the normal users will not be interested in testing the behavior of cptypes alone, so in the rare case that any of them are interested, they can use `#3%cptype`. So my proposal is to remove the second part of this commit and also remove all the cptypes from the cp0-was-not-called branches in the compiler. (I didn't write this yet.) 

(I like the flag `(enable-type-recovery)`, it may be useful to isolate bugs.)

(An additional question: The cp0-was-not-called branches have a call to `$cpletrec`. How bad is that someone redefines `run-cp0` as `(run-cp0 (lambda (cp0 x) (#3$cptypes (cp0 x))))`? Because it will set the flag that `cp0` was called and `$cpletrec` will not be the last pass. (IIUC the properties of the `letrec`s that `$cpletrec` try to enforce are also enforced later in cpnanopass, so this is not a problem.))

Test in version with bootfiles [here](https://travis-ci.org/gus-massa/ChezScheme/builds/410624476).

gus-massa:
I added a small commit changing how `enable-type-recovery` work, so 

    (run-cp0 (lambda (cp0 x) x))

will disable the traditional cp0 pass and also the cptypes pass. I think it's more intuitive this way. More details in my previous comment. This is based in an old version, so it needs some tricks to pass the test. See [branch](https://github.com/gus-massa/ChezScheme/tree/17-5-Types-Pass-Z) / [tests](https://travis-ci.org/gus-massa/ChezScheme/builds/498334437).

I also made a rebased version in case it's useful. It needs new bootfiles, but they don't have to be the very latest one. See [branch](https://github.com/gus-massa/ChezScheme/tree/17-5-Types-Pass-ZR) / [tests](https://travis-ci.org/gus-massa/ChezScheme/builds/499446059).

gus-massa:
This is just a small fix in one of the signatures I modified. I didn't want to forget about it. You can ignore this   commit for now.

mflatt:

ufo5260987423:
Thank for your hard working!

Hi, I just noticed this PR and I'm very interested in it. Would there be any benchmark tests to show the power of this improvements? And I noticed that the "enable-type-recovery" tag would be mentioned in CSUG. When would it be released, in version 10.0.0 or later?

Thank for your hard working again!

PR #198, opened 2017-08-16
added support for command line options -O0, -O1, -O2, -O3

marcomaggi:
Some users may feel more "at home" with command line switches similar to those of the common C compilers.

There are no changes to the TeX documentation (I have no familiarity with submodules).


marcomaggi:
I still suggest this, but I am closing it because I am now using a [command line program](https://github.com/marcomaggi/ChezCompiler) as interface to the compiler functions.

PR #205, opened 2017-08-30
Refactor windows installers and fix vcredist

kleinpa:
- Separate 64 and 32 bit MSIs
- Add wix bundle to combine MSIs into single installer
- Use basic wix UI
- Use merge modules to install vcredist package
- Add script to locate vcredist merge modules
- Fix installer status text by adding WixUI_ErrorProgressText ref

Merge modules seems to be the preferred method for installing the redistributable package, which should decrease the package size and speed up installation. This commit also addresses a bug where the installer does not work when VC redistributable is not already installed.

The 32 and 64 bit components were split into separate MSIs because Windows Installer only officially supports packages containing a single architecture. A package bundle is then created containing both MSIs so only a single file needs to be distributed.

There seemed to be no trivial way to get the path to the vcredist merge modules bundled with visual studio so I added a script (locate-vcredist.bat) that handles this. It will need to be updated in the future for compatibility with newer VS platform toolsets.

Some paths and name were changed, here's a summary:

32-bit install path: C:\Program Files (x86)\Chez Scheme 9.4.1\
64-bit install path: C:\Program Files\Chez Scheme 9.4.1\
installer file name: ChezScheme.exe
installer product name: Chez Scheme 9.4.1 x64

### DIFF
--- a/LOG
+++ b/LOG
@@ -566,4 +566,13 @@
     mats/Mf-a6nt, mats/Mf-ta6nt, mats/ftype.ms
 - added support for building Windows installs with Bash/WSL
     wininstall/Makefile, candle.bat, light.bat
-- added support for building with Visual Studio 2017's BuildTools in c/vs.bat
+- added support for building with Visual Studio 2017's BuildTools
+    c/vs.bat
+- check for git before using to get submodules
+    configure
+- fixed windows installer failure when vcredist is not preinstalled by
+  using the vcredist merge module, split the 32 and 64 bit MSIs and
+  added a wix bundle to combine the MSIs into a single exe installer,
+  added a batch script for locating Visual Studio's vcredist merge
+  modules, updated installer paths and names.
+    wininstall/*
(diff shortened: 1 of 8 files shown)


kleinpa:
This ended up being much more complicated that what I originally set out to build. Let me know if anyone has any suggestions to trim it down.

- Does anyone actually want to choose their own install directory? Dropping WixUI_Advanced in favor of something with fewer options would cut out an ugly workaround.
- Dropping a registry key pointing to the install location could make it easier for for applications to find scheme.exe.
- Will anyone want multiple different versions of Chez installed at one time? How should we handle this? Currently only one 32-bit and one 64-bit installation are allowed and installing a newer version is treated as an upgrade.
- I haven't tested this Bash/WSL since I don't have that environment handy. Does it work?

burgerrg:
Nice work! It builds fine with Bash/WSL. Please squash the two commits, and I'll merge it into master.

PR #212, opened 2017-09-10
Fix signature of port-length and file-length

gus-massa:
The documentation says: 

> For binary ports, the length is always an exact nonnegative integer byte count. For textual ports, the representation of a length is unspecified; it may not be an exact nonnegative integer and, even if it is, it may not represent either a byte or character count.


dybvig:
I believe the documentation is wrong and these procedures do deal in exact nonnegative integers. The restriction on `set-port-length!` doesn't seem to be enforced for generic ports, but it does seem to be enforced otherwise.

Issue #137, opened 2016-12-21
Use structs directly as foreign-procedure parameters / return values.


ecraven:
I am trying to define a foreign-procedure that directly returns a struct. As an example, from the C version of SFML:
````
typedef struct
{
    sfInt64 microseconds;
} sfTime;

sfTime sfSeconds(float amount);
````

I have tried the following, but get the error shown:
````
(define-ftype sfTime
  (struct (microseconds integer-64)))
(define sf:seconds (foreign-procedure "sfSeconds" (float) sfTime))
=> Exception: invalid (non-base) foreign-procedure result ftype sfTime
````

I hope I haven't missed how to do this while reading the csug. Is there any way to achieve this short of writing a custom C function that wraps the struct value in a pointer?

evilbinary:
Meet the same question for passing directly struct to a function.

dybvig:
Passing and returning structs (rather than pointers to structs) is not presently supported.  To call a function that requires struct arguments or returns struct values, you'd need to create an appropriate C wrapper.  Or you'd need to extend `foreign-procedure` and `foreign-callable` to handle struct arguments and return values.

evilbinary:
How to extend foreign-procedure and foreign-callable? I need passing struct and return the struct value.Could you show detail?

dybvig:
If you want to tackle this, you'll need to study and understand the code for handling `foreign-procedure` and `foreign-callable` in syntax.ss, cpnanopass.ss, and (for starters) one of the machine-dependent files like x86.ss.  You'll also need to learn how the calling conventions for a particular target (architecture and O/S) dictate structs be passed and returned.  For documentation on calling conventions, look for an _Application Binary Interface_ description, at least on Unix-based systems.  Also look at the output of 'cc -S testfile.c", where testfile.c contains various C examples passing and returning structs with and without other arguments.  You can use these examples for testing later.  Once you understand how the compiler handles non-struct values as well as the calling conventions for struct values, you'll be ready to start modifying the expander and compiler.  Once you get struct values working and well tested (via tests added to foreign.ms) on one target, the others should be relatively straightforward.  You should be able to get help here or on the mailing list from @akeep, @cfrisz, and me with specific questions if you get stuck on particular issues.  While you'll be charting new ground, we all have experience with the compiler's handling of `foreign-procedure` and `foreign-callable`.

evilbinary:
[...]

PR #213, opened 2017-09-16
Foreign-procedure struct args and results

mflatt:
This patch is a first cut at adding support struct arguments and results when calling foreign procedure or generating a foreign-callable wrapper for a Scheme procedure (as in #137). The initial implementation is only for x86 and x86_64 on Mac OS, but if the approach makes sense, then I can fill in more.

Syntactically, the patch adds support for `(& ftype)` as a foreign-procedure argument or result. On the Scheme side, the representation for `(& ftype)` is the same as for `(* ftype)`. On the foreign side, the content of a pointer for `(& ftype)` is passed as a value or received as a result, instead of passing or receiving the pointer.

For example,

```
typedef struct {
  int x, y;
} pt;

pt get_pt() {
  pt p = { 10, 20 };
  return p;
}
```

can be called with

```
(define-ftype pt
  (struct
    [x int]
    [y int]))

(define get_pt
  (foreign-procedure (foreign-entry "get_pt")
                     () (& pt)))

```

The `ftype` in `(& ftype)` doesn't have to be a struct type. For example, `(& int)` could be used in place of an `int` argument or result, in which case the Scheme side must supply or receive an fpointer to an integer. Unions, packed structs, etc., are all meant to work.

On the Scheme side, a result received as `(& ftype)` or a callable's `(& ftype)` argument is allocated with `foreign-alloc` and must be explicitly freed. Similarly, a callable must return a `(& ftype)` result that is allocated, and it will be freed on the foreign side after the value is copied out. An argument `(& ftype)` is not freed by the call, currently. I would be happier with a way to allocate memory as GCable and avoid the explicit frees, but I don't think an fpointer can wrap GC-allocated bytes.

Some details:

* I'm unsure about the way that `Scopy-argument` is called to copy the content of a `(& ftype)` callable argument. Is it ok to pass an argument via `%ts`? Is the saved-argument list correct?

* For callables that return `(& ftype)`, the way that tail-calling a `S_call_...` variant achieves the right result often involves some trickery – The `S_call_indirect_copy_three_chars` variant is the biggest hack – and I'm not sure it's better than implementing the ABI directly in each backend. But continuing with the `S_call_...` approach was the path of least resistance.


gwatt:
@mflatt You could put in a ftype&-guardian in the collector. It already closes forgotten ports using a guardian: https://github.com/cisco/ChezScheme/blob/master/s/7.ss#L675

mflatt:
@gwatt Thanks for the suggestion! If I understand, you mean registering the fptr that was created to wrap the address to free the `malloc`ed address. I worry about that indirection, since the same address might be wrapped by a different fptr to cast it, and then a programmer would have to take special care to keep the original fptr live as long as the new fptr is used. Also, finalization doesn't currently compose well (when a finalization action for some other object needs to use a finalized fptr), although #190 is an effort to address that kind of problem.

mflatt:
The latest patch includes tests and doc updates. It supports x86 and x86_64 on Linux, Windows, Mac OS, and probably other Unix variants.

mflatt:
The latest version is a good candidate for merging.

Based on a suggestion from Kent, the latest implementation treats a `(& ftype)` result as an extra `(* ftype)` initial argument, which eliminates the implicit calls to `malloc` and `free`.

 * When a program invokes a foreign procedure with a `(& ftype)` the program allocates and provides a `(* ftype)` argument before the normal arguments, and the result of the foreign procedure is unspecified.

 * A callable with a `(& ftype)` result should accept an extra `(* ftype)` first argument to receive the result, and the callable's direct result is ignored. A callable's `(& ftype)` arguments correspond to `(* ftype)` arguments that are valid only during the dynamic extent of the callable invocation (because the associated address is on the stack or provided by the C caller).

The key to avoiding implicit `malloc` and `free` with the new protocol was to get rid of the tail-called `S_call_...` zoo, replacing it with a non-tail call to `S_call_one_value` or `S_call_any_values` plus a direct implementation of return values in each backend.

All backends are implemented: x86, x86_64, ppc32, and arm32.

mflatt:
Two bugs:

 * On i3nt, I must have overlooked `__stdcall` test failures. There's a `20` on line 2710 of "x86.ss" that should be `init-stack-offset`.

 * On ppc32le, the pointer to fill for a `(& ....)` result is stashed too deeply on the stack. The `frame-size` on lines 2392, 2416, and 2439 should be the original `frame-size` from line 2383 instead of the shadowing binding on line 2385.

These bugs are fixed in #267. I could move the repairs to here, but it may make more sense to just consider the patches together.

dybvig:
I haven't taken enough time to look at this yet, but generally speaking, this is a great addition.  Rather than wait until I have had a chance to finish going over the code in detail, I thought I'd provide a quick bit of additional feedback now.

The feedback regards the syntax of foreign-procedure / foreign-callable argument and return types.  For ftype pointers (rather than ftype values), Andy and I used `(* `_ftype-name_`)` to leave open the possibility of later using just _ftype-name_ for ftype values.  Do you think this would work, and is it a good idea?

Base ftype names like `int` that conflict with primitive argument and return type names would be a problem, so we'd probably have to disallow their use except by way of an alias.  In other words, we'd probably have to go with the primitive type rather than the (base) ftype in case of a name conflict.

mflatt:
If I understand what you mean: I think it would work to treat an ftype name for a struct or union differently from other ftype names. I don't think it would work to treat every ftype name as implicitly using the current `&`. For example, if I define `(define-ftype NSUInteger unsigned-long)` then when specifying `NSUInteger` as a function argument type, I would want to pass a number to the foreign procedure, not a pointer to memory containing an integer. I see no problem with `NSUInteger` or base types if only struct/union ftype names are treated as implying a `&`.

Racket's FFI does treat a struct/union argument/result type as implicitly having `&`, but it also manages storage differently for `&` results or callable arguments. Racket's FFI allocates GC-managed memory for those structs/unions, so a function type that returns a struct/union still corresponds to a function that returns one (allocated) result. I've implemented Racket's FFI on top of Chez Scheme's FFI using bytevectors, but it requires copying (and, at least for now, object locking and some worse trickery).

With the memory-management strategy in this patch, `&` seems more uniform and a better signal to readers in a foreign-procedure specification. For example, the `&` in `(foreign-procedure "reel_one_in" () (& fish_t))` helps highlight that the resulting function takes one argument and returns no results. In contrast, `(foreign-procedure "reel_one_in" () fish_t)` looks like a function that takes zero arguments and returns one result — but that wouldn't be true if `fish_t` is defined as a struct or union type that implies `&`.

Either way works for me, but unless there's a better overall design point, I recommend keeping the explicit `&` as a reminder of memory-management obligations.

ecraven:
This would be very helpful for me, is there any way in which I can help with the work to get this into mainline chez?

dybvig:
I see what you mean about `&` in the return type highlighting the interface change.

While I can see why you might want plain `NSUInteger` to indicate that the argument is or can be a simple integer rather than an ftype pointer, I don't know why you'd expect it.  When I declare an argument or return type to be an ftype, regardless of any decorations, I would expect the the argument or return value to be an ftype pointer.

In any case, I don't have a strong feeling one way or another about the change in syntax, so I'm okay leaving it as is.

dybvig:
@ecraven, I'm in the process of reviewing the implementation and expect to be able to pull it when I'm done, possibly with minor corrections.  It wouldn't hurt to have another pair of eyes on it, though.

ecraven:
I just cloned this repo and ran `./configure && make`, it stopped with the following error:
````
...
(if `cmp -s ../boot/a6le/equates.h ../boot/a6le/equates.h.bak`;\
          then mv -f ../boot/a6le/equates.h.bak ../boot/a6le/equates.h;\
          else rm -f ../boot/a6le/equates.h.bak; fi)
Exception in $fasl-file-equal?: gensym comparison failed while comparing ../boot/a6le/sbb and ../boot/a6le/petite.boot within fasl entry 931
make[4]: *** [Mf-base:283: checkboot] Error 255
make[3]: *** [Mf-base:162: allx] Error 2
make[2]: *** [Mf-base:174: bootstrap] Error 2
make[1]: *** [Makefile:21: build] Error 2
make: *** [Makefile:19: build] Error 2
````
Am I building incorrectly? I have a "normal" chez-scheme installed system-wide, is there some interference?

**Comment by mflatt** [comment]
@ecraven That check fails because the patch doesn't include updated bootfiles. But you probably have a working build, and running `make` again will normally succeed.

**Comment by ecraven** [comment]
Thanks, that worked. I'll try wrapping some of CSFML, to see how things work ;)

**Comment by ecraven** [comment]
@mflatt thanks for the hint, works fine, no problems with a few struct parameters and return values in CSFML so far.

**Comment by dybvig** [comment]
Now that the `foreign-callable` code object performs a non-tail call rather than a tail call to the C call helper, it has become necessary to lock the code object before the code runs and not unlock it until after it returns, due to the possibility of a garbage collection in the invoked Scheme routine.

For example, the mat on line 2011 of foreign.ms needs to call `lock-object` on Fcons, something like:

```
(equal?
  (let ()
    (define Sinvoke2
      (foreign-procedure "Sinvoke2"
        (scheme-object scheme-object iptr)
        scheme-object))
    (define Fcons
      (foreign-callable
        (lambda (x y)
          (collect)
          (let ([ls (make-list 20000 #\z)])
            (collect)
            (collect)
            (collect)
            (collect)
            (collect)
            (cons (length ls) (cons x y))))
        (scheme-object iptr)
        scheme-object))
    (define (go)
      (dynamic-wind
        (lambda () (lock-object Fcons))
        (lambda () (Sinvoke2 Fcons 4 5))
        (lambda () (unlock-object Fcons))))
    (go))
  '(20000 4 . 5))
```

It's a bit tricky to get this to fail reliably without the call to `lock-object`, but the following variant results in an invalid memory reference for me on both 32- and 64-bit Linux.

```
(load-shared-object "../mats/foreign1.so")
(parameterize ([collect-request-handler void])
  (define Sinvoke2
    (eval
      '(foreign-procedure "Sinvoke2"
         (scheme-object scheme-object iptr)
         scheme-object)))
  (define Fcons
    (eval
      `(foreign-callable
         (lambda (x y)
           (collect)
           (let ([ls (map (lambda (x) (make-vector 200 x)) (make-list 100))])
             (collect)
             (collect)
             (collect)
             (collect)
             (cons (length ls) (cons x y))))
         (scheme-object iptr)
         scheme-object)))
  (let loop ()
    (let ([got (dynamic-wind
                 (lambda () void #;(lock-object Fcons))
                 (lambda () (Sinvoke2 Fcons 4 5))
                 (lambda () void #;(unlock-object Fcons)))]
          [expected '(100 4 . 5)])
      (unless (equal? got expected)
        (errorf #f "expected ~s, got ~s\n" expected got)))
    (loop)))
```

**Comment by mflatt** [comment]
That makes sense. I had not worked out that it previously could make sense to pass a callable without locking it.

Should the necessary locking be added to the callable's implementation, instead of adjusting the requirements on programs (and the test case)?

**Comment by dybvig** [comment]
That's a good idea.  It should probably be unlocked automatically even when a call returns to some lower point on the stack, i.e., when there is a continuation invocation that causes an outer call into Scheme to return.  This might involve storing something extra in the cchain.

**Comment by mflatt** [comment]
The new commit is intended to fix locking for the non-tail call to the C helper.

**Comment by dybvig** [comment]
Unfortunately, this ends up unlocking only one of the two code objects, resulting not only in one of the code objects being locked forever but also building up the locked object list and greatly slowing collection..  The basic problem is that the extra (dummy) CCHAIN element needs to be on top of the one stored in the frame, rather than below.  Fixing this also obviates the call to Sunlock_object in S_call_help.  The patch below seems to fix the problem.

```
*** ../../c/schlib.c	2018-03-09 16:55:22.000000000 -0500
--- ./schlib.c	2018-03-09 18:42:10.000000000 -0500
***************
*** 219,238 ****
      jb = CREATEJMPBUF();
      if (jb == NULL)
        S_error_abort("unable to allocate memory for jump buffer");
! 
      if (lock_ts) {
        /* Lock a code object passed in TS, which is a more immediate
           caller whose return address is on the C stack */
        callable_code = TS(tc);
        Slock_object(callable_code);
!       CCHAIN(tc) = Scons(Scons(jb, TS(tc)), Scons(Scons(NULL, code), CCHAIN(tc)));
      } else {
        callable_code = NULL;
-       CCHAIN(tc) = Scons(Scons(jb, code), CCHAIN(tc));
      }
  
-     FRAME(tc, -1) = CCHAIN(tc);
- 
      switch (SETJMP(jb)) {
          case 0: /* first time */
              S_generic_invoke(tc, S_G.invoke_code_object);
--- 219,235 ----
      jb = CREATEJMPBUF();
      if (jb == NULL)
        S_error_abort("unable to allocate memory for jump buffer");
!     FRAME(tc, -1) = CCHAIN(tc) = Scons(Scons(jb, code), CCHAIN(tc));
      if (lock_ts) {
        /* Lock a code object passed in TS, which is a more immediate
           caller whose return address is on the C stack */
        callable_code = TS(tc);
        Slock_object(callable_code);
!       CCHAIN(tc) = Scons(Scons(NULL, TS(tc)), CCHAIN(tc));
      } else {
        callable_code = NULL;
      }
  
      switch (SETJMP(jb)) {
          case 0: /* first time */
              S_generic_invoke(tc, S_G.invoke_code_object);
***************
*** 245,255 ****
          case 1: { /* normal return */
              ptr yp = CCHAIN(tc);
              FREEJMPBUF(CAAR(yp));
!             if (lock_ts) {
!               Sunlock_object(callable_code);
!               CCHAIN(tc) = Scdr(Scdr(yp));
!             } else
!               CCHAIN(tc) = Scdr(yp);
              break;
          }
          default:
--- 242,248 ----
          case 1: { /* normal return */
              ptr yp = CCHAIN(tc);
              FREEJMPBUF(CAAR(yp));
!             CCHAIN(tc) = Scdr(yp);
              break;
          }
          default:
```

dybvig:
We should probably expose an foreign entry for the mats to use to check the number of locked objects and check it in various places during the mat run.

mflatt:
Repair applied - thanks!

dybvig:
_We should probably expose an foreign entry for the mats to use to check the number of locked objects and check it in various places during the mat run._

Well, of course we already had something that returns a list of the locked objects to test `lock-object` and `unlock-object` in misc.ms, so I've put it to use in foreign.ms.

This turned up an interesting issue that predates this pull request but that this pull request exacerbates slightly, which is that erroring out of a procedure called via foreign-callable leaves the topmost `CCHAIN` element(s) on the `CCHAIN` stack and leaves the code object(s) locked.  I have not yet determined how to address this.

I don't see any other problems with the pull request.  @mflatt, I would prefer that you update LOG and release_notes.stex so I can do a clean pull.  I'm happy to rebuild the boot files, or you can also include them.  Thanks for all of your efforts on this.

mflatt:
Rebased and squashed, including repairs for the two bugs noted on Jan 22.

dybvig:
The last change exposed another problem whereby `do_error` could pass `LONGJMP` a null jump buffer from one of the dummy `CCHAIN` entries, so I modified the `CCHAIN` entry to include a pair of code objects and eliminated the dummy entries.  `do_error` should probably be invoking a more specific entry rather than whatever happens to be first in the `CCHAIN`, and I'll look at that when I try to address the issue mentioned above in which `CCHAIN` entries are not getting cleaned up when a `foreign-callable` routine raises an exception.

PR #217, opened 2017-10-09
Fix cp0 reduction of fx[+-*]/carry and their signatures

gus-massa:
For example, 

    `(fx+/carry (quote 1) (quote 1) (quote 1))

was reduced to

    `(values 3 0)

but to use the nanopass language correctly it should be reduced to

    `(values (quote 3) (quote 0))


akeep:
Looks good, thanks!

PR #218, opened 2017-10-10
small change in documentation, ouch to shout OUCH

jasonhemann:
The documentation currently shows

    (format "~@:(~a~)" "Ouch!")  =>  "ouch!"

Chez 9.4.1 instead outputs "OUCH!"

    Chez Scheme Version 9.4.1
    Copyright 1984-2017 Cisco Systems, Inc.
    
    > (format "~@:(~a~)" "Ouch!")
    "OUCH!"

Chez 8.4 also yields this same behavior.

### DIFF
--- a/csug/io.stex
+++ b/csug/io.stex
@@ -2729,7 +2729,7 @@ format string with the \scheme{~@(} and \scheme{~)} directives.
 
 \schemedisplay
 (format "~@(~r~)" 2599) ;=>  "Two thousand five hundred ninety-nine"
-(format "~@:(~a~)" "Ouch!") ;=>  "ouch!"
+(format "~@:(~a~)" "Ouch!") ;=>  "OUCH!"
 \endschemedisplay
 
 \noindent


burgerrg:
The Common Lisp HyperSpec http://www.lispworks.com/documentation/HyperSpec/Body/22_cha.htm indicates that ~@:( is bogus. Shouldn't the example be (format "~@(~a~)" "ouch!") => "Ouch!"

akeep:
Yeah, Chez treats ``~@:`` and ``~:@`` the same in the format code, so I think this is doing the:

> ``~:@(`` converts every lowercase character to the corresponding uppercase character.

version of the conversion from the Common Lisp formatted output specification.

jasonhemann:
@burgerrg , are you suggesting adjusting the example in the docs, rather than the output?

burgerrg:
Yes, I think the example should be updated to be in sync with the format specification.

jasonhemann:
Updated to @burgerrg 's suggested example.

dybvig:
As @akeep said, colon and at-sign can come in either order, so `~:@(---)` and `~@:(---)` are equivalent.  I believe this is true generally for Common Lisp format.  If so I assume the referenced document just doesn't bother showing both orders and the original fix should be okay.

PR #233, opened 2017-11-11
add `box-cas!` and `vector-cas!`

mflatt:
I think this change is mostly straightforward, but is `(set! ,(make-live-info) ,%rax (asm ,info ,asm-locked-cmpxchg ...))` on line 934 of "x86_64.ss" (and similar for "x86.ss") the best way to reflect that CMPXCHG may modify the RAX register?


dybvig:
I don't think it's necessary since the preceding instruction kills `%rax` via `urax`, so I would remove the assignment and leave the RHS.

If it were necessary, I think you'd want it to set `urax` instead of `%rax`, since all assignments other than unspillable assignments need to be visible prior to live analysis.

mflatt:
Updated.

mflatt:
Rebased.

csug/objects.stex
```diff
@@ -1,4 +1,4 @@
-% Copyright 2005-2017 Cisco Systems, Inc.
+<% Copyright 2005-2017 Cisco Systems, Inc.
```
burgerrg: The initial < looks like a typo.

mflatt: Fixed - thanks!

burgerrg:
@mflatt, how do I compile this branch? I get `Error in $lookup-library-entry: entry ~s uninitialized
(150)`

mflatt:
@burgerrg I use the usual(?) cross-compile steps:
```
set M=<platform>
export M
set ORIG=<current-version-build-dir>
export ORIG
make
```
But I always feel like I must be missing some better and simpler way.

burgerrg:
Thanks, that worked! If you have a recent Chez Scheme installed, you can do this from the particular workarea:
```
(cd s && make all patchfile=patch Scheme=scheme SCHEMEHEAPDIRS=/usr/lib/csv%v/%m:/usr/local/lib/csv%v/%m)
make
```

burgerrg:
Thanks for this pull request!

PR #237, opened 2017-12-01
arity & JIT-like redirection

mflatt:
This PR can be interpreted as "what's the best way to achieve my goal?", but I found it easiest to phrase the question as an possible patch.

This patch adds `make-arity-wrapper-procedure` which takes a procedure and an arity mask and returns a wrapper procedure with the same behavior as the given one, but returning the given arity mask for `procedure-arity-mask`. The `set-arity-warpper-procedure!` operation takes a wrapper and updates its behavior to be a different procedure.

One use of `make-arity-wrapper-procedure` is to make procedures produced by `interpret` return the right result for `procedure-arity-mask`. The patch updates `interpret` in that way, but it uses a system variant of `make-arity-wrapper-procedure` to hide the wrappers from user-level programs.

A similar use of wrappers is for a user program to implement just-in-time compilation (or other transformation) of procedures with a minimal overhead for calling the procedures, especially after JIT compilation. That's why the wrapper produced by `make-arity-wrapper-procedure` is mutable. Although similar functionality could be achieved with `(lambda args (apply (unbox b) args))`, a wrapper procedure can dispatch to another procedure even more quickly and without allocating a list for the arguments.

Mutable wrappers in this patch are implemented by mutating the closure, so there's a new flag on code objects to tell the GC to put closures over the code into the impure space.

Finally, wrappers have a mutable "data" field to store extra information that a JIT compiler/transformer might need to associate with the procedure.

The arity attached to a wrapper is used only by `procedure-arity-mask`. There's no checking that the procedure in the wrapper has an actual mask that is consistent with the claimed one or to limit the wrapper's application to the claimed arity. That kind of checking is left up to the wrapped procedure.


dybvig:
I haven't looked at the PR code yet, so these initial comments are based only on your description.

I don't think it's sufficient to put mutable closures in space impure unless the new code object is always no newer (i.e., not in a newer generation) than the closure.  If the new code object can sometimes be newer, we'll probably need a space dedicated to mutable closures and to implement the dirty write / dirty scan protocols for them ala other mutable pointer-containing objects.  The problem with space impure for a mutable is that the code pointer in the first field always (by design) looks like a fixnum, so the scan for younger objects would not find it.

Is there also a bit in the code object to say whether the closure is mutable?  If not, and all closures are mutable, there's an issue with direct calls, which don't indirect through the code pointer and will continue to run the original code.

mflatt:
The code pointer in a mutable closure is never changed, so it is always older.

But it sounds like it works (or doesn't) for reasons that are more subtle than intended. I'll try putting the mutable parts in a pair and see whether I can detect enough of a difference from the extra indirection to bother with mutable closures.

mflatt:
No surprise: When indirecting through a pair, I can see a small difference when calling `(make-arity-wrapper-procedure sub1 2 #f)` 10 million times, but the difference is well in the noise for my actual application.

I'm reluctant to give up the more compact representation and more direct dispatch, but I don't have a strong argument for it. I've updated the patch here with a note in the GC, while an alternative with pairs is in the `arity-wrapper2` branch of mflatt/ChezScheme.

mflatt:
Rebased.

"JIT" mode is the primary use for this patchin Racket-on-Chez, and JIT mode looks more and more like a dead end. Still, `make-arity-wrapper-procedure` itself worked well, and it may enable future experiments or JIT-like alternatives, so I still think this addition is worthwhile.

akeep:
Closing this out, for now, since the Racket fork has moved on.

PR #239, opened 2017-12-04
spelling

barak:


burgerrg:
Thank you for finding and fixing these typos!

PR #240, opened 2017-12-05
fix bytevector-[u/s]16-native-set! signature

gus-massa:
This is a fix of the `bytevector-[u/s]16-native-set!` signature, and also `bytevector-sint-set!`.

The signature of `bytevector-[u/s]32-native-set!` is correct, so it's not necessary to fix it.


akeep:
Looks good, thanks!

PR #241, opened 2017-12-06
keeping procedure srclocs without other inspector info

mflatt:
Procedure source locations are useful for error backtraces at a much lower cost (while losing some precision) compared to keeping full inspector information.

This patch adds `generate-procedure-source-information`, which defaults to #f and only applies when `generate-inspector-information` is #f. If `generate-procedure-source-information` is #t, then a source location is preserved with a code object even when  `generate-inspector-information` is #f.

### DIFF
--- a/LOG
+++ b/LOG
@@ -971,3 +971,7 @@
   bootstrap failures after small changes like the recent change to
   procedure names, so we don't have to rebuild the boot files as often.
     Mf-base
+- add generate-procedure-source-information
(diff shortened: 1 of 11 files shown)


mflatt:
Rebased.

For now, Racket-on-Chez's default compilation mode still uses `(generate-inspector-information #f)` and `(generate-procedure-source-information #t)`, and that combination has worked as intended — preserving the same kind of debugging information as traditional Racket with the same minimal cost.

burgerrg:
Thanks for this pull request!

PR #242, opened 2017-12-09
fix a few signatures

gus-massa:
In the case of `unget-{u8/char}`, the value can be an also an `eof`, so I have to change the type to `ptr`.

Also, it is strange that:

    vector-fill! -> (vector)
    bytevector-fill! -> (void)
    string-fill! -> (void)
    fxvector-fill -> (void)
    substring-fill! -> (string)


akeep:
Both ``vector-fill!`` and ``substring-fill!`` return an unspecified value (which is why they are marked void).  The fact that they happen to return a ``vector`` and ``string`` (respectively) is accidental, and I think if there is code that depends upon it, we should probably change that code, but I think the signatures are actually correct in ``primdata.ss``

If these two signatures are leading to bugs, let me know and I'll take a look into what is causing the problem.

The rest to these changes look good.

akeep:
Just a quick update on this.  I've committed a change to ``vector-fill!`` and ``substring-fill!`` that corrects their output to return ``void`` in keeping with their documented behavior.

gus-massa:
I changed the mat of `substring-fill!`, because it used that the return value is the string.

I removed the changes of the signatures of `vector-fill!` and `substring-fill!`, but I added a change to the signature of `fxvector-fill!` (that I mysteriously missed).

akeep:
@gus-massa Good catch on the ``substring-fill!`` test, I'm not sure why my own testing didn't turn that up.

I think the change to ``unget-{u8,char}`` makes sense.  Perhaps at some point we'll have to extend the little type system to express ``u8`` + ``eof-object``, ``char`` + ``eof-object``, etc.  since this does come up relatively frequently in the I/O functions, if it turns out to be profitable anyway.

gus-massa:
While I was writing the fix, I was also thinking that `char+eof` may be a good idea. (I'm not sure about the name.) . It's similar to `maybe-char`.

PR #245, opened 2017-12-21
avoid quadratics in live-variable information

mflatt:
I'm ambivalent about this patch. It makes compilation possible for a real program, but even the improved compilation time is too long, so the real program needs a different approach. Then again, I may run into problem one day where the solution seems worthwhile.

The program that has something like the following shape, where there are N calls each with an average of N/2 live variables (for N = 10000):

```
 (define vars
   (let loop ([i 10000])
     (cond
      [(zero? i) '()]
      [else (cons (gensym) (loop (sub1 i)))])))

 (time
  (begin
   (compile
    `(lambda ,vars
       ,@(map (lambda (v) `(,v)) vars)))
   (void)))
```

One part of the patch adjusts the compiler to keep call-live variables in tree form, since they're already collected that way. Memoizing reductions on the tree preserves sharing from the construction of liver-variable trees. That changes allows the above example to compile in around 6 seconds with around 300 MB on my machine, instead of 25 seconds and 1.8 GB. My actual program is larger; without the change, it requires more memory than I have, but it takes 284 seconds and 8 GB with the change.

Another part of the patch changes the run-time representation(!) of live-pointer masks to use the same tree representation as the compiler. That allows sharing preserved during compilation to turn into sharing in the generated code, so the code has closer to linear size. For example, the code generated above is roughly 6 MB instead of 28 MB. For my actual example, I didn't measure code size, but the difference in peak memory use (since the live-pointer masks are computed during compilation) is 2.5 GB vs 8 GB, with around 50 seconds saved.

A third part is related to register allocation and avoiding re-traversals of poison variables when they're irrelevant. Memozing a filter of non-poison variables helps with the real program, and it saves another 130 seconds, although the expense of higher memory use, ending up at 3 GB instead 2.5 GB.

In case it's of interest, here's the expression generated from the real program: http://www.cs.utah.edu/~mflatt/tmp/i3.ss.gz


mflatt:
Unless there's particular interest in this change, I'm ready to close the PR.

As noted above, a different solution (falling back to interpretation for the outer shell of a module) was needed to avoid some overlong compilation times in Racket-on-Chez. With that more general solution in place, I don't have a real program where this patch is needed.

PR #246, opened 2017-12-22
Check that first argument of map in cp02 expansion

gus-massa:
The first commit adds a checks that first argument of `map` is a procedure in cp02 expansion to raise the same error that the non expanded version

This fix an error in the tests in Windows of the mat "4.ms", [line 1082-1083](https://github.com/cisco/ChezScheme/blob/master/mats/4.ms#L1082-L1083) and [line 1713-L1714](https://github.com/cisco/ChezScheme/blob/master/mats/4.ms#L1713-L1714)

    (error? ; nonprocedure
      (map 3 '(a b c)))
    ;[...]
    (error?
      (for-each 3 '(a b c)))

The summary is something like

    -------- o=0 cp0=t --------
    [...]
    270c273
    < 4.mo:Expected error in mat map: "map: 3 is not a procedure".
    ---
    > 4.mo:Expected error in mat map: "attempt to apply non-procedure 3".
    340c343
    < 4.mo:Expected error in mat for-each: "for-each: 3 is not a procedure".
    ---
    > 4.mo:Expected error in mat for-each: "attempt to apply non-procedure 3".

I really don't understand why the test in Linux don't detect this. (Reading some error messages, I guess that it's a difference between scheme.boot and petite.boot, but I'm not sure if this even make sense. I'm very confused.)

----

The second commit is a small improvement. It avoids building the result list in a map that is called for effect.

### DIFF
--- a/LOG
+++ b/LOG
@@ -766,4 +766,6 @@
 - Check that first argument of map is a procedure in cp02 expansion
   to raise the same error that the non expanded version
     cp0.ss
+- avoid building the result list in a map that is called for effect
+    cp0.ss
 
--- a/LOG
+++ b/LOG
@@ -763,4 +763,7 @@
   to use define-who instead of repeating 'substring-fill! in all the error
   messages.
     5_4.ss, 5_6.ss
+- Check that first argument of map is a procedure in cp02 expansion
+  to raise the same error that the non expanded version
+    cp0.ss
 
--- a/s/cp0.ss
+++ b/s/cp0.ss
@@ -3588,6 +3588,9 @@
           (lambda (?p ?ls ?ls* lvl map? ctxt sc wd name moi)
             ; (map/for-each proc (list a11 a12 ... a1m) (list a21 a22 ... a2m) ... (list an1 an2 ... anm)) =>
             ;   (let ([p proc])
+            ;     (if (procedure? p)
+            ;         (void)
+            ;         ($oops 'map/for-each "~s is not a procedure" p))
             ;     (let ([t11 a11] ... [t1m a1m])
             ;       ...
             ;         (let ([tn1 an1] ... [tnm anm])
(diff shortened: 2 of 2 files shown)


akeep:
The difference here is actually an expected one.  Without consulting with Kent, I'm not exactly sure why this difference exists when, as you've shown it doesn't need to, other than avoiding special casing every little think in cp0, which may be why he didn't go down this road in the first place.

The reason it doesn't fail on Linux (and shouldn't on Windows either) is that the way the expected errors are handled, there is a main expected errors file, and then there are patches for different settings.  Linux is (and Windows should) patch the expected errors with the corrected set.

So, that is the explanation.  Before we decide we want to go down this road, I'd like to double check with Kent.  I know this is not the only instance of differences when cp0 is turned off, and doing this across the board might require a fair amount of special casing.

dybvig:
There's a better reason to include the procedure check at optimize level 2, which is that `(#2%map e ---)` should raise an exception if x is not a procedure even when the code containing the call is run at optimize-level 3.  So this seems like a good change to me.  Cosmetically, I'd prefer the then and else parts of the inserted if expression be indented under the test, and the second and third make-seq arguments be indented two spaces rather than under the first argument.  I'm guessing the patch files need to be rebuilt to reflect the change in error messages.  I'll let Andy do a detailed check of the changes if he wants and leave it to him to pull.

gus-massa:
I fixed the indentation. I can rebase them if necessary.

But I don't understand your comment. Is it posible to expand `(#2%map p list ...)` with the current version and then run the expanded version in a way that it doesn't raise an error if `p` is not a procedure?

On the other hand, `(#2%map 3 '())` is still reduced to `'()` without any error. I didn't fix it. Is this an error? The documentation is not 100% clear. This is caused by a special case that is checkd before calling `inline-lists`.

akeep:
@gus-massa The problem occurs when you have an optimize level 2 call to ``map`` in a program that is otherwise compiled at optimize level 3.

You can see this by setting Chez Scheme to optimize level 3, turning off compile-interpret-simple, and then executing ``#2%map``:

```
% ../bin/scheme
Chez Scheme Version 9.5.1
Copyright 1984-2017 Cisco Systems, Inc.

> (optimize-level 3)
> (compile-interpret-simple #f)
> (#2%map 3 '(1 2 3))
```

This is happening because ``(#2%map 3 '(1 2 3))`` is expanding into ``(#2%list (3 1) (3 2) (3 3))`` and when the calls to ``3`` are compiled at optimize-level 3, the procedure checking code is not being written at the call sites:

```
> (expand/optimize '(#2%map 3 '(1 2 3)))
(#2%list (3 1) (3 2) (3 3))
```

Your change ends up addressing this (even though that this not what it was intended to fix).  It is possible that we could make this change stricter and only do this when the compiler is currently running at optimize level 3 and we are optimizing the optimize level 2 version of ``map``.

gus-massa:
OK, I see the problem. Should I change anything?

I'm still worried about `(map 7 '())`. In my current version with cptypes https://github.com/cisco/ChezScheme/pull/192 the expansion can produce an error, for example in

    (define (f p b)
     (unbox b)
     (map p (if (box? b) '() '(1 2 3)))
     (list p (procedure? p)))
    (f 7 (box 0)) ;==> (7 #t)

(The example is strange to make sure that the reductions of cp0 and cptypes are applied in the right order to get the error.)

One solution is to fix the expansion of `map` in cp0 and other solution is to change the signature of `map`. (And other similar functions.)

akeep:
So, I think this works as expected.  When you compile your example program at optimize-level 2, you get an error, when you compile it at optimize-level 3, you do not, this is exactly the same behavior you would see if cp0 was not being run, which is the goal.

The change you made should change the behavior when you are compiling a program at optimize-level 3, but you've explicitly called the optimize-level 2 version of ``map``:

```
% ../bin/scheme
Chez Scheme Version 9.5.1
Copyright 1984-2017 Cisco Systems, Inc.

> (optimize-level 3)
> (define (f p b)
   (unbox b)
   (#2%map p (if (box? b) '() '(1 2 3)))
   (list p (procedure? p)))
> (f 7 (box 0)) ;==> (7 #t)
Exception in map: 7 is not a procedure
```

Anyway, I think these changes look okay, I'm in the process of testing through things and making sure we've got the expected error and patch files set up correctly.  I'll also be adding some tests to the mats for this.

PR #249, opened 2017-12-28
even? and odd?: reject exceptional flonums, since they're not integers

mflatt:
Make `(even? +inf.0)`, etc., raise an exception.


akeep:
Hey @mflatt,

These changes look good.  I think we should consider using ``who`` in place of the explicit ``'odd?``, ``'even?``, ``'flodd?``, and ``'fleven?``.  Your changes make use of these correctly, but I noticed when looking through the context of the expected error diffs that ``(flodd? 3.2)`` is raising the error: ``Exception in odd?: 3.2 is not an integer`` in stead of ``Exception in flodd?: 3.2 is not an integer`` because it explicitly has ``'odd?`` where it should have ``'flodd?`` or ``who`` in the ``nonifnteger-integer`` error call.

mflatt:
Now with more `who`.

akeep:
Thanks for the changes and the ``who``!

PR #254, opened 2018-01-05
license to live dangerously when recompiling libraries

mflatt:
The first patch makes the `generate-id` function used by the expander externally configurable. Also, `define-record-type` uses the configurable `generate-id`. As shown in the docs, setting `current-generate-id` while compiling a library can enable recompilation such that code previously compiled against the library will continue to work with the new compilation –— in many cases, at least.

The second patch adds `expand-omit-library-invocations`. Setting that parameter to true leaves out any `$invoke-library` calls from the expansion of a form that refers to library exports. Omitting `$invoke-library` can be a minor time and space savings when lots of code fragments are generated and the relevant libraries are known to be present and invoked.

A programmer who uses these hooks takes on the burden of knowing enough about the Chez Scheme expander and compiler to know when a configuration will produce sensible/intended results. Specifically, the `current-generate-id` hook seems to enable a better performance–development trade-off for Racket-on-Chez work. The `expand-omit-library-invocations` option is less significant for Racket-on-Chez, but I rest easier knowing that a bit of unnecessary work is out of the way.

### DIFF
--- a/LOG
+++ b/LOG
@@ -971,3 +971,6 @@
   bootstrap failures after small changes like the recent change to
   procedure names, so we don't have to rebuild the boot files as often.
     Mf-base
+- add current-generate-id and expand-omit-library-invocations, which can be
+  useful for avoiding library recompilation and redundant invocation checks
+    syntax.ss, record.ss, primdata.ss, front.ss, misc.ms, system.stex
--- a/csug/system.stex
+++ b/csug/system.stex
@@ -2119,6 +2119,136 @@ loaded from source, when it is compiled via \scheme{compile-file},
 and when a compiled version of the file is loaded via \scheme{load}
 or \scheme{visit}.
 
+%----------------------------------------------------------------------------
+\entryheader
+\formdef{current-generate-id}{\categorythreadparameter}{current-generate-id}
+\listlibraries
+\endentryheader
+
+\noindent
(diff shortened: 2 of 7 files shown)


jltaylor-us:
Would it make more sense for such inherently dangerous functions to be available only via the "internal" chez names, e.g. `#%$current-generate-id`, and not included in the user documentation?  Seems a shame to omit the docs you wrote, though... maybe there should be a separate doc for internal-only functionality.

mflatt:
@jltaylor-us I don't have a strong opinion, so I'll leave that question to the Chez Scheme maintainers.  

I tried to err on the side of issuing caution when describing `current-generate-id`. In principle, it's hopeless, much as the top level is hopeless – and perhaps it's an extension of that same hopelessness. But I'm finding that the use of `current-generate-id` in the draft docs behaves in an understandable way in practice.

mflatt:
P.S. – Add
```
(define current-generate-id (make-parameter (lambda (s) (gensym (symbol->string s)))))
```
 to "patch.ss" to build via cross-compile.

akeep:
Understanding the potential usefulness of this, I think it is a bit too dangerous for anything other than special-case uses.  If we decide to revisit, it is nice to know the Racket fork of Chez will still have it.

PR #262, opened 2018-01-16
add `load-compiled-from-port`

mflatt:
The main change in this patch adds `load-compiled-from-port`, which roughly complements `compile-to-port`. Racket-on-Chez uses `compile-to-port` to turn a set of "linklets" for a module into a set of object-code fragments; it now uses `load-compiled-from-port` to run those fragments. (Previously, Racket-on-Chez relied on `fasl-read` and a rough imitation of the internals of `do-load-binary`.)

A second change is related, although distinct from `load-compiled-from-port`: `Sregister_boot_file_fd` gives a custom Chez Scheme startup the ability to load a boot file from a place that is not directly identifiable as a path, such as the middle of an executable file – especially when using the implicit EOF at the end of a compressed stream. After booting, a custom executable can use `load-compiled-from-port` to load additional code from something other than a stand-alone file.

### DIFF
--- a/csug/foreign.stex
+++ b/csug/foreign.stex
@@ -2804,6 +2804,7 @@ program.
 \cfunction{void}{Sscheme_init}{void (*\var{abnormal}_\var{exit})(void)}
 \cfunction{void}{Sset_verbose}{int \var{v}}
 \cfunction{void}{Sregister_boot_file}{const char *\var{name}}
+\cfunction{void}{Sregister_boot_file_fd}{const char *\var{name}, int \var{fd}}
 \cfunction{void}{Sbuild_heap}{const char *\var{exec}, void (*\var{custom}_\var{init})(void)}
 \cfunction{void}{Senable_expeditor}{const char *\var{history}_\var{file}}
 \cfunction{void}{Sretain_static_relocation}{void}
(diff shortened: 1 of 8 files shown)


csug/foreign.stex
```diff
@@ -2752,9 +2753,11 @@ for subsequently registered boot files.
 
 \scheme{Sregister_boot_file} searches for
 the named boot file and
-register it for loading.
-The file is opened but not loaded until the heap is built via
-\scheme{Sbuild_heap}.
+register it for loading, while \scheme{Sregister_boot_file_fd}
+provides a specific boot file as a file descriptor.
+When only a boot file name is provided, the is opened but not loaded until the heap is built via
```
burgerrg: "the is opened" => "the file is opened"

mflatt: Fixed (and rebased) - thanks!

mflatt:
Rebased.

These changes are still essential for Racket-on-Chez.

mats/7.ms
```diff
@@ -112,6 +112,14 @@
     "6\n")
  )
 
+(mat load-compiled-from-port
+  (begin
+   (define-values (o get) (open-bytevector-output-port))
+   (compile-to-port '((define lcfp1 'worked) 'loaded) o)
+   (equal? 'loaded (load-compiled-from-port (open-bytevector-input-port (get)))))
```
burgerrg: Suggestion: use `eq?` instead of `equal?` to compare symbols.

PR #266, opened 2018-01-23
add `collect-rendezvous`

mflatt:
This patch exposes the internal thread rezdezvous for GC to applications through a `collect-rendezvous` function.

For many purposes, `collect-rendezvous` is a variant of `collect` that works when multiple threads are active. More precisely, while `collect` actually collects, the `collect-rendezvous` function invokes the collect-request handler (in an unspecified thread).

### DIFF
--- a/LOG
+++ b/LOG
@@ -899,3 +899,5 @@
     Mf-install.in
 - standalone export form now handles (import import-spec ...)
     8.ms, syntax.ss, release_notes.stex
+- add collect-rendezvous
+    prim.c, 7.ss, primdata.ss, 7.ms, smgmt.stex, release_notes.stex
--- a/c/prim.c
+++ b/c/prim.c
@@ -177,6 +177,7 @@ void S_prim_init() {
     Sforeign_symbol("(cs)enable_object_counts", (void *)S_enable_object_counts);
     Sforeign_symbol("(cs)set_enable_object_counts", (void *)S_set_enable_object_counts);
     Sforeign_symbol("(cs)object_counts", (void *)S_object_counts);
+    Sforeign_symbol("(cs)fire_collector", (void *)S_fire_collector);
 }
 
 static void s_instantiate_code_object() {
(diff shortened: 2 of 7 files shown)


burgerrg:
How difficult would it be to modify `collect` so that it performs the rendezvous automatically when multiple threads are active?

mflatt:
@burgerrg The issue is performing the actual garbage collection versus calling the collect-request handler. For my purposes, it's important that `collect-rendezvous` calls the collect-request handler, the same as an automatically triggered collection does. Changing `collect` to call the collect-request handler, meanwhile, would break things, since a collect-request handler normally calls `collect` to do its work.

burgerrg:
Thanks for the clarification!

I think you can avoid calling the C code in `S_fire_collector` by grabbing the tc-mutex, setting `$collect-request-pending` to true, releasing the tc-mutex, and calling `$collect-rendezvous`. You don't need to set `tc-something-pending` because you're immediately calling `$collect-rendezvous`.

burgerrg:
```
(define collect-rendezvous
  (lambda ()
    (if-feature pthreads
      (with-tc-mutex (set! $collect-request-pending #t))
      (set! $collect-request-pending #t))
    ($collect-rendezvous)))
```

mflatt:
I may not understand the suggestion, but calling `S_fire_collector` seems necessary to tell all active threads to check `$collect-request-pending`. Otherwise, each thread's timer callback will skip the check.

More generally, I don't think `collect-rendezvous` can do less than an implicitly triggered GC. (But it does a little more, by calling `$collect-rendezvous`, to ensure that a GC is actually performed before `collect-rendezvous` returns.)

burgerrg:
Thanks, I missed the setting of something-pending in all other threads. We could do that in Scheme, too.

burgerrg:
I think it's wise to call the C code and not duplicate it in Scheme.

burgerrg:
Thanks, I think this is a useful addition to Chez Scheme. Would you please rebase onto the current master and add a section for `collect-rendezvous` in the release notes?

mflatt:
Done.

I see that I had previously neglected to commit a repair to the test. There's now a loop to make sure that a test thread has terminated before trying plain `collect` again.

burgerrg:
Thanks, I had wondered about that race condition. Too bad Chez Scheme doesn't have a `thread-join` procedure.

PR #267, opened 2018-01-23
foreign-procedure thread-activation control

mflatt:
Threads are often used in event-handling libraries like the one shown in the documentation for `foreign-callable`. Prior to this patch, activating and deactivating a thread to make an event-based pattern thread-friendly required glue C code to call `Sdeactivate_thread` and `Sactivate_thread`.

This patch adds `__thread` as a convention modifier for `foreign-procedure` and `foreign-callable`:

 * On a foreign procedure, `__thread` causes the calling thread to be deactivated (at the last moment, after arguments to the procedure are unpacked), and the thread is re-activated when the foreign procedure returns.

 * On a foreign callable, `__thread` causes the calling thread to be activated, and the activation state is reverted when the callable returns. "Reverted" has no effect if the thread was already activated, but "reverted" can mean "destroyed" if the thread had not been registered when the callable was invoked.

As an example, using a pair of `__thread` annotations in the `event-loop` example for `foreign-callable` makes the example play nicely with threads, instead of causing threads to block when they try to GC. A `__thread` annotation on a callable can also make the callable work when invoked by an unspecified thread, instead of requiring that that the thread is first activated.

This patch builds on #213. The changes could stand alone, but since the modifications overlap and since Racket needs both patches, it seems easiest to combine the changes, and it may make sense to review them together.

### DIFF
--- a/LOG
+++ b/LOG
@@ -895,3 +895,9 @@
 - reworked the S_call_help/S_return CCHAIN handling to fix a bug in which
   the signal handler could trip over the NULL jumpbuf in a CCHAIN record.  
     schlib.c
+- add a __thread convention for foreign procedures and callables
+  to automate thread [de]activation
+    syntax.ss, ftype.ss, x86.ss, x86_64.ss, ppc32.ss,
+    cmacros.ss, base-lang.ss, np-languages.ss, cprep.ss
+    thread.c, prim.c, externs.h, foreign.stex, release_notes.stex,
+    mats/Mf-t*, foreign.ms, foreign4.c
--- a/LOG
+++ b/LOG
@@ -895,9 +895,10 @@
 - reworked the S_call_help/S_return CCHAIN handling to fix a bug in which
   the signal handler could trip over the NULL jumpbuf in a CCHAIN record.  
     schlib.c
-- add a __thread convention for foreign procedures and callables
+- add a __collect_safe convention for foreign procedures and callables
   to automate thread [de]activation
     syntax.ss, ftype.ss, x86.ss, x86_64.ss, ppc32.ss,
-    cmacros.ss, base-lang.ss, np-languages.ss, cprep.ss
+    cmacros.ss, base-lang.ss, np-languages.ss, cprep.ss, cpcommonize.ss,
+    cp0.ss, cpcheck.ss, cpvalid.ss, interpret.ss, cpletrec.ss,
     thread.c, prim.c, externs.h, foreign.stex, release_notes.stex,
     mats/Mf-t*, foreign.ms, foreign4.c
--- a/boot/a6le/equates.h
+++ b/boot/a6le/equates.h
@@ -87,8 +87,8 @@ typedef unsigned long U64;
 #define bytevector_length_factor 0x8
 #define bytevector_length_offset 0x3
 #define bytevector_type_disp 0x1
-#define c_entry_name_vector #(thread-context get-thread-context handle-apply-overflood handle-docall-error handle-overflow handle-overflood handle-nonprocedure-symbol thread-list split-and-resize raw-collect-cond raw-tc-mutex handle-values-error handle-mvlet-error handle-arg-error foreign-entry install-library-entry get-more-room scan-remembered-set instantiate-code-object Sreturn Scall-one-result Scall-any-results)
-#define c_entry_vector_size 0x16
+#define c_entry_name_vector #(thread-context get-thread-context handle-apply-overflood handle-docall-error handle-overflow handle-overflood handle-nonprocedure-symbol thread-list split-and-resize raw-collect-cond raw-tc-mutex activate-thread deactivate-thread unactivate-thread handle-values-error handle-mvlet-error handle-arg-error foreign-entry install-library-entry get-more-room scan-remembered-set instantiate-code-object Sreturn Scall-one-result Scall-any-results)
+#define c_entry_vector_size 0x19
 #define cached_stack_link_disp 0x8
 #define cached_stack_size_disp 0x0
 #define card_offset_bits 0x9
@@ -681,6 +681,9 @@ typedef unsigned long U64;
 #define typedef_u8 "unsigned char"
 #define typedef_uptr "unsigned long int"
 #define typemod 0x8
+#define unactivate_mode_deactivate 0x1
+#define unactivate_mode_destroy 0x2
+#define unactivate_mode_noop 0x0
 #define unaligned_floats 1
 #define unaligned_integers 1
 #define underflow_limit 0x80
@@ -695,28 +698,31 @@ typedef unsigned long U64;
 #define wchar_bits 0x20
 
 /* constants from declare-c-entries */
-#define CENTRY_Scall_any_results 21
-#define CENTRY_Scall_one_result 20
-#define CENTRY_Sreturn 19
-#define CENTRY_foreign_entry 14
-#define CENTRY_get_more_room 16
+#define CENTRY_Scall_any_results 24
+#define CENTRY_Scall_one_result 23
+#define CENTRY_Sreturn 22
+#define CENTRY_activate_thread 11
+#define CENTRY_deactivate_thread 12
+#define CENTRY_foreign_entry 17
+#define CENTRY_get_more_room 19
 #define CENTRY_get_thread_context 1
 #define CENTRY_handle_apply_overflood 2
-#define CENTRY_handle_arg_error 13
+#define CENTRY_handle_arg_error 16
 #define CENTRY_handle_docall_error 3
-#define CENTRY_handle_mvlet_error 12
+#define CENTRY_handle_mvlet_error 15
 #define CENTRY_handle_nonprocedure_symbol 6
 #define CENTRY_handle_overflood 5
 #define CENTRY_handle_overflow 4
-#define CENTRY_handle_values_error 11
-#define CENTRY_install_library_entry 15
-#define CENTRY_instantiate_code_object 18
+#define CENTRY_handle_values_error 14
+#define CENTRY_install_library_entry 18
+#define CENTRY_instantiate_code_object 21
 #define CENTRY_raw_collect_cond 9
 #define CENTRY_raw_tc_mutex 10
-#define CENTRY_scan_remembered_set 17
+#define CENTRY_scan_remembered_set 20
 #define CENTRY_split_and_resize 8
 #define CENTRY_thread_context 0
 #define CENTRY_thread_list 7
+#define CENTRY_unactivate_thread 13
 
 /* displacements for records */
 #define eq_hashtable_rtd_disp 1
--- a/boot/a6nt/equates.h
+++ b/boot/a6nt/equates.h
@@ -87,8 +87,8 @@ typedef unsigned long long U64;
 #define bytevector_length_factor 0x8
 #define bytevector_length_offset 0x3
 #define bytevector_type_disp 0x1
-#define c_entry_name_vector #(thread-context get-thread-context handle-apply-overflood handle-docall-error handle-overflow handle-overflood handle-nonprocedure-symbol thread-list split-and-resize raw-collect-cond raw-tc-mutex handle-values-error handle-mvlet-error handle-arg-error foreign-entry install-library-entry get-more-room scan-remembered-set instantiate-code-object Sreturn Scall-one-result Scall-any-results)
-#define c_entry_vector_size 0x16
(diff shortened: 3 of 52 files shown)


Saigut:
Nice feature!

ecraven:
I get compile errors:
````
gcc  -m64 -msse2 -Wpointer-arith -Wall -Wextra -Werror -Wno-implicit-fallthrough -O2 -D_REENTRANT -pthread  -c -DX86_64 -I../boot/ta6le -I../zlib foreign.c
gcc  -m64 -msse2 -Wpointer-arith -Wall -Wextra -Werror -Wno-implicit-fallthrough -O2 -D_REENTRANT -pthread  -c -DX86_64 -I../boot/ta6le -I../zlib prim.c
prim.c: In function ‘create_c_entry_vector’:
prim.c:127:21: error: ‘CENTRY_activate_thread’ undeclared (first use in this function); did you mean ‘S_activate_thread’?
     install_c_entry(CENTRY_activate_thread, proc2ptr(S_activate_thread));
                     ^~~~~~~~~~~~~~~~~~~~~~
                     S_activate_thread
prim.c:127:21: note: each undeclared identifier is reported only once for each function it appears in
prim.c:128:21: error: ‘CENTRY_deactivate_thread’ undeclared (first use in this function); did you mean ‘CENTRY_activate_thread’?
     install_c_entry(CENTRY_deactivate_thread, proc2ptr(Sdeactivate_thread));
                     ^~~~~~~~~~~~~~~~~~~~~~~~
                     CENTRY_activate_thread
prim.c:129:21: error: ‘CENTRY_unactivate_thread’ undeclared (first use in this function); did you mean ‘CENTRY_deactivate_thread’?
     install_c_entry(CENTRY_unactivate_thread, proc2ptr(S_unactivate_thread));
                     ^~~~~~~~~~~~~~~~~~~~~~~~
                     CENTRY_deactivate_thread
make[2]: *** [Makefile:29: prim.o] Error 1
make[1]: *** [Makefile:20: build] Error 2
make: *** [Makefile:19: build] Error 2

````


````
> gcc --version                                                    
gcc (GCC) 7.2.1 20171224
Copyright (C) 2017 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
````

**Comment by mflatt** [comment]
@ecraven The problem is that the boot files aren't updated, where "boot files" includes headers that have changed and are needed to compile the C part. So, you need to cross-compile to try it out.

To build for a platform `$M` (e.g., "ta6osx"), if you have an existing build for platform `$M` in `$ORIG`, then cross-compile with

```
  ./configure
  (cd $M/c && make)
  (cd $M/s && make allx)
```

**Comment by mflatt** [comment]
Rebased, but the cross-compile steps are still needed to build.

**Comment by burgerrg** [comment]
For 32-bit Windows, I assume __thread uses the __cdecl calling convention? Is there any convenient way to be able to specify both __thread and __stdcall or __com? (I'm not writing new code against 32-bit Windows any more, so this is not important to me, but I ask in order to make you aware of the issue.)

**Comment by mflatt** [comment]
@burgerrg Yes, the patch generalizes the syntax of `foreign-prcedure`, etc., to accept multiple "conventions", although `__thread` is the only one that is allowed in combination with any other.

**Comment by mflatt** [comment]
@burgerrg I started to reply "I mean that `__thread` already works with `__stdcall` and `__com`", but then I realized that I forgot to add tests. So, of course, it didn't work for all combinations.

I've added tests and repairs so that `__thread` works with `__stdcall` and `__com`.

**Comment by burgerrg** [comment]
Great, thanks!

**Comment by ecraven** [comment]
I've tested this with running my SLIME swank and a web server in parallel, and it seems to solve the problems I had very well!
Is there any way I can help with getting this merged?

**Comment by burgerrg** [comment]
@dybvig, are you planning to review this pull request? I haven't looked at it in detail because I know you've recently inspected changes related to the foreign-procedure interface and may already be planning to review it.

**Comment by akeep** [comment]
Hey @burgerrg,

@dybvig and I talked a bit about it last week, and I volunteered to pick this up, but I've not had time to review the code in any level of detail, and I probably won't have a chance until next weekend.

One of the concerns we had about this is that any Scheme object passed as an argument to this function would need to be locked before the thread is deactivated, otherwise there is a chance that the collector might start between when the thread deactivation occurs and the argument is passed to the underlying C function.  A ``foreign-procedure`` marked with ``__thread`` would need to be at least somewhat responsible for this, because in some cases (like ``string`` arguments which are copied to a ``bytevector`` and the pointer into the ``bytevector`` data is passed to the underlying function) the programmer never actually has their hands on the object to lock.

We also discussed a little bit having a different name for this which is a little more descriptive.  Something like ``__thread_safe`` or ``__collector_safe``, since the property we are trying to express is that it is safe for the collector to run while the underlying C call is blocking.

**Comment by burgerrg** [comment]
This does sound dangerous. What would you think if it locked all Scheme objects before deactivating and unlocked them after re-activating? That's what my custom C code does in Swish.

**Comment by ecraven** [comment]
Isn't this only relevant for strings? I'm passing bytevectors to recv and send, which I lock myself, so that isn't so bad. Are there any other datatypes except for strings that get copied internally?

**Comment by akeep** [comment]
@ecraven Yes, I think the automatic object creation/copying is only for strings, so if you have non-string arguments and you are locking them already, you should be good to go.  The procedure code for the scheme wrapper is also created, but I think that should already be automatically locked before the C call.  I'd have to double check to be sure.

@burgerrg Yes, @dybvig and I talked about having it auto-lock/unlock. I think where we landed on this, was that it might be nice to lock/unlock automatically but provide a syntax for specifying that an argument is already locked (though now that  I think about it you'd probably want that at the call site for the foreign procedure and we were thinking of this at the ``foreign-procedure`` definition site).

**Comment by mflatt** [comment]
@akeep Checking whether I understand the issue with strings: Even before this PR, that can be an issue with a foreign function that invokes a callbable and is also passed a converted string (which it might access after the callable), right?

**Comment by akeep** [comment]
@mflatt Yes, if the C function holds onto the pointer into the ``bytevector`` across the call to the ``foreign-callable``.

**Comment by ecraven** [comment]
Wouldn't it suffice to automatically lock the temporary bytevector for each __thread foreign, and unlock it afterwards? Everything else would stay the same as right now, wouldn't it?

**Comment by akeep** [comment]
@ecraven The idea with the automatic locking is to allow programmers to use this feature without needing to lock the arguments themselves.  So, you wouldn't need to lock the ``bytevector`` around the call yourself, you could just let the ``foreign-procedure`` code take care of it for you.  I think @mflatt's struct parameter change added automatic locking and unlocking of ``foreign-callable`` arguments for the same reason.

**Comment by ecraven** [comment]
i'm not against automatic locking, but if anyone is, I think automatic locking of only the temporary string bytevectors should be sufficient. Personally, I'd welcome automatic locking of all relevant objects.

**Comment by mflatt** [comment]
@akeep FWIW, there was no new argument locking for callables – only some new locking of the callable itself, since a return address to the callable stays on the stack more than it used to.

I'm not opposed to extra automatic locking, but given that the rule is already "don't do that" when a foreign procedure might invoke a callable, "don't do that" seems equally workable with `__thread` (or its replacement). That is, string types could be considered conveniences that only work in simple cases, and using `__thread` creates obligations for a caller in the same way as using callables does.

**Comment by dybvig** [comment]
It's analogous though not quite the same.  In this case there's no safe place for the user program to lock the temporary bytevector.  The Scheme caller can't lock it, since it doesn't have access to it, and the C callee can't lock it, since it doesn't get control until after the thread is deactivated.  So the implementation should either reject string arguments when `__thread` is used or lock the temporary bytevector, regardless of whether we extend automatic (though possibly optional) locking to all arguments that need to be locked.

I think the `__thread` convention is a great idea, by the way.

**Comment by burgerrg** [comment]
Whereas C code called without `__thread` can safely use unlocked Scheme objects, C code with the `__thread` calling convention cannot safely use an unlocked Scheme object. Consequently, it seems appropriate that the system handles the locking and unlocking automatically of all `string` and `ptr` types. If a particular application cannot tolerate the overhead, it can use `uptr` and `#%$object-address` to pass the address of a locked Scheme object to the C code.

**Comment by mflatt** [comment]
In case it helps, I've added a commit to change `__thread` to `__collect_safe` and to report an error if a string type is misused with `__collect_safe` as an argument (for foreign procedures) or result (for foreign callables). I could squash the commits if that's helpful.

I didn't muster enough interest in automatic locking to try that direction myself, so `__collect_safe` as implemented is an assertion of safety, not a request to make the call safe.

**Comment by burgerrg** [comment]
Thanks, I like the name change and disallowing string types. Not having automatic locking/unlocking is fine with me. Thank you for proposing this important feature addition. With it I'll be able to avoid wrapper code when making blocking calls that are collector safe.

**Comment by akeep** [comment]
@mflatt, I've had a chance to look through the code a bit, and for the most part I think it looks good, but there are few small things I think we should fix.

I ran the mats for ti3osx, ta6osx, a6osx, and i3osx.  I noticed a couple of things in the tests you added with the non-threaded version of the mats.  The added error producing tests at the start of the ``collect-safe`` mat in ``foriegn.ms`` do not raise an exception, because the ``__collect_safe`` convention is being discarded before we check the arguments.  Even though the ``__collect_safe`` convention doesn't have any effect, I think we should consider having it raise the same errors in both threaded and non-threaded versions.

I also noticed, in running the mats, that the checking is done after the original types have been filter and this leads to a bit of a confusing error message when ``string`` or ``wstring`` types are used.  For example:

```
(foreign-procedure __collect_safe "unknown" (string) void)
```

produces

```
Exception in foreign-procedure: utf-8 argument not allowed with __collect_safe procedure
```

instead of

```
Exception in foreign-procedure: string argument not allowed with __collect_safe procedure
```

It might be nice to do this checking before the filtering is done, or pass ``$make-foreign-callable`` and ``$make-foreign-procedure`` the mapping between the original and filtered types, so it can use th original type when it raises errors.

On the language side, I noticed you changed ``convention?`` in ``base-langs.ss`` instead of changing the ``Lsrc`` language.  I'd recommend changing the ``Expr``s:

```
(foreign conv name e (arg-type* ...) result-type)
(fcallable conv e (arg-type* ...) result-type)
```
to

```
(foreign (conv ...) name e (arg-type* ...) result-type)
(fcallable (conv ...) e (arg-type* ...) result-type)
```

Admittedly, this is a bit of a pain and you'll need to bump ``Lsrc``'s nongenerative id. I ran through the exercise myself to see how bad the change was, and it does require going through: ``syntax.ss``, ``cpnanopass.ss``, ``cpcommonize.ss``, ``base-lang.ss``, ``cp0.ss``, ``cpcheck.ss``, ``cpvalid.ss``, ``interpret.ss``, ``cpletrec.ss``, and ``cprep.ss``.  The changes are pretty small and are caught by during compilation, so it was a pretty straightforward change.  The exercise proved to be worth it (for me at least), because I found a place in ``cpcommonize.ss`` where we were doing an ``eq?`` comparison on the conventions, which now needs to be ``equal?``.

I see you also caught a spurious ``convention?`` definition in ``np-languages.ss`` that is not used anywhere, we should probably remove that while we're making these changes.

I noticed you've also updated the documentation (thanks!) and I've got a couple of comments here too:

1. We might want to call out the usefulness of ``__collect_safe`` for blocking I/O calls, since before you added this, we didn't have a good story around that short of wrapping the C calls in a C function that deactivated and reactivated.
1. I think we should consider listing out the types that are not allowed instead of saying ... ``utf-8`` through ``utf-32be``, for consistency in other places where we refer to lists of types in the prose.
1. I wasn't sure wha was meant by "Parameter types other than ``scheme-object`` through ``utf-32be`` are converted to equivalent foreign representations and consequently they can be retained indefinitely in foreign variables...  it seems like ``utf-32be`` is not something that can be retained indefinitely, did you mean ``u32*``?  In this context I'm also wondering if we should be talking about what types need to be locked when using ``__collect_safe`` (for instance ``scheme-object``, ``u8*``, ``u16*`` etc. need to be locked, while ``fixnum``, ``integer-8``, ``integer-16`` etc., do not need to be locked.

Finally, I took a look through the ``x86_64.ss`` version of the assembler changes to support this, and things there look okay to me, though I'm admittedly a little less fresh on this code than @dybvig is.

Thanks again for the pull request!  Also, I like ``__collect_safe`` better than ``__thread_safe`` for the ``foreign-procedure``, I think it is a bit more descriptive.

mflatt:
@akeep Thanks for the feedback. It will be a week or two before I can work on this more. Could you send/post your changes to use `(conv ...)` instead of making `conv` mean a sequence? (It seemed tedious when I tried that originally, but maybe I didn't pick the right approach.)

akeep:
@mflatt No problem.  I've pushed my changes on a branch at [pr/267](https://github.com/cisco/ChezScheme/tree/pr/267) with those changes.

mflatt:
@akeep Thanks for the `conv` conversion. I've added your commit here, and further adjusted some `conv` names to `conv*`.

For the error when using a string type with `__collect_safe`, reporting the original type name would mean adding to the function RTD and more, which seems heavyweight for an error that will be rarely seen. I made the error message slightly more generic by always using "string" instead of naming a specific type. Meanwhile, reporting is now consistent from a non-threaded build.

Finally, I updated the docs as you suggested.

akeep:
Thanks for the changes @mflatt, I've gone ahead and rebuilt the boot files and merged your changes!

PR #269, opened 2018-02-10
Remove special case for (#2%map p '()) in cp0

gus-massa:
When  `(#2%map p '())` is reduced in cp0 to `'()`, the expression of p is discarded without a check that it is a procedure. But the signature of `map` in primdata.ss is `((procedure list list...) ->(list))`.

The problem is that in the [PR of cptypes #192](https://github.com/cisco/ChezScheme/pull/192) uses the signature to mark `p` as a procedure after this expression, assuming that otherwise `map` would have raise an error. 

The code to show the problem should be a mix of some reductions of cp0 and some reduction of cptypes:

    (define (f p b)
     (unbox b)
     (map p (if (box? b) '() '(1 2 3)))
     (list p (procedure? p)))
    
    (f 7 (box 0)) ;==> 
    ; ==> (7 #f) in the current version of Chez Scheme
    ; ==> [error] in Chez Schem + this patch
    ; ==> (7 #t) in Chez Scheme + cptypes [wrong]

I think that the best behavior is to raise an error. This patch removes one of the reductions in cp0 so the other reduction is triggered and an error is raised.

But I think it's also ok to not raise an error, but in this case the signature of map should be weaker, and allow the first argument to be any pointer. The patch to do this is in my [alternative branch](https://github.com/gus-massa/ChezScheme/commits/18-2-Map-Null-Alternative).

But I think that the result `(list p (procedure? p))` --> `(7 #t)` is wrong, and it can cause further problems in cptypes in more complex and realistic examples.

### DIFF
--- a/LOG
+++ b/LOG
@@ -879,3 +879,7 @@
   $verify-ftype-address if the address expression is a call to
   ftype-pointer-address.
     ftype.ss
+- Remove special case for (#2%map p '()) in cp0
+  so the reduced version checks that p is a procedure.
+  Also make the same change for #2%for-each.
+    cp0.ss, 4.ms
--- a/mats/4.ms
+++ b/mats/4.ms
@@ -1080,6 +1080,10 @@
  ; make sure compiler doesn't bomb w/two few args
   (procedure? (lambda (x) (map x)))
   (error? ; nonprocedure
+    (map 3 '()))
+  (error? ; nonprocedure
+    (map 3 '() '()))
+  (error? ; nonprocedure
     (map 3 '(a b c)))
   (error? ; nonprocedure
     (parameterize ([optimize-level 3])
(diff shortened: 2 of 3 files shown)


dybvig:
I think this is a good change, since map and for-each should check to make sure the first argument is a procedure except at optimize-level 3.  A couple of notes:

- you can use void-rec instead of building a call to void
- (not (eq? ctxt 'effect)) should be (not (eq? (app-ctxt ctxt) 'effect))

I see the latter was a preexisting issue.

gus-massa:
Fixed.

I also changed lines [3628](https://github.com/cisco/ChezScheme/pull/269/commits/dd96ca9fcf63148663b3e15c8bbb40d78dfc985c#diff-b056327abff41de13bd32f375c12cc3dR3628) and [3633](https://github.com/cisco/ChezScheme/pull/269/commits/dd96ca9fcf63148663b3e15c8bbb40d78dfc985c#diff-b056327abff41de13bd32f375c12cc3dR3633) so `make-seq` and `make-seq*` get the right value.

dybvig:
This looks good, and good catch on lines 3628 and 3633.  Please add a couple of error? mats for each of map and for-each with one and more than one empty-list argument so someone who attempts to add back the special case in cp0 or elsewhere will see that it is not appropriate.  Plus please mention `#2%for-each` as well as `#2%map` in the LOG entry.

gus-massa:
Fixed.

I copied the long example about the reduction of `#2%map` in a o=3 environment to `#2%for-each`, but I also added the small examples like `(map 3 '())` to all the similar constructions like `andmap`.

PR #282, opened 2018-03-29
standalone export form now handles (import import-spec ...)

owaddell:
The [CSUG entry](http://cisco.github.io/ChezScheme/csug9.5/libraries.html#./libraries:s17) for  standalone `export` forms permits an export-spec to take the form `(import import-spec ...)`, yet the implementation currently seems to require `(import import-spec)`.

```
> (library (Z) (export) (import (scheme)) (export (import)))
Exception: invalid export spec (import)
> (library (A) (export a) (import (scheme)) (define a 1))
> (library (B) (export b) (import (scheme)) (define b 2))
> (library (C) (export) (import (scheme)) (export (import (A) (B))))
Exception: invalid export spec (import (A) (B))
```

This commit tries to make the implementation consistent with CSUG on this point, but could use review by someone with more recent experience in `syntax.ss`.

### DIFF
--- a/LOG
+++ b/LOG
@@ -897,3 +897,5 @@
     schlib.c
 - install equates.h, kernel.o, and main.o on unix-like systems
     Mf-install.in
+- standalone export form now handles (import import-spec ...)
+    8.ms, syntax.ss, release_notes.stex
--- a/mats/8.ms
+++ b/mats/8.ms
@@ -7976,6 +7976,19 @@
   (equal?
     (let () (import ($l3)) (f (f 3)))
     3)
+  (begin
+    ;; (export import-spec ...) empty case
+    (library ($empty) (export) (import (chezscheme)) (export (import)))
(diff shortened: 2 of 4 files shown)


owaddell:
Maybe the last line in my example should have been:
```
> (library (C) (export) (import (chezscheme)) (export (import (A) (B))))
```

Perhaps I should update `8.ms` to match?

owaddell:
Bob Burger pointed out that a macro could do the work for us. The following appears to work for my use case.
```
> (library (C)
    (export)
    (import (chezscheme))
    (define-syntax re-export
      (syntax-rules ()
        [(_ imp ...) (export (import imp) ...)]))
    (re-export (A) (B)))
> (let () (import (C)) (list a b))
(1 2)
>
```

burgerrg:
Thanks for finding and fixing this bug!

PR #284, opened 2018-04-05
add newline to (import-notify) message in compile.ss

owaddell:


burgerrg:
It might be useful to add a mat to 8.ms for this.

burgerrg:
I didn't see any others missing in compile.ss.

On a related note, it seems that the code base uses `~%` and `\n` for newlines. I think the latter is slightly more efficient. You might consider changing all the `~%` to `\n` in compile.ss for consistency.

dybvig:
It shouldn't actually matter since the format-string compiler is designed to produce the same code for `~%` and `\n` in direct calls to `printf`, `fprintf`, `format`, `errorf`, etc.

burgerrg:
I don't think adding a mat for the import-notify "reading ..." message is worthwhile. We checked syntax.ss also and didn't find any other missing newlines.

Issue #69, opened 2016-06-08
Boot files with compiled libraries fail to load


burgerrg:
A.ss:

```
(library (A)
  (export A)
  (import (scheme))
  (define A 12))
```

```
% scheme
Chez Scheme Version 9.4.1
Copyright 1984-2016 Cisco Systems, Inc.

> (compile-file "A")
compiling A.ss with output to A.so
> (make-boot-file "A.boot" '("scheme") "A.so")
> (exit)
% scheme -b ./A.boot
Exception in compiler-internal: install-library/rt-code: unable to install invoke code for non-existant library #{A eldnidvc3vkegxvd5huz435gj-0}
```

This worked in Chez Scheme 8.4.1. Also, _non-existent_ is misspelled.

burgerrg:
Perhaps the load procedure in c/scheme.c does not handle the library/rt-info and library/ct-info records that run-inner of do-load-binary in s/7.ss does?

dybvig:
That's correct.  The c-coded boot-file loader has no way to talk with the Scheme-coded library manager.  One way to fix this would be to switch from the C-coded loader to the Scheme-coded loader once the latter is in place.  It might also be possible to have the C-coded loader invoke Scheme-coded library-manager procedures. I don't know whether there are other issues with compiled libraries in boot files.

owaddell:
The first commit on [this branch](https://github.com/owaddell/ChezScheme/commits/owaddell/library-boot-poc) is a quick stab at this. I restricted it to the non-procedure case of non-base boot files to limit overhead in the common case.

burgerrg:
Thanks, pull request https://github.com/cisco/ChezScheme/pull/288 fixed this.

owaddell:
The [`load`](https://github.com/cisco/ChezScheme/blob/master/c/scheme.c#L887) and [`handle_visit_revisit`](https://github.com/cisco/ChezScheme/blob/master/c/scheme.c#L820) code in scheme.c silently ignore unexpected values, but my change has us now going through [`run-inner`](https://github.com/cisco/ChezScheme/blob/master/s/7.ss#L211), which is not so forgiving.

Should `$make-load-binary` take a `for-boot?` flag that makes `run-inner` tolerate unexpected values?

jltaylor-us:
Is this issue resolved?

PR #288, opened 2018-04-12
handle compiled library code in boot files once base boot is loaded

owaddell:
This aims to fix the issue with compiled library code in (non-base) boot files that is described in #69.


burgerrg:
Thanks, this will be very useful!

PR #289, opened 2018-04-16
Return descriptions for most-positive-fixnum and most-negative-fixnum

gavwhela:
were reversed in csug, flipped to be correct.


akeep:
Thanks for the fix, the previous description was a bit confusing 😀

PR #295, opened 2018-05-02
Use the `system` flag of primitives to avoid check in primvars.ms

gus-massa:
Add a check in primvars.ms to skip the primitives with the `system` flag.

This is useful to add the signature to the system primitives so other parts of the compiler can use it. Currently only the `boolean-valued` flag is automatically using the signature, so I added the signature to all the system boolean primitives (and to `$fx+?` and `$fx-?` in spite they are not boolean).

I also added the other flags to them, but I'm **very** unsure about my selection of the flags. I tried to add `discard` to all of them, but some have a complex implementation so I was not sure and I didn't add the flag. And I'm in particular worried about the `mifoldable` flag.


akeep:
Closing this out, since it is a bit old and the benefit is not clear since other parts of the compiler do not make use of this data currently.

PR #296, opened 2018-05-02
Add `single-valued` flag to primitives to help reduce `call-with-values` expressions

gus-massa:
In particular this is useful to reduce `begin0` expressions that have a single valued result, to avoid the creation of an intermediate `lambda` or `list` to save the value.

The `single-valued` flag is autodetected from the signature, so it would be useful to add the signatures to the `system` procedures.

This needs new bootfiles, so the checks in travis-ci fails. [Look at this test instead](https://travis-ci.org/gus-massa/ChezScheme/builds/373721210).

I'm optimistic and I think this PR is correct, but I have a lot of related questions anyway:

I made a memoized version of `single-valued?` instead of a direct one because it detect many more opportunities during the compilation of the compiler, specially in cpnanopass.ss

There is no memoized version of `true-valued?`. I don't expect that it would be very useful, so I didn't modify it.

The functions `simple?` and `simple/profile?` use the same flag for memoization. Is that correct?

I added a special case to `single-valued?` to detect `call-with-values`, `apply` and `$apply` and go inside the procedure, because they always call the procedure and the result is the result of the call to the procedure. I think it's ok to avoid the arity check here. There are a few (5) similar functions, like `call-with-port`, but I didn't add them because they are probably not used too much.

I made a similar change to `boolean-valued?` because it also doesn't need too much checks. (It's possible to make a similar extension to `pure?` and the other memoized functions, but they must ensure that there are no errors, so I didn't modify them.) 

I was thinking about adding the `single-valued` flag to the `true` and `boolean-valued` flags (like `pure`, that implies `discard`). I finally didn't change that.


gus-massa:
Update:Ijust added an special case for `dynamic-wind`.

gus-massa:
Tiny update: Change the signature of `exact-integer-sqrt` to use `uint` instead of `integer`. (This should be squashed before merging.)

mflatt:

PR #299, opened 2018-05-04
- Double FMTBUFSIZE to fix compilation with gcc-8

weinholt:
Matthias Klose reported in https://bugs.debian.org/897720 that Chez Scheme fails to build with gcc-8, which is due to snprintf() writing to a buffer that is potentially too small. Please review this patch.


weinholt:
Anyone have a look at this yet?

akeep:
I think we could keep the `FMTBUFSIZE` slightly shorter, but when I computed a worst case for the format string that fails it was pretty close to 120 bytes long, though `gcc` seems to do a little more precise computation than I feared it was doing.  At any rate, I think 120 is okay for the buffer here.

Thanks for the update!

akeep:
From my reading it sounds like an error return value indicates a wide
character or other encoding problem in the format string, and since in the
usage within this print out we know the characters are ascii, I'm not sure
those kinds of errors can occur here.  Are there other errors you are
worried about us catching in this code?

-andy:)



On May 23, 2018 at 2:35:43 AM, Saigut (notifications@github.com) wrote:

The valid return value of snprintf is '0 <= return value < buffer size'.
But in C code of Chez Scheme, it seems that all return value of snprintf
have not been properly checked.

—
You are receiving this because you modified the open/close state.
Reply to this email directly, view it on GitHub
<https://github.com/cisco/ChezScheme/pull/299#issuecomment-391236282>, or mute
the thread

Saigut:
No error I can think out for now. The results checking of `snprintf` is just a suggestion, and this checking is looks like too tedious at here. And the compiler seems do a good job to check the `snprintf`.

PR #306, opened 2018-05-18
Reject attempts to visit invisible library

owaddell:
By default, libraries incorporated via `compile-whole-program` are not visible to `environment` or `eval`, yet the library manager could attempt to visit such a library and become unhappy. (If the library in question resides in a boot file (#69), the error message may be even less helpful.)

```
> (generate-wpo-files #t)
> (compile-imported-libraries #t)
> (compile-to-file
   '((library (A) (export a b) (import (scheme)) (define a 1) (define b 2)))
   "A.so")
> (with-output-to-file "prog.ss"
    (lambda ()
      (for-each pretty-print
        `((import (A) (scheme))
          (import-notify #t)
          (let ([args (map string->symbol (command-line-arguments))])
            (when (= b (length args))
              (pretty-print
               (top-level-value (car args) (environment (cdr args)))))))))
    'replace)
> (compile-program "prog.ss")
> (compile-whole-program "prog.wpo" "prog-visible" #t)
> (compile-whole-program "prog.wpo" "prog-not-visible" #f)
```

If run in the stock (5a11fac7fb92cd8a1bc278ddc782d33bd24c7106) system, the above results in:
```
$ scheme --program prog-visible b A
2
$ scheme --program prog-not-visible b A
import: attempting to 'visit' previously 'revisited' "prog-not-visible" for library (A) compile-time info
Exception: loading prog-not-visible did not define compile-time information for library (A)
```

With the change in this pull request, we instead get:
```
$ ~/projects/ChezScheme/a6le/bin/scheme --program prog-visible b A
2
$ ~/projects/ChezScheme/a6le/bin/scheme --program prog-not-visible b A
Exception in visit: library (A) is not visible
```

TODO:
- [x] get test results from Travis
- [x] update root expected errors
- [x] update LOG
- [x] rebase
- [ ] ?


dybvig:
This looks like a good change once LOG and the root-experr files are updated.  Thanks!

owaddell:
I'm not sure how best to update the root-experr-* and patch-* files. Have I missed a handy target in Mf-base or a note in the readme?

I tried something ad hoc for the initial rebase, but am now re-trying with something like the following, which has the CPU fan humming:
```
mats$ make clean
mats$ for O in 0 2 3; do\
       for SPI in f t; do\
        for CP0 in f t; do\
         for CIS in f t; do\
          make report root-experr o=$O eval=compile spi=$SPI cp0=$CP0 cis=$CIS;\
         done; done; done; done
mats$ make patches
````

dybvig:
To remake root-experr-compile-0-f-f-f, run `make all`, verifying the report file has only the new error messages you expect, then `make root-experr`.  root-experr-compile-0-f-f-f should be identical, so just make it a copy of the other.  root-experr-compile-3-f-f-f should not need to be updated.  Then to remake the allx patches, run `make allx`, verify the report files are empty or have nothing but the usual sorts of warning and error-message differences that the patches are intended to handle, and run `make patches`.  You can then check the patches by running `make just-reports` and verifying the report files are empty.

owaddell:
Thank for timely clues. I've updated the pull request.

Was it misguided to update patch-interpret-3-f-{f,t}-f where only the embedded time-stamps have changed?

dybvig:
It doesn't much matter, but if we care, `make patches` should avoid creating a new patch when only the timestamps would have changed.

PR #308, opened 2018-05-22
fix a couple typos and some duplicate words in csug

owaddell:


akeep:
Looks good, thanks for the edits!

PR #310, opened 2018-05-25
clarify "unknown module" error message; flush expand-output and expand/optimize-output

owaddell:
#### compile.ss
Should we revise the flush-output-port change to fetch parameter values once per `(when ...)`?

#### syntax.ss
I left a WIP commit at the tip hoping that @akeep can sanity check it.

Here is the error message issue (note that it is masked by the WIP commit):
```
> (import-notify #t)
> (generate-wpo-files #t)
> (compile-imported-libraries #t)
> (with-output-to-file "B.ss"
    (lambda ()
      (pretty-print
       '(library (B) (export b) (import (scheme)) (define b 9))))
    'replace)
> (with-output-to-file "prog.ss"
    (lambda ()
      (for-each pretty-print
        `((import (B) (scheme))
          (import-notify #t)
          (define (run args)
            (guard (c [#t (display-condition c) (newline)])
              (pretty-print
               (top-level-value (car args) (environment (cdr args))))))
          (when (> b 0) ;; reference export
            (let ([args (map string->symbol (command-line-arguments))])
              (if (= (length args) 2)
                  (begin
                    (run args)
                    (run args))
                  (error #f "expected 2 args")))))))
    'replace)
> (compile-program "prog.ss")
> (compile-whole-program "prog.wpo" "prog-visible" #t)
> (compile-whole-program "prog.wpo" "prog-not-visible" #f)
```

Before:
```
$ scheme --program prog-not-visible b B
Exception in visit: library (B) is not visible
Exception: unknown module #{B b8zzvxtl8z5jmjgbh9vh6q-0}
```

After the "clarify 'unknown module' error message ..." commit:
```
$ scheme --program prog-not-visible b B
Exception in visit: library (B) is not visible
Exception: unknown library (B)
```

After the "WIP" commit:
```
$ scheme --program prog-not-visible b B
Exception in visit: library (B) is not visible
Exception in visit: library (B) is not visible
```

### DIFF
--- a/LOG
+++ b/LOG
@@ -999,3 +999,5 @@
     misc.ms
 - flush expand-output and expand/optimize-output ports
     compile.ss
+- clarify "unknown module" error message in determine-module-imports
+    syntax.ss
--- a/LOG
+++ b/LOG
@@ -1001,3 +1001,5 @@
     compile.ss
 - clarify "unknown module" error message in determine-module-imports
     syntax.ss
+- restore the import code on reset to provide consistent error message
+    syntax.ss, 8.ms
--- a/LOG
+++ b/LOG
@@ -997,3 +997,5 @@
   whose base and/or index is a local save.
     cpnanopass.ss,
     misc.ms
+- flush expand-output and expand/optimize-output ports
+    compile.ss
--- a/mats/8.ms
+++ b/mats/8.ms
@@ -8977,7 +8977,47 @@
                        (eval '(lookup f g) (environment '(testfile-lr-l3) '(lookup))))])
            (eval '(lookup f g) (environment '(testfile-lr-l3) '(lookup))))))
     "Exception in h: user requested failure with (make-it-fail) parameter\nException in h: user requested failure with (make-it-fail) parameter\n")
-)
+
+  ;; re-arm import code if it complains about a library that is not visible
+  (begin
+    (with-output-to-file "testfile-lr-l4.ss"
+      (lambda ()
+        (pretty-print
+         '(library (testfile-lr-l4)
+            (export x)
+            (import (chezscheme))
+            (define x 123))))
+      'replace)
+    (with-output-to-file "testfile-lr-p4.ss"
+      (lambda ()
+        (for-each pretty-print
+         '((import (testfile-lr-l4) (scheme))
+           (define (run args)
+             (guard (c [#t (display-condition c) (newline)])
+               (pretty-print (top-level-value (car args) (environment (cdr args))))))
+           (when (> x 0) ;; reference export
+             (let ([args (map string->symbol (command-line-arguments))])
+               (if (= (length args) 2)
+                   (begin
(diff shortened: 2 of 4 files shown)


akeep:
I like the change, but I think there are a few things missing from the WIP change.

If you look at what we did for `visit-code` and `invoke-code`, we wrap the `on-reset` around the both invoking the appropriate thunk as well as doing the recursive processing on the invoke and/or visit requirements.  I think the `on-reset` for `import-library` should do the same.

The other thing you'll notice if you look at the `visit-code` is that we are also resetting the compile time environment, because it was possible to construct situations where part of the visit code thunk had run, leaving the compile-time environment in an inconsistent state.  It seems like this could also happen in the import code where exported identifiers are bound to indicate the they are either invoke (aka library-global) or visit identifiers and the module interface for the library is also added to the compile time environment.  If we really want to roll this back we may need to unset these in the compile time environment on import reset.  Unfortunately, that would require that we capture the exported invoke-time (aka revisit time) identifiers just as we do with the visit identifiers, so that we know what we are resetting.

Part of the reason I didn't do this when I was working on changing this for the `visit-code` and `invoke-code` procedures was that I didn't think there was ever anything in the body of the import code thunk that could cause an error to be raised, since we were just setting bindings in the compile-time environment to constants (either #f or the interface module), but it sounds like there was something I missed in the potential generated code?

I like the cleaned up error message, definitely better to see the library code error rather than an unknown module error when you were trying to import a library.

owaddell:
Thanks for pointing me to the other code. I updated the `on-reset` for `import-library` and added a mat in 8.ms.

I don't think you missed anything in the generated code: when you made the changes in 6dbd724, nothing in the import code thunk should have raised an error. My change in 220dca3 broke that assumption. Now the body of the import code thunk is either:
 1. a call to `$oops`, or
 2. a sequence installing compile-time bindings as before (quoted data).
 
In the first case, no compile-time bindings are installed, so there is nothing to roll back. In the second, nothing should raise an error.  (There is a call to `syntax-error` in the `do-import` case of `$sc-put-cte`, but `chi-top-library` doesn't generate such bindings.) I could be missing something, but it looks like we don't need to capture anything more for this. Maybe it needs a comment in the code?

owaddell:
@akeep could you look at this again? I fixed the placement of the `on-reset` and rebased. I think my reply above addresses your other feedback.

akeep:
@owaddell, I think these changes make sense.  It took me a little while to look through the code base a bit and see what all might appear in the `import-code` thunk.  After looking, I'm pretty sure there is nothing in that thunk that can raise an error, since it is just setting up compile-time environment bindings, and none of the expressions in the bindings do anything that can raise an exception, so I think it is probably unnecessary to "roll-back" the side-effects in that code.

So, yeah, I think this looks good.

owaddell:
@akeep, thanks for looking at this.

PR #317, opened 2018-06-08
fix OpenBSD build

matthastie:
Three issues are present and patched herein:

xlocale.h not present
time_t is 64 bits
wxneeded flag is present in scheme ELF to prevent "out of memory" (due to ENOTSUP) failure in segment.c.

When building on OpenBSD 6.2 (as tested), it is worthwhile reading this:

https://marc.info/?l=openbsd-ports&m=147060153102724&w=2

Chez must be constructed in a file system with the wxallowed flag present; this means /usr/local for the out-of-box configuration.


burgerrg:
Did 2649600c68ff57efb63d6d5d10c9d9f73368f59a fix the problem?

akeep:
Closing, assuming fixed.

PR #318, opened 2018-06-09
fix limited mrvs case of `boot_call`

owaddell:
The mrvs case of `boot_call` uses a zero-based index when it looks like it intends a 1-based index, but fixing that doesn't help since the value it's going after isn't saved. Since `boot_call` already handles the case of zero return values and otherwise ignores all but the first return value, it seemed simpler to set up `%ac1` which is already covered by the `save-scheme-state` in the hand-coded `invoke`.

Maybe there is a better fix? Either way, perhaps we can add a fix for this the next time we have a compelling reason to rebuild the boot files. (If we take this change, we can make the obvious simplification in `boot_call`, omitted here to minimize the diff.)

Before:
```
$ echo '(exit 7)' | scheme -q; echo $?
7
$ echo '(exit 4 5 6)' | scheme -q; echo $?
1
$ echo '(exit 0 8 9)' | scheme -q; echo $?
1
```

After: the latter two print 4 and 0 respectively.

I'm not sure what to do about the additional test cases in the commit at the tip of this branch. The [documentation](http://cisco.github.io/ChezScheme/csug9.5/system.html#./system:s156) says they should return -1, but `Sscheme_start` returns 1, which is simpler in cross-platform case, so maybe it's better to update CSUG there?

### DIFF
--- a/c/scheme.c
+++ b/c/scheme.c
@@ -340,8 +340,8 @@ static ptr boot_call(tc, p, n) ptr tc; ptr p; INT n; {
             p = Svoid;
             break;
         default:
-            p = S_get_scheme_arg(tc, 0);
-            break;
+            fprintf(stderr, "received %d values in boot_call\n", (int)(iptr)AC1(tc));
+            S_abnormal_exit();
     }
     return p;
 }
--- a/mats/7.ms
+++ b/mats/7.ms
@@ -3622,6 +3622,8 @@ evaluating module init
   (eqv? (exit-code '(exit 0)) 0)
   (eqv? (exit-code '(exit 24 7)) 24)
   (eqv? (exit-code '(exit 0 1 2)) 0)
+  (eqv? (exit-code '(exit 3.14)) 1)
+  (eqv? (exit-code '(exit 9.8 3.14)) 1)
   (begin
     (with-output-to-file "testfile-exit.ss"
       (lambda ()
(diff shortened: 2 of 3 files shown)


owaddell:
Clarification: The example output shown above was generated on **a6le**. On **a6nt** the multiple-value examples both return 0 before patching.

owaddell:
Simplified so it doesn't need to wait for next boot file update.

owaddell:
Bob and I put a couple alternatives on [owaddell/alt-invoke](https://github.com/owaddell/ChezScheme/commits/owaddell/alt-invoke)

burgerrg:
@dybvig To summarize, there are two bugs in the current code base. The C code has `S_get_scheme_arg(tc, 0)` instead of `S_get_scheme_arg(tc, 1)`. Once that is corrected, the return value is junk because the invoke code object doesn't save the scheme-arg registers. I propose a change the corrects the C code and the invoke code to save the scheme-arg registers. Will doing so break anything else?

dybvig:
That seems reasonable.  Having the invoke code object save the argument registers shouldn't cause a problem, as far as I know.

I agree with @owaddel that the documentation should change to reflect the (corrected) behavior.  By my reading of the code and documentation, the documentation is also wrong in the single-value case, both in the handling of a single Svoid value and in the handling of non-32-bit 64-bit integer values on 64-bit machines.  In both cases the documentation says the result should be -1, but the actual value for Svoid is 0 and the actual value for a non-32-bit 64-bit integer value on a 64-bit machine is the result of casting the value to an int.

burgerrg:
Fixed in commit https://github.com/cisco/ChezScheme/commit/7fd080903d96a1ef9ab79780bf73aa1619f33894.

Issue #320, opened 2018-06-14
ftype return code writes past end of allocated area for small ftypes


owaddell:
@mflatt it appears that the ftype return code writes past the end of ftypes whose size is less than the native pointer size. See the valgrind report below.

```
==9392== Invalid write of size 8
...
==9392==  Address 0x6023150 is 0 bytes inside a block of size 1 alloc'd
==9392==    at 0x4C2CB6B: malloc (vg_replace_malloc.c:299)
==9392==    by 0x429F65: s_malloc (in /usr/bin/scheme)
...
```

The output above was generated by running the following excerpt from foreign.ms in an a6le workarea.

```
$ cd mats
$ make foreign1.so
$ valgrind ../bin/scheme
(load-shared-object "./foreign1.so")
(define get (foreign-procedure "f4_get_i8" () (& integer-8)))
(let ([v (make-ftype-pointer integer-8 (foreign-alloc (ftype-sizeof integer-8)))])
  (get v)
  (ftype-ref integer-8 () v))
```

Perhaps change foreign4.c's `malloc_at_boundary` as follows, so that the system will raise an invalid memory reference when we exceed the boundary: 

```
EXPORT void *malloc_at_boundary(int sz)
{
  intptr_t alloc_size = getpagesize();
  char *p;
  p = mmap(NULL, 2 * alloc_size, PROT_READ | PROT_WRITE, MAP_ANON | MAP_PRIVATE, -1, 0);
  mprotect(p + alloc_size, alloc_size, PROT_NONE);
  return p + alloc_size - sz;
}
```

burgerrg:
@mflatt The a6nt `foreign.mo` mats fail earlier than the a6le mat shown above:
```
(eqv?
  (let ()
    (define-ftype foo (function (fixnum fixnum ptr) fixnum))
    (define f
      (lambda (n m k)
        (if (fx= n 0)
            (k m)
            (if (fx= (fxmodulo n 10) 0)
                (k (call/cc (lambda (k) (g (fx- n 1) (fx+ m 1) k))))
                (g (fx- n 1) (fx+ m 1) k)))))
    (define fptr (make-ftype-pointer foo f))
    (define g (ftype-ref foo () fptr))
    (let ([v ($with-exit-proc
               (lambda (k) (f $stack-depth $base-value k)))])
      (unlock-object
        (foreign-callable-code-object (ftype-pointer-address fptr)))
      v))
  (+ $stack-depth $base-value))
[...]

PR #321, opened 2018-06-15
x86_64: fix foreign-procedure `&` results that are not a multiple of a word

mflatt:
This is a repair for the main bug report in #320.

I have not yet run the updated tests on ppc32le, but I'll do that tomorrow.

### DIFF
--- a/LOG
+++ b/LOG
@@ -953,3 +953,6 @@
     prim5.c, scheme.c, schlib.c, schsig.c, stats.c, system.h,
     version.h, windows.c, foreign.stex, system.stex, mkheader.ss,
     prims.ss
+- Repair x86_64 `&` foreign-procedure result type handling for types of a
+  small size that is not a multiple of the word size
+    x86_64.ss, foreign.ms, foreign4.c
--- a/mats/foreign.ms
+++ b/mats/foreign.ms
@@ -2736,12 +2736,12 @@
                      (foreign-callable-code-object
                       (ftype-pointer-address id)))
                     v))]))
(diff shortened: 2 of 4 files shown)


burgerrg:
Thanks for the fix! The changes for `malloc_at_boundary` don't `mprotect` the second page. I think that's important to be able to detect this problem. Use `VirtualProtect(p + si.dwPageSize, si.dwPageSize, PAGE_NOACCESS, &dummy)` in Windows.

burgerrg:
```
EXPORT void *malloc_at_boundary(int sz)
{
  SYSTEM_INFO si;
  char *p;
  DWORD dummy;
  GetSystemInfo(&si);
  p = VirtualAlloc(NULL, si.dwPageSize, MEM_COMMIT | MEM_RESERVE, PAGE_READWRITE);
  VirtualProtect(p + si.dwPageSize, si.dwPageSize, PAGE_NOACCESS, &dummy);
  return p + si.dwPageSize - sz;
}
```

mflatt:
@burgerrg Ok, I'll make that improvement (but allocating two pages before adjusting the second one). The main problem wasn't so much being unlucky with the next page, though, but that the return-argument tests weren't using `malloc_at_boundary` before.

burgerrg:
Thanks, that makes sense.

mflatt:
The "foreign.ms" tests pass on ppc32le, and `malloc_on_boundary` is updated for Windows, so I think this is ready for review.

burgerrg:
Thanks again for this fix!

PR #323, opened 2018-06-22
Add r7rs character names

ecraven:
Make ChezScheme read #\escape and #\null, and also \| inside strings.
This does nothing about actually *writing*, it is just meant to make interoperability with R7RS schemes easier.


dybvig:
I don't object to these changes, but I'm curious.  What is the point of treating \| as | inside a string?

In the implementation, it seems better to handle it in the case just just above the one your modified that handles \\ and \" similarly.  Please add a LOG entry and a couple of mats.

ecraven:
@dybvig I'll rewrite the patch accordingly.
It seems the only reason | is included is for compatibility with '|foo\|bar|. Not a decision I know how to defend myself, but it is what r7rs is ;)

ecraven:
When asking on #scheme (2018-06-22), the answer was: 
````\| is necessary when escaping symbols, and we allowed it in strings for uniformity````

akeep:
Recognizing that this a relatively small change, it should probably update the documentation to reflect the added character names and string escape.  The LOG entries should also include the modified file, as the other entries in the LOG do.

omasanori:
May I help with something if it is stalled?

omasanori:
May I help with something if it is stalled?

PR #324, opened 2018-06-22
prefer nearest binding name for procedures

mflatt:
A macro or compiler can generally give a name to the procedure result of a `lambda` form by wrapping it in `(let ([<name> _]) <name>)`. Currently, though, that local name can get replaced by a different name from the context. For example, in

 (define g (let ([f (lambda (x) x)]) f))

the procedure bound to `g` probably has the name "g", although it depends on whether cp0 is enabled.

In this example, the name "g" ends up being used because `letify` effectively throws away the `let` wrapper and has no way to communicate the discarded name to its context. Although using the outer name could be considered a feature, I think it's probably an intentional consequence of the transformation.

This PR adjusts cp0 so that it doesn't replace a procedure name when one has been given already, and that has the effect of preserving the inner name.

### DIFF
--- a/LOG
+++ b/LOG
@@ -960,3 +960,5 @@
     schlib.c
 - Updated csug socket code to match that in examples folder
     csug/foreign.stex, examples/socket.ss
+- Adjust cp0 to not replace a procedure name from a let wrapper
+    cp0.ss, misc.ms
--- a/mats/misc.ms
+++ b/mats/misc.ms
@@ -4948,6 +4948,23 @@
     (procedure-arity-mask 17))
   )
 
+(mat procedure-name
+  (begin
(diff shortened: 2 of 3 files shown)


dybvig:
Looks good, thanks!

weinholt:
It appears that this change causes a build failure:

```
Exception in $fasl-file-equal?: string comparison failed while comparing ../boot/a6le/sbb and ../boot/a6le/petite.boot within fasl entry 547
```

dybvig:
I've just pushed a change to s/Mf-base to alleviate this problem, now and in similar future cases that might otherwise require creating and pushing new boot files.  The allx target now tries three times rather than two to rebuild the boot files and get an equivalent result.

Issue #325, opened 2018-06-26
X11 path issue on Mac OS X 10.13


tison1:
https://github.com/cisco/ChezScheme/blob/6c738bdfde601b815c296f489697e61c314e4621/c/Mf-a6osx#L20

Actually it is in `/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/Library/Frameworks/Tk.framework/Headers/`, there is not a symbol link from `/opt/include/X11`. See https://support.apple.com/en-us/ht201341

W4anD0eR96:
This causes 
```
(cd a6osx && /Applications/Xcode.app/Contents/Developer/usr/bin/make build)
(cd c ; /Applications/Xcode.app/Contents/Developer/usr/bin/make)
gcc  -m64 -Wpointer-arith -Wall -Wextra -Wno-implicit-fallthrough -Werror -O2 -I/opt/X11/include/  -c -DX86_64 -I../boot/a6osx -I../zlib expeditor.c
expeditor.c:888:10: fatal error: 'X11/Xlib.h' file not found
#include <X11/Xlib.h>
         ^~~~~~~~~~~~
1 error generated.
make[2]: *** [expeditor.o] Error 1
make[1]: *** [build] Error 2
make: *** [build] Error 2
```

I can locally change `-I/opt/X11/include/` t `-I/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX.sdk/System/Library/Frameworks/Tk.framework/Headers/` and `make` done. But wondering whether there is a more compatible approach.

W4anD0eR96:
@burgerrg any idea?

W4anD0eR96:
ummm, assume users are able to fix it by themselves is well

W4anD0eR96:
One can install [XQuartz](https://www.xquartz.org/) to fix this,  and I think it might worth a troubleshooting document/README/INSTALL

burgerrg:
I don't have XQuartz installed, so I cannot compile Chez Scheme for macOS without modifying c/version.h to remove the LIBX11 definition for macOS. I would welcome a pull request that adds a check for X11/Xlib.h and allows systems without it to compile.

W4anD0eR96:
Relevant #84

Hyzeta:
@W4anD0eR96 #328 has bug while build with X11, and it should not be merged. I will try to fix it.

[...]

PR #329, opened 2018-06-29
add an option `--without-x` to configure, which allows you to build ChezScheme without X11

Hyzeta:
On system without X11, you can configure like this, by using the new option `--without-x`, ChezScheme can be built without X11:

```
./configure --installprefix=$HOME/scheme --without-x
make
make install
```

see the new configure's help

```
Purpose:
  ./configure determines the machine type and constructs a custom Makefile
  and Mf-install, taking into account the options below.
Options (defaults shown in parens):
  ...
  --without-x                       build without X11
  ...
```

### DIFF
--- a/LOG
+++ b/LOG
@@ -960,3 +960,5 @@
     schlib.c
 - Updated csug socket code to match that in examples folder
     csug/foreign.stex, examples/socket.ss
+- Add an option --without-x to configure, which allows you to build ChezScheme without X11
+    c/expeditor.c, configure
--- a/c/expeditor.c
+++ b/c/expeditor.c
@@ -15,6 +15,7 @@
  */
 
 #include "system.h"
+#include "config.h"
 
 #ifdef FEATURE_EXPEDITOR
 
@@ -882,6 +883,10 @@ static void s_ee_carriage_return(void) {
 static void s_ee_line_feed(void) {
   putchar(0x0a);
 }
+
+#ifdef _WITHOUT_X_
+#undef LIBX11
+#endif
 
 #ifdef LIBX11
 #include <dlfcn.h>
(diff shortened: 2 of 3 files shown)


akeep:
As I mentioned in my comment in #325, fixing this correctly requires both changes to how the X11 support on macOS is handled in expeditor.c and a change to the configuration.

Having configure sed the expeditor.c file is also not really the right way to approach this, there should be a separate config.h file that controls the configuration options.

Additionally, please add to the `LOG` as detailed in `CONTRIBUTING.md`.

Hyzeta:
@akeep, Yes, I should use `config.h` for this (Not sed), I will create a new PR for this.

akeep:
I think this is a step in the right direction, but instead of modifying `expeditor.c` the `config.h` should be used done in `version.h`, because `version.h` is where this information is centralized.

I'd change the:

```C
#define LIBX11 "libX11.so"
```

lines to

```C
#ifndef DISABLE_X11
#define LIBX11 "libX11.so"
#endif
```

and the 

```C
#define LIBX11 ""/usr/X11R6/lib/libX11.dylib"
```

to

```C
#ifndef DISABLE_X11
#define LIBX11 "/usr/X11R6/lib/libX11.dylib"
#endif
```

I like `DISABLE_X11` a little more than `_WITHOUT_X_` partially because when I see an initial and trailing `_` in C I think of the main use of this as being the old stand-in for making sure your header files are only included once,  and I think I like `--disable-x11` a little more than `--without-x`, because I think it makes clear that you are disabling x11 support whether you have it or not.

Finally, I think an `#include "config.h"` should probably be added to `system.h` before the `#include "version.h"` so that the configuration settings are visible to the system (and importantly in this case to `version.h`.

Hyzeta:
@akeep I will modify the code according to your suggestion, and create a new PR for the update. Close  this PR for now.

PR #331, opened 2018-07-01
add an option `--disable-x11` to configure, which allows you to build Chez Scheme with X11 disabled

Hyzeta:
# add an option `--disable-x11` to configure, which allows you to build Chez Scheme with X11 disabled

now you can build Chez Scheme like this:

```
./configure --installprefix=$HOME/scheme --disable-x11
make
make install
```

see help by `./configure --help`

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,5 @@
+{
+    "files.associations": {
+        "system.h": "c"
+    }
+}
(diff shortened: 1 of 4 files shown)


Hyzeta:
Relevant #325

akeep:
Looks good, thanks!

burgerrg:
Thanks!

PR #334, opened 2018-07-16
update tests for "prefer closest binding name"

mflatt:
Some tests need repair for different evaluation modes – such as `interpret` instead of `compiled`, which doesn't preserve names.

### DIFF
--- a/mats/misc.ms
+++ b/mats/misc.ms
@@ -4952,18 +4952,22 @@
   (begin
     (define (procedure-name f)
       (((inspect/object f) 'code) 'name))
+    (define (ok-name? name expect)
+      (or (equal? name expect)
+          ;; interpreter currently doesn't keep names
+          (eq? (current-eval) interpret)))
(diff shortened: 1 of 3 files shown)


mats/misc.ms
```diff
@@ -4952,18 +4952,22 @@
   (begin
     (define (procedure-name f)
       (((inspect/object f) 'code) 'name))
+    (define (ok-name? name expect)
+      (or (equal? name expect)
+	  ;; interpreter currently doesn't keep names
+	  (equal? (current-eval) interpret)))
```
burgerrg: Suggestion: use `eq?` here and indent lines 4957-8 to line up with 4956.

burgerrg:
Thanks for fixing this!

mats/misc.ms
mflatt: Done: switched to `eq?` and untabified.

PR #335, opened 2018-07-17
Backreference support for debugging memory leaks

mflatt:
This patch adds `enable-object-backreferences` to enable support during collection for collecting a mapping from each live object to the object that made it live (or `#f` for a root). The `object-backreferences` function reports the most recently computed mapping. Enabling and retrieving the mapping for a maximum-generation collection provides a full picture of object references, modulo roots.

It's possible that the same information is available from the inspector interface,  as long as there's a way to get all the roots. It wasn't obvious to me that it's possible to get all the roots, given that some references may start from a static generation. Also, I like the idea that the information comes directly from the GC, so that it's more likely correct and there's no need to repeat all of that tracing. Still, I'm happy to drop this PR if there's a better approach.

To support backtraces, the sweep phase needs to know the start of each object that it's sweeping. When traversing memory regions for generational backpointers, the starts of the relevant object is not derivable for certain spaces. So, enabling backtraces also changes the kinds of spaces that are used for objects, and a couple of new space types are needed.

Otherwise, its straightforward to add backreference registrations during GC using C macros that produce no-ops unless compiled as the slower, object-counting and backrefernce-gathering copy of the GC.

I find a debugging tool like this useful once or twice a year in Racket development, and I needed it to track down leaks in the Racket-on-Chez implementation.

This PR builds on #190, because Racket-on-Chez uses both PRs and they interact significantly; only the second commit is specific to this PR. There are some small and more manageable interactions with #237, but those are ignored here.


mflatt:

PR #337, opened 2018-07-31
add hashtable-cells

mflatt:
This patch adds a `hashtable-cells` procedure that is mostly the obvious companion to `hashtable-keys` and `hashtable-values`, but with an extra twist: `hashtable-cells` takes an optional second argument to limit the resulting vector size — and therefore limit the amount of work performed by `hashtable-cells`. Calling `hashtable-cells` with a size of N takes O(N) time, even if the hashtable contains more than O(N) entries.

To iterate over N entries in O(N) time when N is not known in advance, get Q cells for the first Q iterations (for some small Q), then 2Q cells to continue for roughly up to 2Q iterations, and so on. The second set of iterations will actually go to somewhere between 2Q and 3Q iterations, since the call to get 2Q cells is not guaranteed to include the same cells as the first Q. (An intervening collection can change the order for an `eq?`-based table, for example.) Still, the first Q cells and second 2Q cells can be merged in O(Q) time to produce a vector of 2Q to 3Q cells, and so on.

Meanwhile, getting cells (instead of keys and values) accommodates iterations that change the hash table without exposing stale mappings and without needlessly retaining keys and values from a weak/ephemeron hash table.


dybvig:
I don't understand how you can merge the first Q and second 2Q cells in O(Q) time, given the lists aren't necessarily in the same order, and with the addition of "and so on" I think it gets beyond O(Q) time regardless of the order.  However, I like the idea of providing a `hashtable-cells` procedure, and I don't object to allowing the max-size limit to be specified.

Does it make sense to add optional max-size arguments to `hashtable-values`, `hashtable-keys`, and `hashtable-entries` for consistency?

The code for `hashtable-cells` does some redundant tests and `hashtable-size` calls (which lead to additional redundant tests and unnecessary memory references).  Plus I don't like going back through the top level for `case-lambda` clauses.  Instead, I prefer using a `rec` wrapper, so the call is local, or using a shared helper.  Here is a version that uses a shared helper and eliminates the redundant tests and calls to `hashtable-size`, but I haven't tried it to make sure it actually works.

```
(set-who! hashtable-cells
  (let ()
    (define (dispatch h max-sz)
      (unless (xht? h) ($oops who "~s is not a hashtable" h))
      (case (xht-type h)
        [(eq) ($eq-hashtable-cells h max-sz)]
        [(eqv) ($eqv-hashtable-cells h max-sz)]
        [else ($ht-hashtable-cells h max-sz)]))
    (define (invalid-length max-sz)
      ($oops who "~s is not a valid length" max-sz))
    (case-lambda
      [(h max-sz)
       (cond
         [(fixnum? max-sz)
          (unless (fx>= max-sz 0) (invalid-length max-sz))
          (dispatch h max-sz)]
         [(bignum? max-sz)
          (unless (>= max-sz 0) (invalid-length max-sz))
          (dispatch h (most-positive-fixnum))]
         [else (invalid-length max-sz)])]
      [(h) (dispatch h (most-positive-fixnum))])))
```

mflatt:
Thanks for the improved `hashtable-cells`. I'll look into similar versions of `hashtable-values`, `hashtable-keys`, and `hashtable-entries`.

Merging a first Q and second 2Q cells in O(Q) time works by using a `eq?`-based hashtable in the general case.

dybvig:
Fair enough, in amortized terms.  Still, the process you describe takes O(N) time, as you say, so it's the same asymptotically as calling `hashtable-cells` without the max-size argument, with a larger constant factor in general, so this doesn't seem like a good use of the max-size argument.  I can see wanting to process and remove at most Q elements at a time from a hashtable whose size might be greater than Q, so I can see why the max-size argument might be useful.

mflatt:
The updated commit adds an optional size argument to `hashtable-keys`, `hashtable-values`, and `hashtable-entries`, which involved splitting out a separate R6RS variant for `hashtable-keys` and hashtable-entries`.

-------

I see that I haven't explained my use of `hashtable-cells` well. Maybe it doesn't matter, anyway, but I'll give it one more try.

My overall goal is to adapt hashtables to a generic sequence-iteration interface. The interface includes operations to get a starting index, to increment an index (functionally, so the previous index remains usable as long as the hash table doesn't change in certain ways), and to get the value/key at an index. Those operations all need to be amortized constant-time.

So, it's ok to take O(N) time overall to get to the Nth index/key/value, but it must be O(N) time even if the size of the table is greater than O(N) — and there's no way to predict N in advance, because N corresponds to the number of iterations steps that are performed before the iteration is abandoned.

Adding `hashtable-cells` with a size argument is enough to achieve that goal, but it wasn't obvious to me until I thought of the merging piece, so that's why I commented on the merging strategy.

dybvig:
Merged and committed.  Thanks!

PR #342, opened 2018-08-14
change the way the dependency package

ltoddy:
The code for these changes is mainly for people in China who can compile better using the chez scheme.

### DIFF
--- a/BUILDING
+++ b/BUILDING
@@ -28,6 +28,39 @@ later in this file):
 ./configure
 sudo make install
 
+for chinese:
+
+  As we all know, china has the great firlwall. Executing the `./configure` will be hard to succeed.
+  The reason is that ChezScheme depend on three packages (on github),
+
+  * https://github.com/nanopass/nanopass-framework-scheme/archive/v1.9.tar.gz
+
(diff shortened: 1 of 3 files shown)


TisonKun:
@ltoddy 
1. please update LOG with this PR
2 in BUILDING, I approve using Chinese as information since it is mainly for Chinese user. However, the main content should use English, which is the only general language in dev for now.
3. also I suggestion you would consider more automate configure since you now use relative path. maybe `$(dirname "$0")` helps.

---
1. ChezScheme 要求在 PR 时同时修改 LOG
2. 修改 BUILDING 文件的时候最好用英文表述，当然我很乐意看到作为附加注释的中文
3. `configure` 里你用了相对路径，这个不太好，可以考虑用 `$(dirname "$0")` 获取执行文件路径，然后在 BUILDING 里说明放到项目根目录
---

@burgerrg FYI, people might have difficult to fetch stuff from GitHub in China for some reason.
This PR allows user use their own nanopass, zlib and stex. Regardless the motivation, this PR itself make ChezScheme more configurable. :-)

ltoddy:
no need for `$(dirname)` here.

Issue #364, opened 2018-11-16
$1000 USD Bug Bounty: Invalid memory reference: foreign-callable (gc bug?)


stergiotis:
I encountered an invalid memory reference when calling a foreign-callable multiple times from the foreign code. Strangely the bug is only present when a typed ftype-pointer argument is passed to the foreign-callable and not present when only the address (i.e. void*) is passed.

Below a reproducible minimal example:

`minimal.scm`
```scheme
(import (chezscheme))

(define i 0)
(define myguardian (make-guardian))
(define-ftype mystruct_t (struct (a unsigned-64) (b unsigned-64)))
(define handler-crash (foreign-callable (lambda (dummy n)
                                    (set! i (fx1+ i))
                                    (pretty-print i)
                                    #t) ((* mystruct_t) unsigned-64) boolean))
(define handler-nocrash1 (foreign-callable (lambda (dummy n)
                                    (set! i (fx1+ i))
                                    (pretty-print dummy)
                                    (pretty-print i)
                                    #t) ((* mystruct_t) unsigned-64) boolean))
(define handler-nocrash2 (foreign-callable (lambda (dummy n)
                                    (set! i (fx1+ i))
                                    (pretty-print i)
                                    #t) (void* unsigned-64) boolean))

(alias handler handler-crash)
;(alias handler handler-nocrash1)
;(alias handler handler-nocrash2)
;(define handler handler-crash) ;; --> no crash


(collect-notify #t)
(collect-trip-bytes (fx/ (collect-trip-bytes) 8))

(lock-object handler)
(load-shared-object "./libloop.so")
(define aa #f)
(let ((looper (foreign-procedure "looper" (void*) void)))
  (set! aa (foreign-callable-entry-point handler))
  (looper aa))
```
`libloop.c`
```c
typedef int (func_t)(int i,int dummy);
void looper(func_t f) {
  int i=0;
  while(f(i++,0)) {
    ;
  }
}
```
`run.sh`
```bash
#!/bin/sh
gcc -c libloop.c
gcc -Wl,-soname,libloop.so -shared -fPIC -o libloop.so libloop.o

scheme --program ./minimal.scm
```
Beware: Slight alternations of the code (e.g. replacing `alias` by a `define`) seem to mitigate or delay the heap corruption.
I am running Chez from the master branch (commit `a5db479b68b74dda9f62665c44cfad2b1baf322e`) under 64 bit Linux.

[...]

PR #365, opened 2018-11-16
correct CP(TC) management for callable return

mflatt:
Possibly the repair for #364

### DIFF
--- a/LOG
+++ b/LOG
@@ -1021,3 +1021,7 @@
   get-mode, and path-absolute more consistent with
   https://docs.microsoft.com/en-us/windows/desktop/FileIO/naming-a-file
     6.ss, 6.ms, io.stex, release_notes.stex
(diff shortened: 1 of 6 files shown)


mflatt:
I believe this new commit is much more likely to be the repair for #364.

If receiving arguments to a callable involves allocation — such as when receiving a flonum, bignum, or pointer  — then the value in the `cp` field of the thread context can get changed when `%cp` is a real register and the register value gets saved in the thread context. In particular, if the caller or a callable has set that register, then the `cp` field in the thread context will have that wrong value. Guard against the problem by claiming `%cp` as live and initialized by `cp` from the thread context while unpacking arguments.

dybvig:
Many thanks to @stergiotis, all others who contributed, and especially to @mflatt for fixing the issues.

PR #367, opened 2018-11-27
Avoid reading undefined values by zero-initializing allocation

kalvdans:
When using the USE_MALLOC allocator instead of USE_MMAP, valgrind gives warnings. The merge request silences the warnings. Maybe better to initialize the relevant fields explicitly instead of zeroing the whole segment.

==16544== Conditional jump or move depends on uninitialised value(s)
==16544==    at 0x11E308: copy (in /home/chn/chez/install/bin/scheme)
==16544==    by 0x126BCD: S_gc_oce (in /home/chn/chez/install/bin/scheme)
==16544==    by 0x114A81: Scompact_heap (in /home/chn/chez/install/bin/scheme)
==16544==    by 0x13AAD7: Sbuild_heap (in /home/chn/chez/install/bin/scheme)
==16544==    by 0x10F9D9: main (in /home/chn/chez/install/bin/scheme)
==16544==  Uninitialised value was created by a heap allocation
==16544==    at 0x483577F: malloc (vg_replace_malloc.c:299)
==16544==    by 0x11071E: S_getmem (in /home/chn/chez/install/bin/scheme)
==16544==    by 0x110B31: S_find_segments (in /home/chn/chez/install/bin/scheme)
==16544==    by 0x111590: S_find_more_room (in /home/chn/chez/install/bin/scheme)
==16544==    by 0x11182F: S_alloc_init (in /home/chn/chez/install/bin/scheme)
==16544==    by 0x13A9D4: Sbuild_heap (in /home/chn/chez/install/bin/scheme)
==16544==    by 0x10F9D9: main (in /home/chn/chez/install/bin/scheme)

### DIFF
--- a/LOG
+++ b/LOG
@@ -1025,3 +1025,4 @@
   where the cp register copy in the thread context could be changed
   in the callable prep before S_call_help gets it
     cpnanopass.ss, x86_64.ss, x86.ss, foreign2.c, foreign.ms
+- initialize all fields of seginfo to avoid undefined values
--- a/LOG
+++ b/LOG
@@ -1025,4 +1025,5 @@
   where the cp register copy in the thread context could be changed
   in the callable prep before S_call_help gets it
     cpnanopass.ss, x86_64.ss, x86.ss, foreign2.c, foreign.ms
-- initialize all fields of seginfo to avoid undefined values
+- added initialization of seginfo sorted and trigger_ephemerons fields.
+    segment.c
--- a/c/segment.c
+++ b/c/segment.c
@@ -263,7 +263,6 @@ iptr S_find_segments(s, g, n) ISPC s; IGEN g; iptr n; {
 
         chunk->nused_segs += 1;
         initialize_seginfo(si, s, g);
-        si->sorted = 0;
         si->next = S_G.occupied_segments[s][g];
         S_G.occupied_segments[s][g] = si;
         S_G.number_of_empty_segments -= 1;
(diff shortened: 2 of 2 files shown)


dybvig:
Thanks for the pull request.  Can the code actually reference uninitialized values, i.e., is the chunk not fully initialized by the call that follows the getmem call?  If so, zero-filling is likely not an appropriate solution.  If not, I'd rather not introduce the extra overhead just to satisfy valgrind.

kalvdans:
>  Can the code actually reference uninitialized values, i.e., is the chunk not fully initialized by the call that follows the getmem call?

Yes, the copy() function reads the uninitialized values as seen on the valgrind log.

dybvig:
Do you happen to know what is going on when it happens?  For example, does it happen when running a particular test in the mats?

I've stared at copy() a bit and it looks like one possible culprit is the seginfo `trigger_empherons` field, which does seem to be referenced without being set.  Would you please try initializing it to NULL in `segment.c:initialize_seginfo` and see if that resolves the issue?

kalvdans:
Thanks Dybvig for the reply. Please try valgrind yourself, I added these two lines after the include statements in segment.c
```
#define USE_MALLOC
#undef USE_MMAP
```
Then I passed CFLAGS="-O0 -g" to make to get valgrind to pinpoint the line number as well.
I get warnings on the `trigger_empherons` field and I also got a warning on the `sorted` field of seginfo, by running this small program with scheme -program empty.scm
```
(import (chezscheme))
```

dybvig:
Thanks for bringing this to my attention and for narrowing it down.  I've confirmed the issue and have a more surgical fix, which is to add the following two lines to `initialize_seginfo` in segment.c:
```
  si->sorted = 0;
  si->trigger_ephemerons = NULL;
```
With this fix, valgrind doesn't turn up any errors with the empty program or when rebuilding in the s directory or running the mats.

Would you like to alter your pull request to replace zero-filling with these two lines?  If so, please also add a brief entry in LOG.

kalvdans:
Thanks for verifying a clean valgrind run for you as well.

> Would you like to alter your pull request to replace zero-filling with these two lines? If so, please also add a brief entry in LOG.

Done! However, feel free to brush up the merge request yourself since I'm not familiar with the code style and the right level of details in LOG.

PR #375, opened 2018-12-29
avoid stack overflow on fasl-read

mflatt:
This patch changes `fasl-write` to avoid generating a fasl content that can cause `fasl-read` to crash.

The `fasl-read` function can crash because it's implemented in the kernel using recursive C calls, which can exhaust the C stack. The change to `fasl-write` detects when nesting has gone 500 layers deep, and for each vale in the next layer, it adds a graph reference to the root of the fasl content. The graph reference has the effect of resetting nesting. (Most of the patch is propagating a depth counter through the graph-construction phase of `fasl-write`.)

### DIFF
--- a/LOG
+++ b/LOG
@@ -1027,3 +1027,5 @@
     cpnanopass.ss, x86_64.ss, x86.ss, foreign2.c, foreign.ms
 - added initialization of seginfo sorted and trigger_ephemerons fields.
     segment.c
+- avoid fasl overflow of C stack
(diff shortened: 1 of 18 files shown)


dybvig:
Have you considered modifying the fasl reader to maintain an explicit stack?  It seems like any limit, including 500, could result in a stack overflow depending on the C run-time system.

mflatt:

PR #376, opened 2018-12-29
Faster format for loading code

mflatt:
The good news is that this patch can cut startup time in half when loading uncompressed boot files. The bad news is that it's a significant addition to the kernel: about 2/3 the size of the GC implementation, and duplicating some of the GC code in its current form.

On my machine, startup time from uncompressed "petite.boot" and "scheme.boot" files is reduced from 60ms to 30ms. With Racket-on-Chez's "racket.boot" in addition (which is a little larger than "petite.boot" and "scheme.boot" combined), startup goes from 140ms to 70ms. The extra code seems worth it to save 70ms on every Racket's startup — unless, of course, something simpler can provide the same benefit.

The "vfasl" format implemented by this patch is mostly an image of the code and data that can read directly into a GC-managed memory space. Just as important as loading the image directly, it can be loaded into the static generation, saving work to compact loaded code into the static generation.
(The vfasl loader does not update object counts for static content.)

The vfasl format can also cut loading non-boot code in half, at least for relatively large code files, but the benefit of vfasl loading is limited if the code has to be traversed later by the GC. Also, the in-memory format of data is larger than the fasl format, and the size difference tends to defeat any benefit of vfasl for small code files.

Use `vfasl-convert-file` to convert to the vfasl format, something like this:
 ```
     (compile-compressed #f)
     (vfasl-convert-file "orig/petite.boot" "new/petite.boot" '())
     (vfasl-convert-file "orig/scheme.boot" "new/scheme.boot" '("petite"))
     (vfasl-convert-file "orig/mine.boot" "new/mine.boot" '("petite" "scheme"))
```

### DIFF
--- a/LOG
+++ b/LOG
@@ -1027,3 +1027,7 @@
     cpnanopass.ss, x86_64.ss, x86.ss, foreign2.c, foreign.ms
 - added initialization of seginfo sorted and trigger_ephemerons fields.
     segment.c
+- add vfasl format for faster loading
+     compile.ss, back.ss, library.ss, primdata.ss, cmacros.ss,
+     vfasl.c, fasl.c, alloc.c, scheme.c, intern.c, globals.h, externs.h,
+     Mf-base, misc.ms, system.stex, release_notes.stex
--- a/boot/a6le/equates.h
+++ b/boot/a6le/equates.h
@@ -222,6 +222,7 @@ typedef unsigned long U64;
 #define fasl_type_symbol 0x2
 #define fasl_type_symbol_hashtable 0x20
 #define fasl_type_vector 0x4
+#define fasl_type_vfasl_size 0x15
 #define fasl_type_visit 0x22
 #define fasl_type_weak_pair 0x1E
 #define fixnum_bits 0x3D
--- a/boot/a6nt/equates.h
+++ b/boot/a6nt/equates.h
@@ -222,6 +222,7 @@ typedef unsigned long long U64;
 #define fasl_type_symbol 0x2
 #define fasl_type_symbol_hashtable 0x20
 #define fasl_type_vector 0x4
+#define fasl_type_vfasl_size 0x15
 #define fasl_type_visit 0x22
 #define fasl_type_weak_pair 0x1E
 #define fixnum_bits 0x3D
--- a/boot/a6osx/equates.h
+++ b/boot/a6osx/equates.h
@@ -222,6 +222,7 @@ typedef unsigned long U64;
 #define fasl_type_symbol 0x2
 #define fasl_type_symbol_hashtable 0x20
 #define fasl_type_vector 0x4
+#define fasl_type_vfasl_size 0x15
 #define fasl_type_visit 0x22
 #define fasl_type_weak_pair 0x1E
 #define fixnum_bits 0x3D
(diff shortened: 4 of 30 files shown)


samth:
This looks potentially related to work in GHC: http://ezyang.com/compact.html I wonder if they managed to avoid the complexity you're worried about.

mflatt:
@samth That work doesn't deal with code, shared objects like symbols and record-type descriptors, or mutable values (such as symbols that have property lists) – all ways to reduce complexity, certainly, but not applicable to the problem of loading and linking Scheme code.

dybvig:
For many years, Chez Scheme supported the notion of "saved heaps", and "incremental saved heaps", which is similar to this.  Saved heaps were essentially pre-built heap images.  There was a saved heap that corresponded to the base boot file and incremental saved heaps that corresponded to non-base boot files.  With mmap'd saved heaps, startup time was practically zero, and sharing of memory between processes cut total memory usage considerably when many Scheme processes were used.

We ended up abandoning saved heaps when address randomization became ubiquitous.

Perhaps it's time to come up with a new implementation of saved heaps.  One approach is to have a "live" saved heap in the form of a background process created upon the first invocation of Chez Scheme (or Racket), and have each subsequent invocations merely fork a new copy of the background process.  This has the downside that the forked process is a child of the background process rather than the program (e.g., shell) of invocation and its environment, but it's not unlike how other applications work these days.

Another approach is to save an image as we used to and adjust all the pointers on start-up to account for the new random base address.  Of course, we can always request the original address when mmap'ing and avoid the adjustment if we get it.

If machines supported EA-relative addressing (like PC-relative jumps add in the address of the jump, EA-relative loads would add the effective address of a loaded pointer to the loaded pointer), we could avoid the adjustment entirely.  We could of course do the add ourselves at the cost of an extra instruction or two at each ptr load.

dybvig:
Thanks for the suggestion, @mromyers.  I'll take a look at the slides when I get a chance, though I know how to go about saving the Chez Scheme heap from prior experience.

A warning for anyone else considering working on saved heaps for Chez Scheme: The slides are as deep as we can go, unless there's a more detailed writeup.  Unless I'm mistaken, the Emacs source code is GPL'd, so even if someone just perused the code to gain knowledge of its workings before working on an implementation for Chez Scheme, that could taint our (permissive open-source) code.

fweimer:
Perhaps the comment at the top of `c/fasl.c` should be updated to mention the vfasl alternative?

I've started to look at translating existing fasl files to `ET_REL` ELF objects, basically letting the ELF link editor handle the relocation processing. I do not have any code yet, but I believe that ELF is sufficiently powerful for this (e.g., large section alignment to meet GC requirements). The initial goal would be to produce position-dependent code only, which simplifies the task somewhat.

dybvig:
I'll be interested in seeing where you end up with this.  Code objects presently contain pointers to other objects, e.g., quoted constants, symbols used to store top-level values, and other code objects.  It will be difficult to support all of those with position-independent code without going to relocation tables.

Whether ELF is powerful enough or not, we might want to go with a solution of our own for portability.

LiberalArtist:
Just as another point in the design space, according to [Andy Wingo's blog](https://wingolog.org/archives/2014/01/19/elf-in-guile), Guile has been using ELF as its object file format for several years. For portability, apparently they implement their own linker and loader and use them on all platforms.

mflatt:
While vfasl has worked well for Racket, the implementation in this PR is no longer the one used. The main change is that the vfasl writer was rewritten in Scheme. (See racket/ChezScheme@729dbd09ba35f48c45ab1c060a62178a68a6ad16 for more information.) So, I'll close this PR.

Issue #293, opened 2018-04-30
Difference in error semantics with cp0 enabled


laqrix:
Using commit `0be4cfd` (the current tip of master):

```
Chez Scheme Version 9.5.1
Copyright 1984-2017 Cisco Systems, Inc.

> (let ([v (values 1 2)]) v)
1
2
> (run-cp0 (lambda (f x) x))
> (let ([v (values 1 2)]) v)
Exception: returned two values to single value return context
> 
```

gwatt:
This behavior is actually OK. It's explicitly called out in r6rs here: http://www.r6rs.org/final/html/r6rs/r6rs-Z-H-14.html#node_sec_11.20 at the bottom of the page. It's exactly the same in r5rs and r7rs as well.

However, you have disabled cp0 and thus Chez Scheme does not reduce the let expression to the values form.

laqrix:
Practically, this hasn't been a problem. I was surprised to find one of my own macros could reduce to an expression where multiple values seem valid.  Disabling cp0 was just to demonstrate the difference in error behavior.

mflatt:
There's no difference, anymore, as of the v9.9.9 merge.

PR #377, opened 2018-12-31
cp0: move only known-single-valued out of single-value context

mflatt:
This patch changes cp0 to avoid turning an error like
```
 (let ([x (values 1 2)]) x)
```
or
```
 (car (list (values 1 2)))
```
into a program that returns multiple values.

The change had no effect on benchmarks run via Racket-on-Chez as measured by time. Also, the effect on compiling Chez Scheme itself is very small as measured by temporarily adjusting cp0 to print an alert when a transformation is suppressed by the patch. Most of the differences are in calls to `tree-bit-set` and similar in "cpnanopass.ss", such as in
```
                  (let ([new (tree-bit-set live* live-size index)])
                    (safe-assert (or (eq? new live*) (not (tree-same? new live*))))
                    new)
```
Here, the binding to `new` is there for `safe-assert`, but when assertions are not used, then the `new` binding could be eliminated — except for the fact that it's in tail position with respect to the enclosing function, and the always-single-valued result of`tree-bit-set` is not apparent to the optimizer. The patch includes one a replacement of one macro-inserted `values` to `begin` to avoid some other optimization changes.

This patch is meant to complement #336 by avoiding optimizations that might change the continuation frame that is modified or inspected by the expression, such as removing the `values` around `(values (call-with-current-continuation-attachment ....))`. No specific cooperation is included in this patch; it's just that cp0 doesn't currently treat `call-setting-continuation-attachment` or `call-with-current-continuation-attachment` specially, and those functions might return multiple values.


dybvig:
This comes up from time to time.  R6RS and Chez Scheme don't actually require an exception to be raised when multiple values are returned to a single-value context, so the transformation is valid as well as useful, albeit sometimes surprising.  If there is a way to achieve what you're trying to achieve with respect to #336 without disabling the optimization in the general case, I would prefer that.  Note that cp0 can affect whether an exception is raised in other unspecified situations, at least by changing the order of evaluation.

mflatt:
The general possibility of the compiler exploiting unspecified behavior is a problem for building Racket on Chez Scheme, in principle. But making Racket-on-Chez's schemify compiler pass guard against certain kinds of unspecified behavior has worked well enough. (In the long run, a more principled audit is needed.) This is one place where I don't see a good Chez Scheme target for the semantics that Racket needs.

In the same way that proper handling of tail calls obliges a compiler and runtime system to avoid extending the continuation in some cases, the semantics of continuation marks in Racket oblige  a compiler and runtime to extend the continuation in some cases. The continuation of attachments of #336 similarly make sense only when there's some guarantee about some forms not being in tail position.

I can imagine adding a `value` form to Chez Scheme that is essentially a `begin` with a single subform, but where the subform is specified to be non-tail with respect to the `value` form. Then, schemify could insert `value` where needed. But without relying on any cp0 details – that is, if the semantics of Chez Scheme is that no other form guarantees a non-tail position – then schemify will have to insert many `value` wrappers. And then I'd want cp0 to remove nearly all of them them, which is pretty much the analysis in this patch. More practically, schemify could rely on some information about which non-tail positions cp0 and the compiler preserve, although I'd still worry about `value` forms that cp0 could have easily removed.

That chain of thought led to me investigate whether just adapting Chez Scheme's guarantees for `let` (and some primitives like `values`) would make more sense. Although I understand that RnRS and Chez Scheme include various form of unspecified behavior (I did see #293 at some point), my impression is that Chez Scheme is designed to detect unspecified behavior when practical and when the run-time cost is low, even when it adds some complexity to the compiler. That goal seems reflected, for example, in the extensive analysis used to transform `letrec[*]` to reliably report use before definition. Flagging an incorrect return arity seemed in the same spirit.

Flagging an incorrect return arity and providing a guarantee about non-tail behavior are not the same thing, but they turn out to be close enough that a unified analysis makes sense. So, trying to hit two birds with one stone is how I got here.

There's one other obvious possibility, which is to add a parameter that enables the optimization constraint that is imposed by this patch, but where the parameter is `#f` by default.

mflatt:
On further investigation, I see that this current attempt is woefully incomplete. I'll try again.

mflatt:
The new version of the patch is a better attempt to consistently catch multi-value results to single-value contexts in safe compilation modes. It pulls in part of #296, which has a more complete way of inferring single-valued primitives.

Most transformation constraints (especially the new ones) are disabled at optimization level 3. The constraints to keep a non-tail expression as non-tail stay in effect in all optimization modes, currently, since the intent is to refine the semantics of non-erroring programs.

The constraint against moving a non-tail expression to tail position can prevent the optimizer from turning a deep non-tail recursion into a loop. Otherwise, the new constraints rarely affect compilation output even in safe mode, and I haven't been able to construct with a significant performance difference.

As a minor point, the patch corrects a few transformations that I think probably were not intended to be allowed in safe mode, even before this change. For example, `(string-ref (string 1 #\a 2) 1)` was converted to `\#a`, but now it reports that 1 is not a character in safe mode.

mflatt:
Subsumed by #336.

PR #378, opened 2019-01-05
add `compute-size-increments`

mflatt:
This patch adds an extension of `compute-size` that supports reachability-based memory accounting.[*] Reachability-based accounting is how DrRacket supports a memory limit for user programs, even though user programs run on the same Racket instance as DrRacket itself.

[*] See http://www.cs.utah.edu/plt/publications/ismm04-wf.pdf and http://www.cs.utah.edu/plt/publications/ismm04-addendum.txt

The main change is to add `compute-size-increments`, which takes a list of objects and reports the size reachable from each object — skipping any objects that were already counted for a previous object in the list, and also not counting as reachable any immediate objects later in the list. (Objects reachable from later in the list still count as reachable for earlier objects).

Roughly, Racket calls `compute-size-increments` with a list containing a root thread followed by child threads, so the root thread gets most charges, but anything reachable only through a child thread is charged to that child. The actual set of roots is determined by custodians instead of threads, and the actual relevant roots are a little more complex, but that's the idea.

The `compute-size-increments` function shares most of its implementation with `compute-size`, but `compute-size-increments` uses a bitmap for its visited set, so it uses less memory by a factor of 10 or 20. It also treats weak pairs and ephemeron pairs a little differently; for example, the value in a emphemeron typically should be charged to the thread that holds the key, and not the thread that holds the pair.

Racket threads are not Chez Scheme threads, but Racket places are implemented as Chez Scheme threads, and accounting for places needs `compute-size` to traverse the stack of a deactivated thread. (Traversing the stack of an active thread would not be safe, but Racket wires accounting into the collect callback, so all threads are deactivated except for the one running accounting, and `call/cc` can be used to get the relevant stack for that last one.) Since Chez Scheme thread objects become slightly more useful this way, the patch adds `get-initial-thread` as a way of getting the initial thread object.

Finally, since `compute-object-increments` needs to recognize weak pairs and emphemeron pairs, it seemed worthwhile (10-20% performance improvement) to inline `weak-pair?` and `ephemeron-pair?`, so this patch does that. It also implements an inlined `$generation`, which is almost the same. In fact, it adds inlined `$maybe-seginfo`, `$seginfo-generation`, and `$seginfo-space`, which lets `compute-size-increments` avoid repeated segment-table lookups the same as in the GC, although that probably doesn't matter.


mflatt:

Issue #372, opened 2018-12-19
Mutexes and condition variables with names


jessymilare:
In many other scheme implementations, CL implementations and also SRFI 18 and SRFI 21, mutexes and condition variables have an optional field called 'name', both for developing and debugging.

This could be accomplished by creating a new field "name" in mutex and condition records and then redefining make-mutex and make-condition to accept an optional argument 'name'. This would change the number of arguments for $make-mutex and $make-condition, which AFAIK are only directly called in s/prims.ss, so that would not be a big issue.

If this is desired, I can make these changes (and also tests and LOG entry) and create a pull request.

yjqww6:
How about just using a weak eq hashtable to attach a name to all sorts of objects?

jessymilare:
It would (1) not be printed and (2) occupy more space. During debugging, if you ask to show local variables or the arguments of a procedure and see a mutex or condition, it's name wouldn't be printed, you would need to inspect it, etc. I just think displaying the mutex/condition name is a nice feature Chez Scheme could have.

jessymilare:
Alright, I'm not gonna implement an ideia the maintainers didn't like.

dybvig:
Thanks for the feature suggestion and offer to implement it.  Please don't assume the maintainers don't like the idea simply because we're silent---as you can see from the number of oustanding pull requests and issues, we're not keeping up at present.

In this case, while I don't recall encountering a situation where a mutex or condition-variable name would have been useful, I don't object to supporting the feature if others would find it useful.  So if you think it's useful enough to spend time on a pull request, I'll look it over and, possibly after requesting changes, pull it.  I would suggest that the name, if supplied, be a symbol to simplify printing and for consistency with procedure and record-type names.

In general, when deciding whether to add a debugging aid, the decision hinges on whether and how it can be done without significant run-time cost.  In this case, the simple solution of adding an extra name field to each mutex and condition variable would make the objects larger and increase the cost of allocating and collecting them.  On the other hand, a typical program probably doesn't allocate many mutexes and condition variables or allocates them as part of other structures, so the overhead of the extra field would likely be small if not insignificant for most programs.

Alternatively, you could subclass the mutex and condition record types so that mutexes and condition variables without name fields wouldn't be any larger.  This would make the mutex? and thread-condition?  predicates slightly more expensive, but perhaps not significantly.

jessymilare:
All right, thank you for the response.

I think that it would be better if the name is a string, but a symbol would be ok too.

[...]

PR #380, opened 2019-01-10
Mutexes and condition variables with names

jessymilare:
Related to issue 372

Now `make-mutex` and `make-condition` accept an optional
argument name, which must be a symbol. When the mutex or
condition is printed, the name is also printed. The name
is immutable and can be accessed by mutex-name and
condition-name.

https://github.com/cisco/ChezScheme/issues/372


jessymilare:
Added entries to LOG and release notes and updated CSUG.

LOG
```diff
@@ -1042,3 +1042,9 @@
 - clarified required use of scheme-start to start an application
   packaged as a boot file and added a short "myecho" example.
     use.stex
+- add name fields for mutexes and condition variables, now `make-mutex` and
+  `make-condition` accept an optional argument `name`, which must be a
+  symbol.  The name is printed every time the object is printed, which is
+  usefull for debugging
```
burgerrg: usefull => useful

csug/threads.stex
```diff
@@ -123,10 +123,16 @@ in all threads.
...
 \noskipentryheader
 \formdef{make-mutex}{\categoryprocedure}{(make-mutex)}
+\formdef{make-mutex}{\categoryprocedure}{(make-mutex \var{name})}
 \returns a new mutex object
 \listlibraries
 \endnoskipentryheader
 
+\noindent
+\var{name}, if supplied, must be a symbol which identifies the mutex. The
+name is printed every time the mutex is printed, which is usefull for
```
burgerrg: useful

csug/threads.stex
```diff
@@ -207,10 +213,16 @@ Using \scheme{with-mutex} is generally more convenient and safer than using
...
 \formdef{make-condition}{\categoryprocedure}{(make-condition)}
+\formdef{make-condition}{\categoryprocedure}{(make-condition \var{name})}
 \returns a new condition object
 \listlibraries
 \endnoskipentryheader
 
+\noindent
+\var{name}, if supplied, must be a symbol which identifies the condition
+object. The name is printed every time the condition is printed, which is
+usefull for debugging.
```
burgerrg: useful

release_notes/release_notes.stex
```diff
@@ -58,6 +58,13 @@ Online versions of both books can be found at
 %-----------------------------------------------------------------------------
 \section{Functionality Changes}\label{section:functionality}
 
+\subsection{Mutexes and condition variables can have names (9.5.1)}
+
+The procedures \scheme{make-mutex} and \scheme{make-condition} now
+accept an optional argument \scheme{name}, which must be a symbol
+that identifies the object. The name is printed every time the mutex
+or condition object is printed, which is usefull for debugging.
```
burgerrg: useful

s/prims.ss
```diff
@@ -1529,15 +1533,28 @@
...
-      (mutex-guardian m)
-      m)))
+  (case-lambda
+    [()
+     (let ([m ($make-mutex (mm) #f)])
+       (mutex-guardian m)
+       m)]
+    [(name)
+     (unless (symbol? name)
+      ($oops 'make-condition "~s is not a symbol" name))
```
burgerrg: should be indented 1 more space

s/prims.ss
```diff
@@ -1561,15 +1578,28 @@
...
-      (condition-guardian c)
-      c)))
+  (case-lambda
+    [()
+     (let ([c ($make-condition (mc) #f)])
+       (condition-guardian c)
+       c)]
+    [(name)
+     (unless (symbol? name)
+      ($oops 'make-condition "~s is not a symbol" name))
```
burgerrg: should be indented 1 more space

burgerrg:
What do you think of allowing #f for the optional second argument of make-mutex and make-condition? This would make the definitions of the single-argument cases simpler and make it easier to write functions that take an optional name.

dybvig:
That's probably a good idea.  I don't see how it simplifies the single-argument cases, however.  In fact, they would need to explicitly allow `#f` as well as symbols, and the error messages would have to be adjusted.

dybvig:
@jessymilare, if you use `set-who!` rather than set! for `make-mutex`, `mutex-name`, `make-condition`, and `condition-name`, you can replace the quoted names with `who` in the error messages and just put each of the tests on one line to tighten up the code.  I've been doing that with most of the things I've added recently.

burgerrg:
For the single argument cases, I meant that one could simply call the two-argument case with #f and wouldn't need to duplicate the body.

dybvig:
Ah, I guess you mean the zero-argument cases.  Yes, but that would send the zero-argument case through an unnecessary check.  Although if one the zero-argument case calls the one-argument case, the `case-lambda` form should be wrapped in a rec, e.g., `(rec make-mutex <case-lambda form>)` to make the call local, in which case cp0 would presumably inline the one-argument code into the zero-argument code and discard the check.  Though I would probably instead just add a helper that both cases use to make clear the test isn't necessary in the zero-argument case.

burgerrg:
Thanks, I got my argument counts off by one! Yes, a helper without checks that is used by both front doors would be a good way to implement it.

jessymilare:
Sorry for the polution in the commits. I use aggressive-indent-mode and I forgot to disable it before editing.

burgerrg:
The build system found an error while building the threaded versions of Chez Scheme: [Details](https://travis-ci.org/cisco/ChezScheme/builds/510617020?utm_source=github_status&utm_medium=notification)
<pre>Warning: undeclared variable assignment to $make-mutex at line 1485, char 1 of prims.ss</pre>

jessymilare:
> The build system found an error while building the threaded versions of Chez Scheme: [Details](https://travis-ci.org/cisco/ChezScheme/builds/510617020?utm_source=github_status&utm_medium=notification)
> 
> Warning: undeclared variable assignment to $make-mutex at line 1485, char 1 of prims.ss

Fixed.

burgerrg:

Issue #379, opened 2019-01-06
Stack corruption in FFI


lockie:
Hello!
I'm using Chez Scheme in a programming competition, and some performance-critical bits of my application are written in C. FFI was working perfectly, but at some point, after I needed to call Scheme function with a lot of arguments (over a dozen) from C I seem to come across some weird stack corruption. The problem manifests itself on both `ta6le` and `a6le` Chez variants.

Here's the MCVE:

**lib.c**
```C
#include <stdint.h>
typedef void (*callback_t)(uint32_t i, const char* s1, const char* s2, const char* s3,
    const char* s4, uint32_t i2, const char* s6, const char* s7, uint32_t i3);
void test(callback_t callback)
{
    callback(0, "this", "is", "working", "just", 1, "fine", "or does it?", 2);
}
```
**test.scm**
```Scheme
(load-shared-object "./lib.so")
(define test (foreign-procedure "test" (void*) void))
(test
 (foreign-callable-entry-point
  (foreign-callable
   (lambda (i s1 s2 s3 s4 i2 s6 s7 i3)
    (display (format "i=~a~%" i))
    (display (format "s1=~a~%" s1))
    (display (format "s2=~a~%" s2))
    (display (format "s3=~a~%" s3))
    (display (format "s4=~a~%" s4))
    (display (format "i2=~a~%" i2))
    (display (format "s6=~a~%" s6))
    (display (format "s7=~a~%" s7))
    (display (format "i3=~a~%" i3))
    )
   (unsigned-32 u8* u8* u8* u8* unsigned-32 u8* u8* unsigned-32)
   void)))
```
**Makefile**
```Makefile
all: test

lib.so: lib.c
        gcc -shared lib.c -o lib.so

test: lib.so
        @scheme -q --script test.scm

.PHONY: test
```
**Test output**
```
$ make test
i=0
s1=#vu8(116 104 105 115)
s2=#vu8(105 115)
s3=#<foreign>
s4=#vu8(106 117 115 116)
i2=1
s6=#vu8(102 105 110 101)
s7=#vu8(111 114 32 100 111 101 115 32 105 116 63)
i3=2
```
(notice how the `s3` string becomes `#<foreign>`, when it really shouldn't).

I'm feeling that I'm missing something obvious. Or is it that I'm too frivolous with the callback pointer?

Some observations:
- this happens with different arguments depending on specific argument count;
- the bug seem to manifest itself when callback's argument count is larger than 8;
[...]

PR #387, opened 2019-01-27
fix string allocation for foreign-callable argument or foreign-call return

mflatt:
A repair for #379

The `dofretu...*` intrinsics use ac1 without declaring it as killed, which effectively broke register allocation for handling string/bytevector foreign-call results or callable arguments.

There was a "FIXME" in the implementation of `dofretu...*` that suggests avoiding ac1, so maybe that intent is why ac1 never got declared. With respect to the "FIXME", though, `get-room` does preserve td and ts, but I didn't find a way to avoid ac1 on x86 (because `alloc` involve ts and xp?). So, that's why the repair declares ac1 as killed by `dofretu...*` instead of trying to avoid it.

### DIFF
--- a/LOG
+++ b/LOG
@@ -1042,3 +1042,6 @@
 - clarified required use of scheme-start to start an application
   packaged as a boot file and added a short "myecho" example.
     use.stex
+- fix allocation of string/bytevector for a foreign-callable argument
+  or foreign-call return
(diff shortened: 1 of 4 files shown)


dybvig:
Thanks for the fix!

Issue #389, opened 2019-01-29
apply doesn't throw exception when last argument isn't a list


Neowizard:
In some cases, apply will throw an exception if the last argument isn't a list (as required in r6rs), while in other cases, it executes just fine.

For instance:
```
> (apply + 1)
Exception in apply: 1 is not a proper list
Type (debug) to enter the debugger.
> (apply zero? 0)
#t
```

owaddell:
This appears to be an issue with `cp0`, which is enabled by default.
```
> (apply null? 'hypothesis)
#f
> (apply zero? 0)
#t 
```
But if we disable `cp0`:
```
> (run-cp0 (lambda (f x) x)) ;; disable cp0
> (apply null? 'hypothesis)
Exception in apply: hypothesis is not a proper list
Type (debug) to enter the debugger.
> (apply zero? 0)
Exception in apply: 0 is not a proper list
Type (debug) to enter the debugger.
```

dybvig:
Thanks @Neowizard for reporting the error and @owaddell for the fix.

PR #391, opened 2019-01-30
heed app-convention to fix apply issue

owaddell:
Sketch of a fix for #389 

Is there any point in trying `#2%apply` in `fold-primref2` when `app-convention` is not `call`? Seems that if `objs-if-constant` succeeds here, but the `apply` handler didn't turn `apply` into `call`, then we're going to end up residualizing anyway, so I'm inclined to check the convention earlier in that first `cond` clause and simplify.

- [x] get review from someone who has been in this code more recently
- [x] do we need more tests?
- [x] update LOG

### DIFF
--- a/LOG
+++ b/LOG
@@ -1042,3 +1042,7 @@
 - clarified required use of scheme-start to start an application
   packaged as a boot file and added a short "myecho" example.
     use.stex
+- fixed a bug in cp0 bug that could fold the apply of a primitive, where
+  the last argument is not a list, as if it were a call to the primitive
+  with those arguments
+    cp0.ss, cp0.ms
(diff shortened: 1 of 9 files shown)


owaddell:
Thanks. I've updated my branch to add some tests and to bail earlier if `app-convention` is not `call`.

I may be bailing too early, since inline handlers won't be able to special-case based on `app-convention`. Offhand, I'm not sure who would want to do that (maybe to avoid a value-visit somewhere?).

Travis will be unhappy since I haven't yet updated the expected errors.

owaddell:
Attempted to update expected errors and restored original indentation to minimize diffs from master.

owaddell:
Added a test for the 'effect unrestricted discard case, updated LOG, and squashed.

owaddell:
Re-pushed to update the release notes.

dybvig:
Thanks for the fix!

Issue #279, opened 2018-03-23
Partial r7rs support


ecraven:
I've started writing simple wrapper libraries for `(scheme base)` and the others, most functions in R7RS are either already in Chez or trivial to implement by wrapping existing functions.
Would you be willing in general to consider merging this at some point in the future? I'm assuming there would be some necessary support for `define-library`, or the `#u8` vector syntax.
Or is R7RS something you don't want to support in any way?

dybvig:
Supporting R7RS is a fine idea as long as doing so doesn't overly complicate the expander, compiler, or run-time system.  We should also keep in mind that R7RS is not a successor to R6RS and does not replace R6RS.  While we might lament this unfortunate situation and the reactionary forces that brought it about, it is what it is.  Thus, we should maintain R6RS compatibility and emphasize support for R6RS over R7RS when choices must be made.

I understand some R7RS changes are at the lexical level and will require changes to the reader.  Those will have to be built-in and possibly enabled via a `#!r7rs` marker in the input stream, similar to the existing `#!r6rs` and `#!chezscheme` markers.  This is a good first step, since it would allow the construction of a separately loadable R7RS library that implements the non-lexical features of R7RS, some via wrappers as you suggest.

If we eventually decide to build in support for all of R7RS, we'll want to revisit the use of wrappers where they introduce overhead and/or negatively affect error reporting.

weinholt:
There's another way to get R7RS code running on Chez Scheme. I have recently added R7RS support in Akku.scm, my Scheme language package manager: https://akkuscm.org/

Essentially, Akku translates `define-library` into one `library` per implementation that appears in the `cond-expand` clauses (the `cond-expand` requested in #91). There is one caveat: unquoted vectors are still syntax errors (but R7RS also supports quoted vectors, so they can simply be added to the R7RS code). The `akku-r7rs` package provides the R7RS standard libraries as R6RS libraries.

Here's an example of how to use it: https://gitlab.com/akkuscm/akku/wikis/R7RS-example

I've tried it with complicated R7RS libraries found in the wild and have had good success. I've passed all libraries from snow-fort.org through it, and they tend to work except the ones that have implementation-specific code.

amirouche:
Here is another PR with r7rs changes https://github.com/cisco/ChezScheme/pull/323

amirouche:
See also this conversation about r7rs cond expand https://github.com/cisco/ChezScheme/issues/91

PR #396, opened 2019-02-13
Add support for R7RS define-library form

amirouche:


amirouche:
Here is the full specification of `define-library` form in [r7rs](https://small.r7rs.org/attachment/r7rs.pdf):

```scheme
〈library〉→(define-library〈library name〉〈library declaration〉*)

〈library name〉→(〈library name part〉+)

〈library name part〉→ 〈identifier〉 
  |〈uinteger 10〉

〈library declaration〉→(export〈export spec〉*)
  |〈import declaration〉
  | (begin〈command or definition〉*)
  |〈includer〉
  |(include-library-declarations〈string〉+)
  |(cond-expand〈cond-expand clause〉+)
  |(cond-expand〈cond-expand clause〉+(else〈library declaration〉*))

〈import declaration〉→(import〈import set〉+)

〈export spec〉→ 〈identifier〉
  |(rename〈identifier〉 〈identifier〉)

〈import set〉→ 〈library name〉
  |(only〈import set〉 〈identifier〉+)
  |(except〈import set〉 〈identifier〉+)
  |(prefix〈import set〉 〈identifier〉)
  |(rename〈import set〉(〈identifier〉 〈identifier〉)+)

〈cond-expand clause〉→(〈feature requirement〉 〈library declaration〉*)

〈feature requirement〉→ 〈identifier〉
  |〈library name〉
  |(and〈feature requirement〉*)
  |(or〈feature requirement〉*)
  |(not〈feature requirement〉)
```

amirouche:
TODO:

- [ ] mats
- [ ] documentation
- [ ] LOG

Support status:

- ok: `include` works like chezscheme `include`
- ok: `import` / `export` works
- maybe : `include-ci` is not supported yet
- maybe: `cond-expand` seems to me that it could clash with srfi-0, I need to read the spec properly
- maybe: number in library names

Memorytoco:
Awesome!

amirouche:
Do maintainers have interest to merge this Pull-Request?

What will be the requirements?

dybvig:
If Chez Scheme can be made to implement R7RS fully and efficiently and in a manner that doesn't overly complicate the implementation or negatively impact support for existing Chez Scheme language forms and features or support for R6RS, that's fine with me.  Alternatively, I'm open to partial support where necessary to allow a reasonable user-defined R7RS wrapper to be created.  This likely includes reader extensions, as I mentioned in response to #279, and it might require built-in support at some level for R7RS libraries and the R7RS top level if, for example, the user-defined wrapper would otherwise have to redefine the entire expander.

So there are three interesting points in the space to me: we do nothing specifically to support R7RS; we add just enough to allow a user-defined R7RS wrapper to be created; or we add full support for R7RS.  Full support means 100% compatibility, full efficiency, error messages coming from the right places, source information in error messages, the inspector, and the profiler coming from the right places, interoperability with Chez Scheme and R6RS libraries, etc.  The final option is a big job, and the best path to it probably starts with the second.  The second is a fairly big job as well, since someone will have to articulate the extensions required to support a reasonable user-defined R7RS wrapper, which probably requires writing the wrapper itself and seeing where the holes are.

With respect to this pull request, I would want it either to grow up to be a full implementation of the R7RS define-library form or to turn into just what's necessary to allow a reasonable user-level definition of define-library.

Toward the second end, can you identify what, if anything, currently prevents a reasonable user-level definition of define-library?

amirouche:
> With respect to this pull request, I would want it either to grow up to be a full implementation of the R7RS define-library form or to turn into just what's necessary to allow a reasonable user-level definition of define-library.
>
> Toward the second end, can you identify what, if anything, currently prevents a reasonable user-level definition of define-library?

The current code rely on the following:

> (global-extend 'macro 'define-library (lambda (orig) (r6rs-library (r7rs-library orig)))))

To allow user code to implement `define-library`, it would require to be able to do what `global-extend` does which is I think change the behaviour of the expander.

PS: The current code implements what is necessary to support `define-library` top level form with support for `import` and `export` whereas `include` and  `cond-expand` should be implemented in user code. I am not sure `cond-expand` fully works as R7RS intend it since it can `import` things and current behaviour would be translated into a R6RS library `import` in body form. 

PS2: Based on my experience on [Arew](https://git.sr.ht/~amz3/chez-scheme-arew) it will also require to add a `#!r7rs` to support R7RS programs but that is less an issue imo since programs are not meant to be portable?!

amirouche:
> If Chez Scheme can be made to implement R7RS fully and efficiently and in a manner that doesn't overly complicate the implementation or negatively impact support for existing Chez Scheme language forms and features or support for R6RS, that's fine with me. Alternatively, I'm open to partial support where necessary to allow a reasonable user-defined R7RS wrapper to be created. This likely includes reader extensions, as I mentioned in response to #279, and it might require built-in support at some level for R7RS libraries and the R7RS top level if, for example, the user-defined wrapper would otherwise have to redefine the entire expander.
> 
> So there are three interesting points in the space to me: we do nothing specifically to support R7RS; we add just enough to allow a user-defined R7RS wrapper to be created; or we add full support for R7RS. Full support means 100% compatibility, full efficiency, error messages coming from the right places, source information in error messages, the inspector, and the profiler coming from the right places, interoperability with Chez Scheme and R6RS libraries, etc. The final option is a big job, and the best path to it probably starts with the second. The second is a fairly big job as well, since someone will have to articulate the extensions required to support a reasonable user-defined R7RS wrapper, which probably requires writing the wrapper itself and seeing where the holes are.
> 
> With respect to this pull request, I would want it either to grow up to be a full implementation of the R7RS define-library form or to turn into just what's necessary to allow a reasonable user-level definition of define-library.

Responding to that question:
 
> Toward the second end, can you identify what, if anything, currently prevents a reasonable user-level definition of define-library?

**tl;dr: I think nothing prevents user-level definition of `define-library`.**

I think I find a way to implement R7RS `define-library` and as far as I know the rest of R7RS without requiring changes to Chez.

I think I can even achieve this:

> Full support means 100% compatibility, full efficiency, error messages coming from the right places, source information in error messages, the inspector, and the profiler coming from the right places, interoperability with Chez Scheme and R6RS libraries, etc.

With the help of [annotations](http://cisco.github.io/ChezScheme/csug9.5/syntax.html#g108), in particular with the help of [`open-source-file`](http://cisco.github.io/ChezScheme/csug9.5/syntax.html#./syntax:s71) or using [laesare](https://github.com/weinholt/laesare).

As far as I understand, it boils down to rewrite a given `program.scm` and include dependencies inside Chez's [`module`](http://cisco.github.io/ChezScheme/csug9.5/syntax.html#./syntax:h5), preserve source annotations and call `eval` or `compile`.

amirouche:
I made good progress on arew scheme, not perfect but getting there.

PR #399, opened 2019-02-14
Avoid memory leak on buffer s used for putenv

pmatos:


pmatos:
This is actually broken, we shouldn't `free` the space passed into the POSIX `putenv`. Retracting.

PR #401, opened 2019-02-15
Make true flag in primdata.ss automatic

gus-massa:
The main idea is to use the signatures in [primdata.ss](https://github.com/cisco/ChezScheme/blob/master/s/primdata.ss) to automatically generate the `true` flag for the primitives. To avoid loosing information, it's necesary to make a few signatures more specific. The patch is split in many commits, because some of them can be merged individual, in particular (2), (3) and (6).

### (1) Squashed version of PR #296 about single-valued flag https://github.com/cisco/ChezScheme/commit/989a95d1d16df37fe1279201fa0372cf802ae3b2

It's just a squashed version of the other PR. Most of the other commit don't depend on this.

### (2) fix signatures with errors https://github.com/cisco/ChezScheme/commit/b34c73067b5da79aa8ab7bacd6d82460fe2f8000

These are some errors I found in the signatures. It should be merged ASAP. Actually it it not urgent because these functions are seldom used in test position. I'm only worried about

    (if (fasl-read port) 'always 'never) ==> (begin (fasl-read port) 'always)

I should take a look at `vector-for-each `and `string-for-each` that changed recently.

### (3) use signatures with eof/something https://github.com/cisco/ChezScheme/commit/2ea5df12b65463f2b94073f629839615df4e15d0

The arguments/results of the IO primitives use `eof` or something useful a lot. This was suggested while fixing the signature of `read-char` or a similar primitive.
 
### (4) use more specific signatures to imply true https://github.com/cisco/ChezScheme/commit/bca44bacf609e48318effcb07da61b6b7df35e78

Some assorted changes of some assorted primitives. Perhaps some types are too specific and it's better to create a new type `true` that can be anything except `#f`.

### (5) make true flag automatic https://github.com/cisco/ChezScheme/commit/e78522c62a254ab608b379806fa898191b702363

This is the main part, use the result of the signature to infer the `true` flag.

### (6) rename who -> maybe-who https://github.com/cisco/ChezScheme/commit/32713e9cbcfc52eb1260caf00f5a127cf70b8473

Just a minor style choice. The type `who` was unusual because it could be `#f`. It's more clear if it is renamed to `maybe-who`.

Also, what is the difference between `sint` and `exact-integer`? I think they are equivalent.

### (7) add unnecessary true flags https://github.com/cisco/ChezScheme/commit/4efa0f2331fd0784ca8d2dcdd6b04bb93fc99c77

Add the explicit `true` flag to all the primitives that don't have it. This should not be merged, I think it's better to go in the other direction and remove all the explicit `true` flags.

### (8) add check to enforce synchronization of the true flag https://github.com/cisco/ChezScheme/commit/022c039215292aff3becf550c84fedc9ca798611

A temporal check that all the `true` flags and signatures agree. This should not be merged, but it is useful to discover inconsistencies.


akeep:
I'm curious what the payoff of this change is for Chez Scheme users.  The log entry specifically calls out `begin0`, which is not a form in Chez Scheme.

Worth noting, the examples of how the `cp0` optimizations occur with this change, would already be optimized by the `np-push-mrvs` pass which pushes the multiple-return value consumer into the producer.  This pass is based on "An Efficient Implementation of Multiple Return Values in Scheme" by Ashley & Dybvig, with some improvements that allow us to avoid creating closures in more places.

gus-massa:
> I'm curious what the payoff of this change is for Chez Scheme users. 

The idea is to avoid some duplication of the information in primdata.ss. There are some minor inconsistencies, like functions that have a result that is clearly truthy  but don't have the `true` flag or functions that have `ptr` as result and they have the `true` flag. The idea is to remove duplication to reduce the number of mistakes.

---

> The log entry specifically calls out begin0, which is not a form in Chez Scheme.

This is a leftover of a previous PR https://github.com/cisco/ChezScheme/pull/296 It's somewhat unrelated, but also increase the use of the information in primdata.ss .  (I was optimistic and thought the previous PR would be merged before, and both PR make small changes in primdata.ss, so it's painful to rebase them separately, so I wrote this on top of the other. But I think both can be analyzed independently. )

I wrote `begin0` as a lazy shorthand for "a macro that has to handle single and multiple return cases, and it would be nice that the compiler removes the overhead when possible". IIRC while writing this PR I added a `display` to show when it's used, and cpnanopass.ss makes many similar constructions that this PR can simplify.

I'll take a look at `np-push-mrvs`. Anyway, I prefer that this kind of reductions are applied as early as possible, so other reductions can benefit and compose with it.

akeep:
Okay, I'll give it another within the next couple of weeks.  I agree that if we can reap additional benefits it is worth doing earlier in the compiler.  I'm afraid I don't have a great setup for benchmarking right now, and I might try to set something up, though admittedly it is unlikely we have any standard benchmarks that really exercise the multiple return value support.

Thanks for getting back to me!

mflatt:

PR #406, opened 2019-03-12
expose a library-search-handler parameter

owaddell:
This pull request exposes a `library-search-handler` parameter to permit more control over the search for libraries during `import`, `compile-whole-library`, and `compile-whole-program`.


csug/libraries.stex
```diff
@@ -998,6 +998,35 @@ When the new parameter \scheme{import-notify} is set to a true value,
...
+\formdef{library-search-handler}{\categorythreadparameter}{library-search-handler}
+\listlibraries
+\endentryheader
+
+When searching for library source or object code, \scheme{import},
+\scheme{compile-whole-program}, and \scheme{compile-whole-library}
+use the value of the \scheme{library-search-handler} parameter.
+This parameter must be set to a procedure that accepts four arguments:
+a symbol that provides context in \scheme{import-notify} messages,
+a list naming the desired library,
```
akeep: I wonder if we should just call this a library name.  I noticed that is the language we use in Section 10.3 in CSUG (and 10.2 in TSPL).

s/syntax.ss
```diff
@@ -4692,6 +4692,17 @@
...
+      (let-values ([(src-path obj-path obj-exists?) ((library-search-handler) who path dir* all-ext*)])
+        (unless (or (not src-path) (string? src-path))
+          ($oops 'library-search-handler "returned invalid source-file path ~s" src-path))
+        (unless (or (not obj-path) (string? obj-path))
+          ($oops 'library-search-handler "returned invalid object-file path ~s" obj-path))
+        (when (and obj-exists? (not obj-path))
+          ($oops 'library-search-handler "claimed object file was found but returned no object-file path"))
+        (values src-path obj-path obj-exists?))))
+
+  (define default-library-search
```
akeep: I think it might be nice to expose the `default-library-search` as a prim in the same way `library-search-handler` is exposed.  That way, if a programmer wants to avoid any other hooks that might have been put on the `library-search-handler` it is possible to get back to the base library search functionality.

owaddell:
Thanks for the quick feedback. I force-pushed changes to incorporate your comments.

I exposed `default-library-search-handler` which turned out to be a good thing all around. It let me be more precise with `$oops`. It enlisted `check-prim-arg-errors` which caught me being lax in checking the library name. It may even have clarified the documentation.

akeep:
Looks good, thanks for putting in those changes!

PR #410, opened 2019-03-20
Add LZ4 support and use it by default for compressing files

mflatt:
This patch speeds up compressed-code loading (and decompression in general) by using LZ4 in place of zlib. Loading LZ4-compressed code on my machine takes about the same time as loading uncompressed code — in contrast to decompression via zlib, which makes loading take about twice as long.

This patch does not remove the use of zlib, but it adds LZ4 and makes it the default. Setting the `compress-format` parameter to `'gzip` makes compression use zlib instead of LZ4. Reading a compressed file infers the compression format, so `compress-format` has no effect on loading/reading.

The LZ4 implementation is pulled in via a submodule, analogous to zlib, but I added `ZLIB` and `LZ4` options to `configure` to enable selecting an existing zlib and/or lz4 build. The LZ4 source does not appear to have support for a non-DLL Windows build, so that part's especially ugly.


burgerrg:
How do we build the Scheme side? I get invalid memory accesses when using make, and with the patch file technique, I get unexpected primref records from $uncprep.

mflatt:
@burgerrg Thanks for the repair, and I'll rename "compress.*" soon.

I expect a build in a clean checkout to work (i.e., with the old checked-in bootfiles). Assuming that you don't already have some older bootfiles in your checkout, I don't know what went wrong. Technically, the checked-in boot files will call some kernel functions without a new boolean argument while bootstrapping, but a random value for that boolean should be harmless and merely cause a compression format to be selected at random for some intermediate steps.

burgerrg:
I did a clean build on both macOS and Windows but couldn't get it to work.

On macOS make in the s directory yields:
```
echo '(reset-handler abort)'\
             '(base-exception-handler (lambda (c) (fresh-line) (display-condition c) (newline) (reset)))'\
             '(keyboard-interrupt-handler (lambda () (display "interrupted---aborting\n") (reset)))'\
             '(optimize-level 3)'\
             '(debug-level 0)'\
             '(commonization-level (commonization-level))'\
             '(generate-inspector-information #f)'\
             '(subset-mode (quote system))'\
             '(compile-file "cmacros.ss" "cmacros.so")'\
             | ../bin/a6osx/scheme -q
compiling cmacros.ss with output to cmacros.so
Exception in fasl-write: invalid fasl object #<foreign>
```

On Windows make in the s directory yields:
```
echo '(reset-handler abort)'\
             '(base-exception-handler (lambda (c) (fresh-line) (display-condition c) (newline) (reset)))'\
             '(keyboard-interrupt-handler (lambda () (display "interrupted---aborting\n") (reset)))'\
             '(optimize-level 3)'\
             '(debug-level 0)'\
             '(commonization-level (commonization-level))'\
             '(generate-inspector-information #f)'\
             '(subset-mode (quote system))'\
             '(compile-file "cmacros.ss" "cmacros.so")'\
             | ../bin/a6nt/scheme.exe -q
Error in foreign-procedure: no entry for ~s
("(cs)find_pcode")
```

mflatt:
@burgerrg Still no idea about the build problems, since things work for me from a clean build on macOS, Windows (WSL), and Linux, and the build works on Travis-CI. I note that `find_pcode` was removed in 68e20f7216 and the bootfiles were updated in 4b42811557. Maybe it will help somehow that I rebased to pull in yesterday's commit.

mflatt:
I see that I overlooked lots of compiler warnings on Windows. I'll fix those.

burgerrg:
The changes look good! I built the Windows versions and linked them against another library successfully. Please rebase onto master.

mflatt:
Thanks! Rebased.

dybvig:
This looks good to me---just a couple of nits:

- In primdata.ss, the second signature for compress-format should be [(sub-symbol) -> (void)].
- The copyright notice on compress-io.c identifies the file as system.h.
- The copyright notice on compress-io.h is missing.

I like the new mat parameters subform.

I ran some quick tests on my machine (with solid state drive): start-up time (presumably mostly loading the boot files) dropped by about 9% with the change.  Bootstrapping time also dropped, by 6%.  Nice!  Though start-up time still took nearly twice as long for me with lz4-compressed boot files as with uncompressed boot files.

dybvig:
Correction: start-up time with lz4 compression is actually 35% faster for me than with zlib compression and just 31% slower than with no compression.  (It turns out I was getting zlib-compressed boot files as a result of bootstrapping in an existing workarea.)

mflatt:
Thanks for the copyright and "primdata.ss" repairs – now pushed.

PR #411, opened 2019-03-21
record revision control information alongside boot files

owaddell:
This pull request aims to help answer support questions like:
1. what source did I build in this particular workarea?
1. what source was used to build this installed binary?

My original stab at this exposed a `#%$scheme-revision` that we could get at via the repl. That was convenient at run time, but the extra rounds of bootstrapping made that approach irritating at compile time.

I've since scaled back to the much simpler approach of recording minimal revision-control information in a `revision` file alongside the boot files. This seems to work well enough for my use cases.

The `update-revision` script could be extended for folks who mirror the git source in some other revision control system. The git attributes on `update-revision` insert the appropriate revision if folks export the source code via `git archive` or download a tarball or zip file from github.

The WiX updates are guesswork, but seemed to work.

### DIFF
--- a/LOG
+++ b/LOG
@@ -1247,3 +1247,7 @@
     newrelease pkg/Makefile release_notes.stex
 - update Windows spin-loop count for deleting files and directories
     windows.c
+- install a file containing revision information alongside boot files;
+  embed git revision in exported source archives
+    bintar/Makefile Mf-install.in wininstall/*nt.wxs
+    s/Mf-base s/.gitattributes s/update-revision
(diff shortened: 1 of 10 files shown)


owaddell:
Thanks for looking this over, Bob. I updated the LOG and rebased.

Before merging anything from this branch, I'd like to know whether @dybvig is open to merging 
only d0c9ad1a37adfa7c746a4ad1388816380654b905, or
1. squashing in _variation 1_ 162c044d7a0e9d56b197d95119c6112d511aabb1, or
2. squashing in _variation 2_ a8ba65329ae752886d7babedfc16cce13080729f, or
3. squashing in both _variation 1_ and _variation 2_ with that base commit

... or whether he wants no part of this. It has been really handy for our internal development.

dybvig:
This seems like a good idea.  I would go for squashing in variation 1 162c044 (with experimental hg support).

PR #418, opened 2019-04-11
Add a compress-level parameter

dybvig:
This patch adds a `compress-level` parameter that can be used to control the amount of effort expended compressing, whether with the default LZ4 compression or gzip compression.  The parameter can take on one of the symbolic values `low`, `medium`, `high`, or `maximum`.   The default is medium, which for LZ4 currently maps to `LZ4HC_CLEVEL_MIN` and is both more effective and slower (at compression, but not decompression) than the previous default.  The patch also makes several changes to the LZ4-support code added in #410; in particular, it restructures the code a bit, puts the compress format directly in the thread context (which now also holds the compress level), and changes how the compressed bytevector compression format and length are stored in the header word.  I have tested the changes on OSX and Linux but not Windows.


mflatt:
These changes look good to me.

(Just to confirm, Racket-on-Chez builds fine with the patch, its boot file size goes down by about 17%, and load time remains about the same and maybe slightly faster.)

c/externs.h
```diff
@@ -169,21 +169,42 @@ extern wchar_t *S_malloc_wide_pathname PROTO((const char *inpath));
 #endif
 extern IBOOL S_fixedpathp PROTO((const char *inpath));
 
+/* compress-io.c */
+extern glzFile glzdopen_output PROTO((INT fd, INT compress_format, INT compress_level));
+extern glzFile glzdopen_input PROTO((INT fd));
+extern glzFile glzopen_input PROTO((const char *path));
+#ifdef WIN32
+extern glzFile glzopen_input_w PROTO((wchar_t *path));
```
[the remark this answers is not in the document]
dybvig: Done

c/new-io.c
```diff
@@ -268,38 +259,28 @@ ptr S_compress_input_fd(INT fd, I64 pos) {
...
-#else
-  as_append = fcntl(fd, F_GETFL) & O_APPEND;
-#endif
+  ptr tc = get_thread_context();
 
-  if (as_gz)
-    file = glzdopen_gz(fd, as_append ? "ab" : "wb");
-  else
-    file = glzdopen_lz4(fd, as_append ? "ab" : "wb");
+  file = glzdopen_output(fd, (int)UNFIX(COMPRESSFORMAT(tc)), (int)UNFIX(COMPRESSLEVEL(tc)));
```
burgerrg: The glzdopen_output function prototype uses INT instead of int, so wouldn't you want the casts to use INT instead of int also?

c/new-io.c
```diff
@@ -358,26 +339,23 @@ static ptr new_open_output_fd_helper(
   }
 
   glzFile file;
-  if (as_gz)
-    file = glzdopen_gz(fd, append ? "ab" : "wb");
-  else
-    file = glzdopen_lz4(fd, append ? "ab" : "wb");
+  file = glzdopen_output(fd, (int)UNFIX(COMPRESSFORMAT(tc)), (int)UNFIX(COMPRESSLEVEL(tc)));
```
burgerrg: Use INT instead of int.

burgerrg:
All the Windows mats passed!

dybvig:
Thanks for the corrections!

c/compress-io.c
```diff
@@ -236,334 +284,389 @@ glzFile glzdopen_input(INT fd) {
...
+    return -1;
+  }
 
-    while (lz4->out_pos < lz4->out_len) {
-      INT r = WIN32_IZE(write)(lz4->fd, (char*)lz4->out_buffer + lz4->out_pos, lz4->out_len - lz4->out_pos);
-      if (r >= 0)
-        lz4->out_pos += r;
+  out_pos = 0;
+  while (out_pos < out_len) {
+    r = WIN32_IZE(write)(lz4->fd, out_buffer + out_pos, out_len - out_pos);
```
burgerrg: The WIN32 definition of `_write` uses `unsigned int` for the count, not `size_t`. As a result, we get a compiler warning. Here's one way to fix it:

<pre>#ifdef WIN32
    r = _write(lz4->fd, out_buffer + out_pos, (UINT)(out_len - out_pos));
#else
    r = write(lz4->fd, out_buffer + out_pos, out_len - out_pos);
#endif</pre>

c/thread.c
```diff
@@ -224,7 +226,9 @@ static IBOOL destroy_thread(tc) ptr tc; {
         }
       }
 
-      free((void *)THREADTC(thread));
+      if (LZ4OUTBUFFER(tc) != NULL) free(LZ4OUTBUFFER(tc);
```
burgerrg: This line is missing a close parenthesis:
<pre>free(LZOUTBUFFER(tc));</pre>

owaddell:
Thanks for all the changes on this branch! Valgrind's memcheck tool is happy now.

Note typo on line 85 of compress-io.c: "lcoal" -> "local".

c/compress-io.c
dybvig: Thanks!  I ended up fixing this one by using INT for the lengths and positions.

burgerrg:
It compiles cleanly in Microsoft Visual Studio, and so far the ta6nt mats are passing.

PR #421, opened 2019-04-16
Fix record-ref reduction in cp0

gus-massa:
In expressions like

    (record-ref ... (begin (newline) (record ...)))

the current reduction was dropping the possible side effect expressions
in this case the `(newline)`.

 A complete example is

    (run-cp0 (lambda (cp0 x) (cp0 (cp0 (cp0 x))))]) ; three times
    
    (let ()
      (define-record mybox (val))
      (mybox-val (begin (display 1) (make-mybox 2))))

==> `2` instead of `12`

---

As a side note, running only two times the cp0 pass, it is not enough to reduce some expressions like `(mybox-val (make-mybox 2)))`. It is necessary to run cp0 three times. I don't know after expansions and inlining if this kind of pattern is common. I tried compiling with three times cp0 instead of two and the compilation time changes very little. Perhaps it could be useful to change the default.

### DIFF
--- a/LOG
+++ b/LOG
@@ -1337,3 +1337,9 @@
     externs.h, compress-io.c, new-io.c, scheme.c, fasl.c
 - added entries for mutex-name and mutex-thread
     threads.stex
+- fix record-ref reduction in cp0
+  in expressions like
+    (record-ref ... (begin (newline) (record ...)))
+  the reduction was dropping the possible side effect expressions
+  in this case the (newline).
+    cp0.ss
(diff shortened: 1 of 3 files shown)


dybvig:
Thanks for the bug report and fix, which looks correct.

Please pull the common `non-result-exp` call outside of the `if` on line 4711 of cp0.ss and rebase.

Were you able to come up with a mat for the second of the two cases that were missing the `non-result-exp` calls?  I tried a few things but something else seems to be pulling the effect out so that it was residualized properly without `non-result-exp`.

There are always things an extra pass of cp0 will pick up.  My guess is cases like this aren't particularly common, and it takes only one or two calls to cp0 to when the field is immutable.

gus-massa:
Done.

I also changed in the test two `equal?` to `member?`, because there are two possible correct results.

dybvig:
Looks good, thanks.

PR #422, opened 2019-04-21
fix uses of `fxzero?` on a 32-bit immediate in 32-bit mode

mflatt:
Building the compiler in safe mode for 32-bit x86 reveals incorrect uses of `fxzero?` in "x86.ss": a 32-bit immediate is not necessarily a fixnum.

For example, compilation of "newhash.ss" triggers an error, since it involves an `fx+` on 523658599 which turns into an inline `+` on the immediate 2094634396.

The misuses of `fxzero?` cause no trouble when compiled as unsafe, since that turns into a `eq?` that works as intended in this case.

### DIFF
--- a/LOG
+++ b/LOG
@@ -1353,3 +1353,6 @@
   irritants in the original call when who or message is found not to
   be of the right type.
     exceptions.ss
+- fix incorrect uses of fxzero? x86.ss backend, since a 32-bit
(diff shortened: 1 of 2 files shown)


burgerrg:
Thanks for finding and fixing this bug!

PR #423, opened 2019-04-22
Small fix in relop-length

gus-massa:
Make two small fixes in `relop-length` for the mirrored case.

Also, in the test replace `(and ...)`  with `(list ...)`, so they detect any discrepancy.

---

Note: While debugging, I tried to test this in scheme.exe adding a few `(display <something>)` here and there. But I couldn't see any change in the output. Do I have to turn on some parameter like `likely-to-be-compiled`?

### DIFF
--- a/LOG
+++ b/LOG
@@ -1368,3 +1368,5 @@
     primdata.ss
 - small fix in relop-length to enable the optimization
     cpnanopass.ss
+- make test for relop-length more sensitive
+    5_2.ms
--- a/mats/5_2.ms
+++ b/mats/5_2.ms
@@ -188,7 +188,7 @@
           [(_ prim)
            (let ()
              (define (f x)
-               (and
+               (list
                  (prim (#3%length x) 0)
                  (prim 0 (#3%length x))
                  (prim (#3%length x) 1)
@@ -204,7 +204,7 @@
                  (let ([n (length x)])
                    (equal?
                      (f x)
-                     (and
+                     (list
                        (prim n 0)
                        (prim 0 n)
                        (prim n 1)


burgerrg:
What is wrong with the existing code?

gus-massa:
The `relop-length2` function in [line#L3767](https://github.com/cisco/ChezScheme/blob/master/s/cpnanopass.ss#L3767) is only useful to reduce expressions where the second argument is an explicit small number, for example in `(eq? (length l) 2)` but it is not useful for expressions like `(eq? 2 (length l))`.

So the last reduction attempt in `relop-length` in [line#L3775](https://github.com/cisco/ChezScheme/blob/master/s/cpnanopass.ss#L3775) tries to call first the `relop-length2` with the original order, and then with the reversed order. But here is an error and the arguments are not swapped in the second call.

To make the second call equivalent to the first one, it is necessary to mirror the comparison argument. But there is an error in the numbering of the compactors ([lines#L3718-L3723](https://github.com/cisco/ChezScheme/blob/master/s/cpnanopass.ss#L3718-L3723)), so when `(< 2 (length l))` is mirrored, it is transformed to `(>= (length l) 2)` instead of `(> (length l) 2)`.

burgerrg:
I noticed that, too, but was surprised when I set `(#%$assembly-output #t)` that it generated the code I expected. Have you found any cases where the current code gets the incorrect answer?

Use `(compile-interpret-simple #f)` at the REPL to make sure small expressions go through the compiler instead of the interpreter.

burgerrg:
OK, I figured out what happened. A recent change to cpnanopass.ss introduced the relop-length optimization. In Chez Scheme v9.5.2, good code was emitted for (<= 2 (#3%length ls)), but not with the recent optimization.

Since the relops return booleans, using `and` works well enough to flag trouble, so let's not modify 5_2.ms at this time.

Thank you very much for finding and fixing this optimization defect!

burgerrg:
Fixed in 587e0f4c01bf6bc187fc4e829fba9cc8ca7c38dd

gus-massa:
The problem of using `and` to compare the results is that it only notices a change when one of the versions is always `#t` . Probably `>=` when the length of the sample string is `0`, and `<=` when the length of the sample string is `100`, or something like that. Only a few cases in the borders are sensible to errors, most cases have at least and easy `#f` that shadows any possible error in the other tests.

My solution to use `list` was a little lazy, it is enough to find any error. But perhaps it is better to rewrite the test in another way.

burgerrg:
Thanks, I now understand the value of the change to the mats. Would you mind rebasing this pull request to include the mats change?

gus-massa:
Done.

It would be good to add `(compile-interpret-simple #f)` somewhere in the test suit, so the test check also this pass.

burgerrg:
Thanks, and I verified that the mats set compile-interpret-simple to #f for most cases. See [mats/Mf-base](https://github.com/cisco/ChezScheme/blob/master/mats/Mf-base#L99) for details.

PR #425, opened 2019-05-01
repair multiply of (- (expt 2 30)) to itself

mflatt:
On a 64-bit platform, the test for "short" arguments to
avoid overflow was incorrect, because `(- (expt 2 30))`
counted as short.


burgerrg:
Thanks for finding and fixing this bug! Please add a paragraph in the release notes for the fixed bug.

c/number.c
```diff
@@ -637,7 +637,7 @@ static ptr big_mul(tc, x, y, xl, yl, sign) ptr tc, x, y; iptr xl, yl; IBOOL sign
   return copy_normalize(&BIGIT(W(tc),0),xl+yl,sign);
 }
 
-#define SHORTMIN (most_negative_fixnum / (1 << (fixnum_bits / 2)))
+#define SHORTMIN ((most_negative_fixnum / (1 << (fixnum_bits / 2))) + 1)
```
burgerrg: Is it worth doing something to make the range tighter for 32-bit systems? Maybe something like
<pre>#define SHORTMIN ((most_negative_fixnum / (1 << (fixnum_bits / 2))) + (fixnum_bits % 1))</pre>

burgerrg: I suppose on 32-bit systems SHORTMIN = -23170, which is approx. -sqrt(2^29-1)

burgerrg: And SHORTMAX = 23170 on 32-bit systems.

burgerrg: <pre>/* SHORTRANGE is -floor(sqrt(most_positive_fixnum))..floor(sqrt(most_positive_fixnum)).
   We don't use sqrt because it rounds up for fixnum_bits = 61 */
#if (fixnum_bits == 30)
#define SHORTRANGE(x) (-23170 <= (x) && (x) <= 23170)
#elif (fixnum_bits == 61)
#define SHORTRANGE(x) (-0x3FFFFFFF <= (x) && (x) <= 0x3FFFFFFF)
#endif
</pre>

mflatt:
The refinement to `SHORTMIN` seems good, and I assume you meant `% 2` instead of `% 1`.

I can add a note to the release notes, but I'm a little surprised by the request. The convention seems to be that feature additions and changes are listed there, but not bug fixes. So, just to make sure: bug reports should generally be included in "release_notes.stex"?

burgerrg:
Yes, bug fixes go into a section labeled section:bugfixes in release_notes.stex.

I proposed simpler C code for SHORTRANGE and a brief description of how the bounds were determined. The mats may need to be updated to test the edges in a 32-bit system.

mflatt:
I see. But I think I should probably just close the PR and leave it to you. Thanks!

burgerrg:
Thanks again for finding and fixing this bug!

PR #429, opened 2019-05-14
Fix signatures of $file-options and friends

gus-massa:
They were defined inside a block with the `proc` flag, but they are `enum-set`s.

This confuses `cp0`. For example

    (enum-set? #3%$file-options) ; ==> #t
    (procedure? #3%$file-options) ; ==> #t (but should be #f)
    (run-cp0 (lambda (cp0 x) x)) ; disable cp0
    (procedure? #3%$file-options) ; ==> #f


burgerrg:

PR #430, opened 2019-05-20
Correct a simple typo in csug

teeghee:
Too small to pull-request, but anyways...


burgerrg:
Thanks, I added an entry in LOG and squashed into master.

PR #437, opened 2019-06-14
Changes to allow ChezScheme to build on FreeBSD

abmclin:
In this PR are commits which allow ChezScheme to build on FreeBSD successfully.

I have only tested those commits on `a6fb` and `ta6fb` variants.

Some of the changes have been motivated by gcc's `-Wimplicit-fallthrough` which is enabled by `-Wextra` in gcc 7 or above. It has been already commented in other issue tickets that the warning can be suppressed by adding `-Wno-implicit-fallthrough` to relevant Mf-* makefiles under `c` directory. I'm taking a different approach, changing the source code to resolve the warning. I think this approach is more robust because this should automatically work for all platforms and this avoids requiring people to send in makefile corrections one by one via PRs whenever they encounter the same issue for their own platform.

The main change is addition of `NORETURN` macro to `externs.h` which expand to `___attribute__((noreturn))` or `__declspec(noreturn)` depending on compiler type.

gcc 7 or above will raise implicit fallthrough warnings when certain functions are used in switch case statements. If it has no way of knowing those functions won't return then it assumes fallthrough could occur.

Based on visual inspection of the relevant definitions for the various S_error* and friends in `schsig.c` and directly calling those functions in my test cases, as far as I can tell those functions don't return. Annotating those functions with `NORETURN` help gcc' static analysis.

Also in some places I added `/* fall through */` comments as appropriate since gcc 7 or above will recognize certain comments as signaling that fallthrough is intended.

Other commits deal with fixing some things and adding `-luuid` as needed.

### DIFF
--- a/c/externs.h
+++ b/c/externs.h
@@ -45,6 +45,18 @@ off64_t lseek64(int,off64_t,int);
 #include <setjmp.h>
 #endif
 
+#ifndef NORETURN
+
+#if defined(__GNUC__) || defined(__clang__)
+#define NORETURN __attribute__((noreturn))
+#elif defined(_MSC_VER)
+#define NORETURN __declspec(noreturn)
+#else
+#define NORETURN
+#endif /* defined(__GNUC__) || defined(__clang__) */
+
+#endif /* NORETURN */
+
 /* external procedure declarations */
 /* prototypes gen. by ProtoGen Version 0.31 (Haydn Huntley) 1/18/93 */
 
@@ -304,14 +316,14 @@ extern void S_handle_overflow PROTO((void));
 extern void S_handle_overflood PROTO((void));
 extern void S_handle_apply_overflood PROTO((void));
 extern void S_overflow PROTO((ptr tc, iptr frame_request));
-extern void S_error_reset PROTO((const char *s));
-extern void S_error_abort PROTO((const char *s));
-extern void S_abnormal_exit PROTO((void));
-extern void S_error PROTO((const char *who, const char *s));
-extern void S_error1 PROTO((const char *who, const char *s, ptr x));
-extern void S_error2 PROTO((const char *who, const char *s, ptr x, ptr y));
-extern void S_error3 PROTO((const char *who, const char *s, ptr x, ptr y, ptr z));
-extern void S_boot_error PROTO((const ptr who, ptr s, ptr args));
+extern void S_error_reset PROTO((const char *s)) NORETURN;
+extern void S_error_abort PROTO((const char *s)) NORETURN;
+extern void S_abnormal_exit PROTO((void)) NORETURN;
+extern void S_error PROTO((const char *who, const char *s)) NORETURN;
+extern void S_error1 PROTO((const char *who, const char *s, ptr x)) NORETURN;
+extern void S_error2 PROTO((const char *who, const char *s, ptr x, ptr y)) NORETURN;
+extern void S_error3 PROTO((const char *who, const char *s, ptr x, ptr y, ptr z)) NORETURN;
+extern void S_boot_error PROTO((const ptr who, ptr s, ptr args)) NORETURN;
 extern void S_handle_docall_error PROTO((void));
 extern void S_handle_arg_error PROTO((void));
 extern void S_handle_nonprocedure_symbol PROTO((void));
@@ -319,7 +331,7 @@ extern void S_handle_values_error PROTO((void));
 extern void S_handle_mvlet_error PROTO((void));
 extern void S_register_scheme_signal PROTO((iptr sig));
 extern void S_fire_collector PROTO((void));
-extern void S_noncontinuable_interrupt PROTO((void));
+extern void S_noncontinuable_interrupt PROTO((void)) NORETURN;
 extern void S_schsig_init PROTO((void));
 #ifdef DEFINE_MATHERR
 #include <math.h>
--- a/c/externs.h
+++ b/c/externs.h
@@ -49,8 +49,6 @@ off64_t lseek64(int,off64_t,int);
 
 #if defined(__GNUC__) || defined(__clang__)
 #define NORETURN __attribute__((noreturn))
-#elif defined(_MSC_VER)
-#define NORETURN __declspec(noreturn)
 #else
 #define NORETURN
 #endif /* defined(__GNUC__) || defined(__clang__) */
--- a/c/externs.h
+++ b/c/externs.h
@@ -49,6 +49,8 @@ off64_t lseek64(int,off64_t,int);
 
 #if defined(__GNUC__) || defined(__clang__)
 #define NORETURN __attribute__((noreturn))
+#elif defined(_MSC_VER)
+#define NORETURN __declspec(noreturn)
 #else
 #define NORETURN
 #endif /* defined(__GNUC__) || defined(__clang__) */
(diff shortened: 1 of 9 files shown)


c/externs.h
```diff
@@ -45,6 +45,16 @@ off64_t lseek64(int,off64_t,int);
 #include <setjmp.h>
 #endif
 
+#ifndef NORETURN
+
+#if defined(__GNUC__) || defined(__clang__)
+#define NORETURN __attribute__((noreturn))
+#else
```
pmatos: Why is there no MSVC clause with `__declspec(noreturn)`? 
In https://github.com/cisco/ChezScheme/blob/master/BUILDING it says it is possible to build with MSVC.

pmatos:
Thanks for this PR, it is revelant even outside of the context of FreeBSD as the `NORETURN` annotations help GCC8 stop false positive warning about use of uninitialized variables.

pmatos:
I just confirmed that this PR fixes (due to the `NORETURN` addition) building Chez on a x86_64 with gcc8 at `-O3`. HEAD of master returns:
```
scheme.c: In function ‘find_boot’:
scheme.c:633:8: error: ‘n’ may be used uninitialized in this function [-Werror=maybe-uninitialized]
     if (n != scheme_version) {
```

With this PR it compiled successfully.

akeep:
Thanks for the pull request!  I've been trying to test this on a FreeBSD 12.0 release, but after applying the patch when I try to build, the C section of the compile fails reporting:

```
(cd a6fb && gmake build)
(cd c && gmake)
ld -r -X -o ../boot/a6fb/kernel.o statics.o segment.o alloc.o symbol.o intern.o gcwrapper.o gc-ocd.o gc-oce.o number.o schsig.o io.o new-io.o print.o fasl.o stats.o foreign.o prim.o prim5.o flushcache.o schlib.o thread.o expeditor.o scheme.o compress-io.o i3le.o ../zlib/libz.a ../lz4/lib/liblz4.a
ld: error: section type mismatch for .eh_frame
>>> ../lz4/lib/liblz4.a(lz4.o):(.eh_frame): SHT_X86_64_UNWIND
>>> output section .eh_frame: SHT_PROGBITS

ld: error: section type mismatch for .eh_frame
>>> ../lz4/lib/liblz4.a(lz4frame.o):(.eh_frame): SHT_X86_64_UNWIND
>>> output section .eh_frame: SHT_PROGBITS

ld: error: section type mismatch for .eh_frame
>>> ../lz4/lib/liblz4.a(lz4hc.o):(.eh_frame): SHT_X86_64_UNWIND
>>> output section .eh_frame: SHT_PROGBITS

ld: error: section type mismatch for .eh_frame
>>> ../lz4/lib/liblz4.a(xxhash.o):(.eh_frame): SHT_X86_64_UNWIND
>>> output section .eh_frame: SHT_PROGBITS
gmake[2]: *** [Makefile:35: ../boot/a6fb/kernel.o] Error 1
gmake[1]: *** [Makefile:20: build] Error 2
gmake: *** [Makefile:19: build] Error 2
```

I've traditionally built on FreeBSD by aliasing `make` to `gmake`.

Is there something special I need to set to get the `lz4` pieces to build correctly?

abmclin:
@akeep I had the same error, worked around it and then completely forgot all about it.

The problem appears to be due to ChezScheme's Makefile directly calling `ld` and on FreeBSD the `ld` corresponds to Clang's `ld` installed at `/usr/bin` not GCC's `ld` which is actually installed in `/usr/local/bin`

I was able to work around that by arranging for GCC's `/usr/local/bin/ld` to be called instead which makes the problem go away. But I recently learned that ChezScheme is using `ld` to process *.o object files which apparently is a dicey thing to do.

There is a solution posted at https://github.com/racket/racket/issues/2704 where Matthew Flatt solves that by changing `Makefile.in` so `ld` creates `libkernel.a` instead of `kernel.o`. I will try that solution and see if it solves the current problem.

c/externs.h
abmclin: I took it out because I noticed that it was breaking the Travis Windows CI builds and I hadn't actually tested that piece so didn't want to leave it in if it broke the CI build.

When I get a chance I'll make arrangements to try building ChezScheme on Windows and test it out. If I'm able to get it working I'll push a commit adding it back in. Though in all fairness I'm wondering it's even necessary? If we can build CS on Windows fine without `___declspec(noreturn)` then maybe it's not necessary? 

One difficulty I noticed is that it appears that `__declspec` has to be put at the front of the function declarations while `__attribute__` needs to be put at the end, I'm not entirely sure how to resolve that using a macro but I'll do some research.

pmatos: You can add it at the front. It will work for both GCC and MSVC. @mflatt did that already in racket/ChezScheme. I always think that it's a good idea to provide compiler with our intentions (as developers). Even if it compiles without the `__declspec`, I would prefer if it's provided.

mflatt:
A clarification on racket/racket#2704:

The issue references an experimental commit that used `gcc -nostdlib -Xlinker ...` to access the linker via the C compiler and pass it linker options. For Racket purposes, I abandoned that approach in favor of using `ar` to build an archive.

Building an archive has the advantage that works in a more typical way. It has the drawback that clients linking Chez Scheme from an archive also need to know to link in libz, liblz4, libcurses, etc.

For a unified approach, it might make sense for Chez Scheme's `configure` to have a flag that picks between building "kernel.o" or "libchezscheme.a".

We've also made changes to support OpenBSD and NetBSD and (earlier) to support cross-compilation of Windows executables using MinGW. I'm starting to think it makes sense to assemble an omnibus PR to cover all of those.

pmatos:
> The issue references an experimental commit that used `gcc -nostdlib -Xlinker ...` to access the linker via the C compiler and pass it linker options. For Racket purposes, I abandoned that approach in favor of using `ar` to build an archive.
> 

Need to check how you're using `ar`. Using `ar` directly breaks `gcc` in LTO mode which needs to use `gcc-ar` wrapper. One way to do this is manually doing `./configure AR=gcc-ar ...` of course. However I haven't tested yet the new commit and have no CI job running compilation in LTO mode yet.

c/externs.h
abmclin: I've moved to the front and confirm it compiles without issues on my Windows machine. Thanks for the suggestion!

abmclin:
> We've also made changes to support OpenBSD and NetBSD and (earlier) to support cross-compilation of Windows executables using MinGW. I'm starting to think it makes sense to assemble an omnibus PR to cover all of those.

I agree, it seems that there are quite a few more changes which coordinate which each other for rest of the BSD family so for efficiency it makes sense to collect all of that into one PR so ChezScheme maintainers can check all of them at once.

abmclin:
By the way, if this PR ends up being accepted, please give me an opportunity to squash and clean up the commits prior to merging so I can give proper credit to Paulo Matos and Matthew Flatt for their suggestions that I have implemented.

akeep:
Thanks for the updates, and the note on compiling with `ld`.  I aliased this to `/usr/local/bin/ld`, but I did still run into some trouble and ended up running the `ld` command line that failed by hand.  After that things seem to build.

I'm currently running the tests.  I know there are some known inconsistencies in the mats on FreeBSD, but it is good to check that nothing else has gone out of whack.

I'd be interested in seeing updated support for the the BSDs.  Last time I looked at getting things going on OpenBSD, I ran into a read-only executable memory issue with their security model.  It seemed like there were two ways to work around this, don't rely on writable executable memory, which for us would mean writing the results of compilation or garbage collection into writeable non-executable memory, then marking the code segments read-only and executable.  I think I also saw that there was a specially directory you could put your system in to avoid this security check, I'd be interested to see what solutions people have found.

mflatt:
The Racket branch of Chez Scheme builds on OpenBSD only on a filesystem mounted with the wxneeded flag. On NetBSD, the makefiles include `paxctl +m` steps.

With threads, I don't think implementing W^X will be as easy as toggling between W and X at the right places. A thread might be trying to execute code on a page that is concurrently written for new code or for a GC. Maybe each chunk of new code could be forced to a fresh page, and it helps that a GC requires a rendezvous of all active threads — but "active" is the catch. There might be a `__collect_safe` callable that an inactive thread triggers during a GC, and even though the callable stub will immediately block the formerly inactive thread until the GC completes, that's too late if the GC happens to be relocating pointers on the stub's page (which may be necessary even if the stub is on its very own page).

The two solutions I know are (1) map the same memory to two virtual-address ranges, one writable and one executable; and (2) catch SIGSEGV and toggle pages between W or X in the hope of making progress. Solution (1) is used in libffi, but it's a hack, and keeping separate write and execute pointers seems painful. Solution (2) is used in Racket for x86_64 on OpenBSD, where the `siginfo_t` simply and reliably identifies the relevant page, but it's definitely a hack.

I recommend just requiring `wxneeded` or `paxctl +m`.

abmclin:
I had indeed noticed the failing mats tests earlier, I was going to open a separate issue but might as well discuss now in this thread. I was able to trace the failures to FreeBSD libm `pow` implementing the Scheme `expt` procedure returning a number that's slightly off from the expected `1e-20` value when `(expt 10.0 -20)` is evaluated.

I found this issue ticket filed with the FreeBSD about `pow` https://bugs.freebsd.org/bugzilla/show_bug.cgi?id=237800

based on the discussion in the ticket and the classic Goldberg article https://docs.oracle.com/cd/E19957-01/806-3568/ncg_goldberg.html I've tentatively arrived at the conclusion that floating-point equality with a number returned from a generic floating-point routine is too much to expect. 

I evaluated `(expt 10.0 -20)` in various Schemes on Windows and Linux, some of them also returned slightly different results that were off. Guile on Linux returned a different result which surprised me since I was assuming Guile would be using the same C math library  but guess not.

I think a more reasonable position to adopt is to test whether the number is within 1 ulp of the expected value based on my understanding so I believe the tests should be changed to avoid using equality. I'm not a floating-point arithmetic expert though.

samth:
For reference (not saying that anyone should implement this here), w^x + compiling is implemented in most/all modern browsers on all platforms. Here's a description of the Firefox work: https://jandemooij.nl/blog/2015/12/29/wx-jit-code-enabled-in-firefox/ For Edge, this document about vulnerabilities in the system is the most comprehensive: https://github.com/googleprojectzero/p0tools/blob/master/JITServer/JIT-Server-whitepaper.pdf

burgerrg:
Did pull request #443 resolve this pull request?

abmclin:
Apologies for the late response, yes #443 has resolved this pull request so I am closing it.

PR #442, opened 2019-07-01
fix `open-string-input-port` on immutable strings

mflatt:
The string-length-in-chars calculation in `$make-textual-input-port` needed updating for immutable strings in case an immutable string is passed to `open-string-input-port`.

### DIFF
--- a/release_notes/release_notes.stex
+++ b/release_notes/release_notes.stex
@@ -1695,6 +1695,12 @@ in fasl files does not generally make sense.
 %-----------------------------------------------------------------------------
 \section{Bug Fixes}\label{section:bugfixes}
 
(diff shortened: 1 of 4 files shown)


release_notes/release_notes.stex
```diff
@@ -1695,6 +1695,12 @@ in fasl files does not generally make sense.
 %-----------------------------------------------------------------------------
 \section{Bug Fixes}\label{section:bugfixes}
 
+\subsection{String ports form immutable strings (9.5.4)}
```
burgerrg: form => from

mflatt: Fixed.

PR #443, opened 2019-07-02
BSD support, Windows cross-compile via MinGW, and more configuration

mflatt:
This patch bundles a set of corrections and improvements to support builds on different platforms and with different options. Most of the changes are to makefiles and scripts, but some C sources are changed in small ways to support various platforms and build modes. The changes are a cleaned-up combination of changes that needed to build Racket on Chez Scheme for various platforms and to integrate it with the Racket build process. This PR is includes the changes in #437.

 * Restores support for FreeBSB, OpenBSD, and NetBSD using stock UUID libraries each platform. Building on OpenBSD requires a filesystem mounted with wxneeded. Building on NetBSD runs `paxctl +m`.

 * Adds makefile support and instructions for cross-compiling the `scheme.exe` executable for Windows using MinGW.

 * Adds a `--libkernel` configuration option, which builds (and installs) "libkernel.a" instead of "kernel.o". Building an archive with `ar` instead of merging objects with `ld` is a more conventional process, avoids problems when `cc` and `ld` are not in sync (historically a problem on Solaris, and apparently still a common problem on FreeBSD), and supports linking with system-wide shared libraries for zlib and/or LZ4. I think `--libkernel` module would be better as the default, but object mode remains the default for compatibility with anything that uses the old build outputs.

 * Adds a `run` target to the main makefile so that it's easy to run without installing with `make run`.

 * Adds a `boot` target to make it easy to generate boot files for a target machine with `make boot XM=<machine>`, and adds information to `BUILDING` on how to build on Unix variants other than Linux and OS X.

 * Adds `--disable-curses` to support limited build environments where [n]curses development files are not installed.

 * Adds `c\build.bat` and information to `BUILDING` for non-development compilation on Windows using Visual Studio.


mflatt:
Thanks for the alert, @jessealama, but that build error is an unrelated change in Racket (which we'll fix over there).

abmclin:
@mflatt thank you for preparing the omnibus PR. 

Currently the new PR fails to build on FreeBSD using gcc 7 with `implicit-fallthrough` errors, there is still a need to add more `NORETURN`s and `\* Fall Through *\` comments to various places.

`S_error_reset`, `S_error_abort`, `S_noncontinuable_interrupt`, and `S_error` in `/c/externs.h` trigger implicit fallthrough errors and they are wrapped around `do_error` in `c/schsig.c` so they all need `NORETURN` to compile cleanly. For sake of completeness I also added `NORETURN` to `S_error{1|2|3}` functions since those functions are all variants of `S_error` accepting different numbers of arguments.

`c/stats/c` has a couple places where the fall through comment is needed to compile.

For your convenience, I have attached a a tar.gz archive of three patch files which apply the changes to the relevant files in this PR.

[patches.tar.gz](https://github.com/cisco/ChezScheme/files/3355331/patches.tar.gz)

BUILDING
```diff
@@ -194,15 +203,41 @@ The make file supports several targets:
 'make distclean'
   removes nanopass, Makefile, and all workareas.
 
+OTHER UNIX VARIANTS
+
+To build on Unix variants other than Linux and OS X, you will first
+need to build boot files on a Linux or OS X machine. On that machine,
+after builing Chez Scheme, create boot files for the target machine
```
abmclin: `builing` -> `building`

BUILDING
```diff
@@ -194,15 +203,41 @@ The make file supports several targets:
...
+
+Remember to use gmake if make is not GNU make. If you run into linker
+trouble, try running configure with --libkernel so that the build
+avoids running ld directly.
+
+On OpenBSD, Chez Scheme must be built and installed on a filesystem
+that is mounted with wxneeded.
+
+On NetNSD, note that the makefiles run "paxctl +m" to enable WX pages
+(i.e., pages that have both wirte and execute enabled).
```
abmclin: `wirte` -> `write`

BUILDING
```diff
@@ -267,3 +302,37 @@ the Cygwin values are incompatible with the Microsoft C Runtime
...
+
+WINDOWS EXECUTABLE VIA COMMAND PROMPT
+
+To build the Chez Scheme executable using Microsoft Visual Studio,
+first set up command-line tools. The c\vs.bat script can help if you
+supply amd64 for a 64-bit build or x86 for a 32-bit build:
+
+ c\vs.bat amd64
+
+The, sun c\build/bat with a machine name, either ta6nt (64-bit
```
abmclin: I think `The, sun` -> `Then, run`?

c/types.h
```diff
@@ -338,11 +338,24 @@ typedef struct {
 #define tc_mutex_release() {}
 #endif
 
+#ifdef __MINGW32__
+/* With MinGW on 64-bit Windows, setjmp/longjmp is not reliable. Using
+   __builtin_setjmp/__builtin_longjmp is reliable, but
+   __builtin_longjmp requires 1 as its second argument. So, allocate
+   room in the buffer for a return value. Using 16 bytes of extra
+   room should preserves any relavant alignment. */
```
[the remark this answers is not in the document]
mflatt: Ah, right. I was confusing standard C malloc with some other allocator whose alignment guarantee depended on the requested size's alignment. I'll fix this.

BUILDING
mflatt: Got it - thanks!

mflatt:
Patches applied and typos fixed - thanks!

mats/5_3.ms
```diff
@@ -2346,7 +2346,7 @@
     (error? (expt 0 -1))
     (error? (expt 0 +1i))
     (eqv? (expt 2+2i 4) -64)
-    (eqv? (expt 10.0 -20) 1e-20)
+    (fl~= (expt 10.0 -20) 1e-20)
```
burgerrg: I'm not comfortable removing this precise test from the operating systems that implement the pow function correctly. What do you think of using a case statement with (machine-type) to express the expected result and the broken result on the operating systems that don't get it quite right?

mats/misc.ms
```diff
@@ -3483,8 +3483,8 @@
 )
 
 (mat sci-bug
-  (equal? (expt 10.0 (- 21)) 1e-21)
-  (equal? (flexpt 10.0 (- 21.0)) 1e-21)
```
burgerrg: It seems that this check of pow ought to be precise on all operating systems. See my comment on mats/5_3.ms.

mats/5_3.ms
mflatt: Casing on the machine type seems strange to me. If it's a bug that `expt` loses precision, then it seems like running the test should report that.

Calling it a bug does make sense, but then I'm inclined to say that the problem is outside the scope of this PR and just revert the test changes.

burgerrg: I agree on both counts. It does look like a bug outside the scope of this pull request that we don't want to mask.

burgerrg:
Thanks for this set of updates! I combined the LOG updates into the merge commit.

PR #444, opened 2019-07-11
Try to eliminate dead stores

alshopov:
Sometimes this allows removal of conditional or docs

### DIFF
--- a/c/new-io.c
+++ b/c/new-io.c
@@ -842,8 +842,6 @@ ptr S_bytevector_compress(ptr dest_bv, iptr d_start, iptr d_count,
         if (!is_valid_lz4_length(s_count))
           return Sstring("source bytevector ~s is too large");
 
-        destLen = (int)d_count;
-
         destLen = LZ4_compress_default((char *)&BVIT(src_bv, s_start), (char *)&BVIT(dest_bv, d_start), (int)s_count, (int)d_count);
 
         if (destLen > 0)
--- a/LOG
+++ b/LOG
@@ -1385,3 +1385,5 @@
     s/update-revision, BUILDING
 - export `ee-backward-delete-sexp` binding in the expression-editor module.
     expeditor.ss
+- remove dead stores in files
+    c/compress-io.c, c/new-io.c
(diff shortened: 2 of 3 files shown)


alshopov:
See https://github.com/racket/ChezScheme/pull/9

c/new-io.c
```diff
@@ -383,9 +383,6 @@ ptr S_close_fd(ptr file, IBOOL gzflag) {
  /* refuse to close stdin, stdout, and stderr fds */
   if (!gzflag && fd <= 2) return Strue;
 
- /* file is not locked; do not reference after deactivating thread! */
```
burgerrg: Please leave the comment here, because it still applies.

c/new-io.c
```diff
@@ -424,8 +421,6 @@ ptr S_bytevector_read(ptr file, ptr bv, iptr start, iptr count, IBOOL gzflag) {
   INT fd = gzflag ? 0 : GET_FD(file);
   glzFile gzfile = gzflag ? GZXFILE_GZFILE(file) : NULL;
 
- /* file is not locked; do not reference after deactivating thread! */
```
burgerrg: Please leave the comment here, because it still applies.

alshopov: Comment restored

c/new-io.c
alshopov: Comment restored

alshopov:
> update LOG
Done

dybvig:
That assignment is there to prevent accidental reference.  I'd hate to remove it just because C compilers think it has no purpose.

alshopov:
@dybvig 
> prevent accidental reference

@burgerrg 
> It's easy to overlook the comment

Oh boy, thanx for that! I thought `(ptr)-1` was decrementing 1 from ptr while in fact it is a cast.
I've removed these. What is the reason to use `-1` rather than `NULL`?

burgerrg:
0 is a valid file descriptor (stdin), but -1 isn't. Moreover, (ptr)(-1) is an invalid pointer so that if one tries to dereference it, an exception will be raised.

PR #445, opened 2019-07-12
Use setenv rather than putenv on non WIN32 environments

alshopov:

### DIFF
--- a/c/prim5.c
+++ b/c/prim5.c
@@ -1372,11 +1372,7 @@ static void s_putenv(name, value) char *name, *value; {
   if (rc == 0)
     S_error1("putenv", "environment extension failed: ~a", S_LastErrorString());
 #else /* WIN32 */
-  iptr n; char *s;
-  n = strlen(name) + strlen(value) + 2;
-  if ((s = malloc(n)) == (char *)NULL
-       || snprintf(s, n, "%s=%s", name, value) < 0
-       || putenv(s) != 0) {
+  if (setenv(name, value, 1) != 0) {
     ptr msg = S_strerror(errno);
 
     if (msg != Sfalse)
--- a/csug/system.stex
+++ b/csug/system.stex
@@ -5095,8 +5095,6 @@ is associated with \var{key}.
 environment of the process,
 where it is available to the current process (e.g., via \var{getenv})
 and any spawned processes.
-The key and value are copied into storage allocated outside of
-the Scheme heap; this space is never reclaimed on non-Windows systems.
 
 \schemedisplay
 (putenv "SCHEME" "rocks!")
--- a/release_notes/release_notes.stex
+++ b/release_notes/release_notes.stex
@@ -1695,6 +1695,12 @@ in fasl files does not generally make sense.
 %-----------------------------------------------------------------------------
 \section{Bug Fixes}\label{section:bugfixes}
 
+\subsection{\protect\code{putenv} memory leak (9.5.3)}
+
+\scheme{putenv} now calls the host system's \scheme{setenv} instead of
+\scheme{putenv} on non-Windows hosts and avoids allocating memory that
+is never freed, although \scheme{setenv} might do so.
+
 \subsection{String ports from immutable strings (9.5.4)}
 
 A bug that miscalculated the buffer size for
--- a/LOG
+++ b/LOG
@@ -1374,3 +1374,4 @@
     BUILDING, c/vs.bat, wininstall/locate-vcredist.bat
 - fixed open-string-input-port on immutable strings
     cpnanopass.ss, io.ms, release_notes.stex
+- use setenv rather than putenv on non WIN32 environments


alshopov:
See https://github.com/racket/ChezScheme/pull/10

burgerrg:
The [manual](https://cisco.github.io/ChezScheme/csug9.5/system.html#./system:s241) states that this leak is intentional. The BSD putenv manual page says that the memory becomes part of the environment, so it's not safe to free it. Perhaps we could use setenv instead?

alshopov:
Thank you very much for the review. I agree with you - it is not safe to free this.
What should I do? As you said we could move to using `setenv` as
`putenv("name=value")` is equivalent to `setenv("name", "value", 1)`
 * [*BSD](https://www.freebsd.org/cgi/man.cgi?query=getenv&apropos=0&sektion=0&manpath=FreeBSD+12.0-RELEASE+and+Ports&arch=default&format=html)
* [Linux setenv](http://man7.org/linux/man-pages/man3/setenv.3.html)
* [Posix setenv](http://man7.org/linux/man-pages/man3/setenv.3p.html)
* [glibc](https://www.gnu.org/software/libc/manual/html_node/Environment-Access.html#index-putenv)

all point to preferring `setenv` as safer - strings are not parsed and copies are used, rather than allowing direct change of the environment

Another way to resolve this is to just add single line comment in source explaining the rationale. Or it may turn out that this is common knowledge amongst system programmers.

burgerrg:
It would be a valuable change to use setenv. If you're willing to make the change, please update this pull request, and I'll review it. Don't forget to update LOG and remove the sentence about memory not being reclaimed. We should also update the release notes.

alshopov:
@burgerrg: can you review now?

c/prim5.c
```diff
@@ -1372,11 +1372,8 @@ static void s_putenv(name, value) char *name, *value; {
   if (rc == 0)
     S_error1("putenv", "environment extension failed: ~a", S_LastErrorString());
 #else /* WIN32 */
-  iptr n; char *s;
-  n = strlen(name) + strlen(value) + 2;
-  if ((s = malloc(n)) == (char *)NULL
-       || snprintf(s, n, "%s=%s", name, value) < 0
-       || putenv(s) != 0) {
+  ;
```
burgerrg: Why is there an empty statement here?

alshopov: fixed

alshopov:
Ready for a new review - LOG, manual, release notes updated, empty statement removed.

alshopov:
Added more unit tests as well

csug/system.stex
```diff
@@ -5095,8 +5095,12 @@ is associated with \var{key}.
...
 where it is available to the current process (e.g., via \var{getenv})
 and any spawned processes.
-The key and value are copied into storage allocated outside of
-the Scheme heap; this space is never reclaimed on non-Windows systems.
+There are no particular requirements for the scope or storage of
+arguments however \var{key} should not be empty, {\tt null} or contain
+the character {\tt =}.  \var{value} should not be {\tt null}.  The
+arguments are copied before being incorporated in the environment.
+Therefore their storate is outside of the Scheme heap and the space
+cannot be reclaimed on non-Windows systems.
```
burgerrg: I'd prefer to simply delete the sentence about storage allocation.

mats/misc.ms
```diff
@@ -1310,9 +1310,16 @@
   (equal? (getenv "FUBULY") "FRATZ")
   (eq? (putenv "FUBULY" "fratz") (void))
   (equal? (getenv "FUBULY") "fratz")
+  (eq? (putenv "FUBULY" "") (void))
+  (equal? (getenv "FUBULY") "")
   (error? (getenv 'hello))
   (error? (putenv 'hello "goodbye"))
   (error? (putenv "hello" 'goodbye))
+  (error? (putenv '() "goodbye"))
+  (error? (putenv "hello" '()))
```
burgerrg: Chez Scheme requires the arguments to putenv to be strings, so the two cases with () aren't necessary.

alshopov: Done

csug/system.stex
alshopov: Done

release_notes/release_notes.stex
```diff
@@ -58,6 +58,11 @@ Online versions of both books can be found at
 %-----------------------------------------------------------------------------
 \section{Functionality Changes}\label{section:functionality}
 
+\subsection{Process environment setting (non WIN32) (9.5.3)}
+
+\scheme{s_putenv} delegates to {\tt setenv} rather than {\tt putenv} on non WIN32 targets.
+It is potentially safer and is the recommended way to add to environment.
+
```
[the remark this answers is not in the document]
alshopov: done (almost verbatim)

csug/system.stex
burgerrg: The inserted lines 5098-5101 are still there.

release_notes/release_notes.stex
```diff
@@ -58,6 +58,12 @@ Online versions of both books can be found at
 %-----------------------------------------------------------------------------
 \section{Functionality Changes}\label{section:functionality}
 
+\subsection{Process environment setting (non WIN32) (9.5.3)}
+
+\scheme{putenv} now calls host's \scheme{setenv} instead of
```
burgerrg: calls host's => calls the host system's

release_notes/release_notes.stex
```diff
@@ -58,6 +58,12 @@ Online versions of both books can be found at
 %-----------------------------------------------------------------------------
 \section{Functionality Changes}\label{section:functionality}
 
+\subsection{Process environment setting (non WIN32) (9.5.3)}
```
burgerrg: I think this may better be characterized as a bug fix, because you didn't change what putenv does.
\subsection{\protect\code{putenv} memory leak (9.5.3)}

alshopov:
After latest code review:
 * There are no changes left in csug/system.stex
 *  release_notes/release_notes.stex: subsection moved to {Bug Fixes}, subsection title updated
   note that I put the subsection at top of section and since it is targeted at 9.5.3 it comes before String ports from immutable strings which is 9.5.4

burgerrg:
What happened to the deleted sentence in csug/system.stex?

alshopov:
> What happened to the deleted sentence in csug/system.stex?
I am not sure what I should do but I will delete the sentence:
```
The key and value are copied into storage allocated outside of
the Scheme heap; this space is never reclaimed on non-Windows systems.
```
If this is not an the correct thing - please state it again as I am getting lost in the versions of the patch.

release_notes/release_notes.stex
```diff
@@ -1695,6 +1695,12 @@ in fasl files does not generally make sense.
 %-----------------------------------------------------------------------------
 \section{Bug Fixes}\label{section:bugfixes}
 
+\subsection{\protect\code{putenv} memory leak (9.5.3)}
+
+\scheme{putenv} now calls host system's \scheme{setenv} instead of
```
burgerrg: "calls host" => "calls the host"

alshopov: as requested

burgerrg:
Thank you for this fix!

PR #446, opened 2019-07-15
Add additional check to prevent going before start of buffer

alshopov:
p is a pointer that iterates over path, which is buffer.
We should not try to get to an address preceding its start
Since there was an execution path that leads to that
guard against this with an additional check.


alshopov:
See https://github.com/racket/ChezScheme/pull/11

alshopov:
> update LOG.
Done, merge conflicts resolved

PR #455, opened 2019-08-22
Restore ability to build on Solaris-like systems

richlowe:
- Need `-luuid` for `uuid_generate`
- Use the native `ld(1)` instead of GNU.  If this is undesirable, the machine specification needs to be changed to `elf_i386_sol2` and `elf_x86_64_sol2`, which I can do instead
- Use the libc global `timezone` and `altzone` to make up for the lack of `tm_gmtoff`.

Even after this, there are two classes of problems in the tests, one, from the summary file, is

```
> io.mo:Error in mat iconv-codec clause 6: "transcoded-port: unsupported encoding GB18030" at line 688, char 7 of io.ms
> io.mo:Error in mat iconv-codec clause 14: "put-string: iconv UTF-8 codec cannot encode #\xFFFE" at line 979, char 7 of io.ms
> io.mo:Bug in mat iconv-codec clause 10 at line 788, char 7 of io.ms
> io.mo:Bug in mat iconv-codec clause 11 at line 813, char 7 of io.ms
> io.mo:Bug in mat iconv-codec clause 12 at line 824, char 7 of io.ms
6607,6608d6611
< io.mo:Expected error in mat iconv-codec: "close-port: iconv CP1252 codec cannot encode #\x3BB".
< io.mo:Expected error in mat iconv-codec: "close-port: iconv CP1252 codec cannot encode #\newline with eol-style ls".
```

This seems entirely due to the different behaviour of the iconv implementation on Solaris-like systems.

The other problem is a deadlock running the tests with `eval=interpret`, where both parent and child process are waiting on a `read()` from a pipe between them.  This seems likely to be the classic deadlock brought on by the unusually (for the times) small buffer used for `pipe(2)` on Solaris-like systems, but I'm not sure how to break into either scheme process to get the scheme stack at the time to narrow down exactly where it is.

### DIFF
--- a/LOG
+++ b/LOG
@@ -1393,3 +1393,4 @@
     compress-io.c, new-io.c
 - fixed tab character in makefiles
     c/Mf-*nt
+- restore building on Solaris
--- a/c/Mf-a6s2
+++ b/c/Mf-a6s2
@@ -16,7 +16,7 @@
 m = a6s2
 Cpu = X86_64
 
-mdclib = -lnsl -ldl -lm ${cursesLib} -lrt
+mdclib = -lnsl -ldl -lm ${cursesLib} -lrt -luuid
 C = ${CC} ${CPPFLAGS} -m64 -Wpointer-arith -Wextra -Werror -O ${CFLAGS}
 o = o
 mdsrc = i3le.c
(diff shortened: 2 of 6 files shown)


mflatt:
Solaris-like systems should build correctly after the v9.9.9 merge 47daa9b34016de84fd111801d9d589d15a523abe, but it would be good to confirm.

PR #464, opened 2019-09-18
Initialize n to 0

pmatos:
gcc 9.1.0 fails to compile with -Werror and -O3 because it detects that `n` might be used uninitialize in line
`if (n != scheme_version) {`

We do know that `n` will be initialized in `if (zget_uptr(file, &n) != 0) {` but gcc doesn't know that unless compiled with LTO.

### DIFF
--- a/LOG
+++ b/LOG
@@ -1759,3 +1759,5 @@
 - used with-object-file to restore accidentally dropped close-port in
   compile-whole-program and compile-whole-library
     compile.ss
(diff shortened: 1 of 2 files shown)


pmatos:
ping?

dybvig:
Looks like a necessary change.  Please add a short entry to LOG.

pmatos:
@dybvig thanks. Added LOG, rebased with master and force-pushed to branch.

dybvig:
Thanks for the fix!

PR #465, opened 2019-09-25
close port in compile-whole-program and compile-whole-library

owaddell:
The `(close-port op)` in the `finish-compile` called by `compile-whole-program` and friends appears to have been dropped during the recent reorganization.

I don't have a simple test program, but changing this fixed some issues for me in the Swish mats (some are still broken since we relied on concatenating .so files).

Edit: updated branch to use the new-fangled with-object-file. Seemed to work for me.

### DIFF
--- a/LOG
+++ b/LOG
@@ -1756,3 +1756,6 @@
   argument-type-check error messages, for Windows multibyte->string
   and string->multibyte.
     primvars.ms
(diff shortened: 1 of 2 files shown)


dybvig:
Looks like I dropped that with my refactoring.  The fix looks good and appropriately fangled.

PR #472, opened 2019-10-18
latest subprojects

zaoqi:


burgerrg:
Please use the latest released lz4 (see https://github.com/lz4/lz4/releases).

dybvig:
Thanks @zaoqi for this pull request.  I have just committed updated stex (1.2.2) and lz4 (1.9.2) subprojects.  I'm awaiting word from @akeep about a suitable released nanopass version to adopt, if any, and will update the nanopass subproject accordingly.  Since those updates address the intent of this pull request, I'm going to go ahead and close it.

For future reference, updating a subproject involves updating the subproject itself to the tag of a released version and also updating the relevant curl command in [configure](https://github.com/cisco/ChezScheme/blob/master/configure) to pull down the corresponding release.

Issue #484, opened 2019-12-28
read-char issues with unicode standard input on Windows 10


NHALX:
read-char seems to not be able to process unicode characters from text mode standard input ports on Windows 10 Pro 1909.

I've tried every combination of transcoder and buffer-mode options and it does not produce the correct result, nor does it produce any errors. Standard output works perfectly fine with UTF-8, and disk files do as well for both input and output. To test the terminal I checked that python3 deals with unicode input properly and it does.

The port is being converted like so:
(standard-input-port
         (buffer-mode block)
         (make-transcoder (utf-8-codec)))

The result is either #\nul or #\newline depending on which options/codecs are used. Various Greek letters were presented as input.

burgerrg:
See issue https://github.com/cisco/ChezScheme/issues/466.  The Windows console does not support UTF-8. You need to use a Unicode terminal such as Cygwin's Terminal or Git Bash.

NHALX:
@burgerrg recent builds seem to support it by setting the code page to 65001 (utf-8)

The default shell reads and writes unicode characters fine (tested with: mkdir Δ). Python3, when run in the terminal, does as well; as does Haskell's GHCI. Chez prints unicode ok but can't read.

Chez also has the same problem when run via an Emacs inferior mode on Windows. 

Most likely the windows APIs used by Chez don't interact well with the new UTF-8 codepage, but it should be possible to support unicode on Windows 10 without third party shells (which probably don't solve the problem with Emacs).

burgerrg:
Thanks, I see that Windows 10 now supports code page 65001 (utf-8).

First, the expression editor uses the ANSI API, so one must disable it with the --eedisable command-line switch.

Second, Chez Scheme calls the _read function in the Microsoft C Runtime library on file descriptor 0, which seems to be in a mode that does not handle UTF-8. Perhaps a _setmode call with _O_U8TEXT could fix this, but then the rest of the code would need to handle UTF-16LE if I understand the MSDN documentation correctly.

If you're willing to figure out how to get Unicode console input and output to work in Windows, I would be happy to review a pull request.

NHALX:
@burgerrg 

read-char produces the correct results when using GetStdHandle + ReadConsoleW + WideCharToMultiByte instead of _read at new_io.c:440. 

```
Chez Scheme Version 9.5.3
Copyright 1984-2019 Cisco Systems, Inc.

> (read-char)
Δ
#\Δ
```

However there are issues with this: 

[...]

PR #489, opened 2020-01-09
added unicode support to windows console i/o

NHALX:
Fixes issue #484. 

Please note the comment regarding SetConsoleCtrlHandler in os_read.


burgerrg:
Thank you for figuring out how to get Unicode BMP console I/O in Windows! I read on one of Microsoft's blogs that the console currently supports 16-bit Unicode characters only (no surrogates) but that they plan to add support for this in a future update. In the meantime, what we have seems to work.

burgerrg:
Please look at my proposed simplifications in https://github.com/cisco/ChezScheme/tree/unicode. Sometimes when I pressed CTRL-C at the repl with --eedisable, it would print two prompts as though CTRL-C were pressed twice. I think this is acceptable.

NHALX:
@burgerrg on a threaded build, when --eedisable is enabled, CTRL-C hangs with your changes. I had the same problem in one of the previous versions of my patch, and there is a comment about it in the os_read function. I'm not sure what was causing it, but after some refactoring it went away.

burgerrg:
Thank you, I'll investigate why refactoring changes the CTRL-C behavior. There must be a race somewhere.

burgerrg:
The MSDN description of [HandlerRoutine](https://docs.microsoft.com/en-us/windows/console/handlerroutine) explains the race. When CTRL-C is pressed, the handler routine is run in a new thread. I think this causes a bad interaction with our handler and the ReadConsoleW call in new-io.c. I'm investigating it.

burgerrg:
I found that by sleeping for 1 ms when ReadConsoleW returns EINTR before re-enabling the ConsoleCtrlHandler, we don't get the spurious double CTRL-C.

Then I found that threaded Chez Scheme does not handle console input and keyboard interrupts on any platform because the thread-safe handler uses the with-tc-mutex macro that disables interrupts, which makes the call to $event in io.ss not process the keyboard interrupt. On Windows, I handle this case by calling S_noncontinuable_interrupt, which is what happens on other platforms when you press CTRL-C the second time.

See if this works for you.

burgerrg:
I entered issue https://github.com/cisco/ChezScheme/issues/492 for the CTRL-C trouble on threaded Unix-like versions.

burgerrg:
@NHALX I got CTRL-C working reliably in threaded and non-threaded Windows builds. See what you think of the [changes](https://github.com/cisco/ChezScheme/compare/master...unicode).

NHALX:
@Burgerrg would it be better to use SetConsoleMode and toggle the ENABLE_PROCESSED_INPUT flag instead of dealing with signal handler masks the current way? Presumably this would bypass any races, which would be preferable to trying to win/lose them with Sleep.

My workstation is being repaired so I will need to wait a few days to test anything.

burgerrg:
No, we want processed input so that the line is editable. I believe I'm using the APIs correctly and there shouldn't be a race, but apparently there is. I'm OK with a small sleep in the CTRL-C case only. We do similar things to deal with Windows' DeleteFile returning before the file is actually deleted.

NHALX:
@burgerrg Sorry for the delay. The code seems to work fine here.

burgerrg:
The Windows Travis-CI build was failing because the building of mat output left it waiting for console input. I finally reproduced the behavior in my Windows VM and am now trying to debug it.

burgerrg:
Thanks again for your work on this!

PR #490, opened 2020-01-20
fix missing continuation call in `cp-push-mvrs`

mflatt:
In `np-push-mrvs`'s `Mvcall`, the continuation `k` wasn't used in the case for alloc, inline, and foreign-call expressions, so the result of a `call-with-values` form could get lost in a non-tail position. In the example below, the result of `(thing-pos posx)` gets lost.

```
;; cp0 is needed to expose `record-ref`:
(run-cp0 (lambda (cp0 x)
           (let ([r (cp0 x)])
             r)))

(define miscompiled
  (lambda (bx)
    (define-record-type thing
      (fields pos)
      (nongenerative #{thing hlg584lmg5htbdauw7dkid2sh-0}))
    (set-box! bx (make-thing 'x))
    (let ([posx (unbox bx)])
      (cons 'result
            (call-with-values
                (lambda ()
                  (if (thing? posx)
                      ;; compiled as inline load:
                      (thing-pos posx)
                      (do-something-else)))
              list)))))

;; Should print`(result x)`
(printf "~s\n" (miscompiled (box #f)))
```

### DIFF
--- a/LOG
+++ b/LOG
@@ -1769,3 +1769,7 @@
   is enabled, so local transformer code can be profiled.
     syntax.ss,
     profile.ms
+- fix compiler bug related to call-with-values and a first argument
+  whose body result is compiled to an allocation, inline form, or
+  foreign call
+    cpnanopass.ss, 3.ms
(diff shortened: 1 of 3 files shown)


dybvig:
Thanks for identifying the problem and for the fix!

PR #494, opened 2020-02-18
Add concatenate-object-files procedure.

jltaylor-us:
In previous versions of Chez Scheme, multiple object files could be
combined by concatinating them into a single file.  To support faster
object file loading and loadability verification, recompile information
and information about libraries and top-level programs within an object
file is now placed at the top of the file.  The new
concatenate-object-files procedure can be used to combine multiple object
files while moving this information to the top of the combined file.

### DIFF
--- a/LOG
+++ b/LOG
@@ -1858,3 +1858,7 @@
   'visiting' or 'revisiting' in a couple of places.
     syntax.ss,
     7.ms, 8.ms
+- added concatenate-object-files
+    compile.ss, primdata.ss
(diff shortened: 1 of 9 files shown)


dybvig:
This looks good to me, and  the changes are already being used internally at Cisco.   The travis-ci Windows builds are failing, but it looks like they started failing with [this commit](https://github.com/cisco/ChezScheme/commit/eb02b3ca882a11c5854658e2c854a0064bb04272).

PR #496, opened 2020-02-23
newrelease fixes and improved csug build reproducibility

weinholt:
The copyright years in csug/ have not been updated by newrelease due to slightly incorrect substitution patterns, which are fixed in this PR. The csug was also embedding the current date, which meant that the output would be different depending on the current month. Fixing this is a step towards reproducible builds.


dybvig:
Thanks for the newrelease fix.  I'm okay with the fixed revision date in csug.stex as well.  We'll just have to remember to update the date whenever we change the user's guide.

PR #500, opened 2020-02-29
Fasl compression

dybvig:
This pull request changes the way fasl files (hence object files) are compressed.  Instead of the file being compressed (or not) as a whole, each fasl object is compressed separately.  This allows the fasl loader to seek past (rather than reading each byte of) compile-time code and data at run time and run-time code and data at compile time.  This speeds load time (in my experiments, 9-10%) for libraries and top-level programs and increases fasl writing time (in my experiments, 2-3%).  In my experiments, it actually decreased object-file sizes slightly for libraries and top-level programs and for scheme.boot, but it significantly increased the size of petite.boot, since it consists of many small fasl objects so the compressor has less to work with at one time.   As part of the pull request, I wrapped a lot of the 's' directory files in begin forms to offset this and managed to reduce the hit from 27% increase to around a 11% increase in petite.boot size.


burgerrg:
The Windows build failures seem to be caused by the make line in .travis/test.sh taking over 10 minutes to produce output. For some reason, the Windows build system in Travis doesn't see the incremental output as make runs but sees it only at the end. I think we will need to break this command into shorter-running ones to get the build to work in Windows under Travis.

melted:
I have used travis_wait in those situations https://docs.travis-ci.com/user/common-build-problems/#build-times-out-because-no-output-was-received

burgerrg:
@melted, thank you for the suggestion. I see that travis_wait buffers the entire output of its subcommand before sending any of it. That caused me to experiment with the .travis/test.sh script. I found that grep on Windows does not flush its output. By using a bash script to filter the matting lines, I can avoid the need for travis_wait and get incremental output.

burgerrg:
The Travis-CI Windows build hung because the Git for Windows shell processor didn't process the arguments to cmd.exe properly. Now it's fixed on the master branch.

burgerrg:
@dybvig, is there anything else that needs to be done before merging to master?

dybvig:
No, so I just did the merge.  Thanks for reviewing and for fixing the travis-ci setup for Windows!

PR #502, opened 2020-03-01
Fix fasl-read signature

gus-massa:
The result of `fasl-read` may be `#f`.

Example from @mflatt :

    (let-values ([(o get) (open-bytevector-output-port)])
      (fasl-write #f o)
      (if (fasl-read (open-bytevector-input-port (get)))
          'wrong
          'correct))


dybvig:
True enough.  Would you please add a LOG entry?

gus-massa:
Done.

dybvig:
Thanks!

PR #506, opened 2020-04-02
 Added support for building chez with VS2019 Preview and VS2017 Preview

damon-kwok:
Added support for building chez with VS2019 Preview and VS2017 Preview


burgerrg:
The [VS2017 release notes](https://docs.microsoft.com/en-us/visualstudio/releasenotes/vs2017-preview-relnotes) have a warning:

> Preview releases are not "go-live" and not intended for use on production computers or for creating production code

As a result, I don't think it makes sense to put support for preview releases into the master branch. You might keep this as a gist or note for others to see how to add a different version of Visual Studio.

PR #508, opened 2020-04-06
Fix makefile for MSYS2

melted:
When MSYS2 bash sees the command "cmd.exe /c make.bat" it will
silently convert the /c into c:\, as it translates unixy paths
to Windows paths. This can be disabled by setting the environment
variable MSYS2_ARG_CONV_EXCL to *. So do that for this particular
command. It will be harmless in other environments like Cygwin.

### DIFF
--- a/LOG
+++ b/LOG
@@ -1997,3 +1997,6 @@
 - fixed csug copyright year substititions and changed revisiondate
   to not be generated, making the csug build reproducible
     newrelease csug/csug.stex
+- fixed Windows build using MSYS2
+    c/Mf-a6nt, c/Mf-i3nt, c/Mf-ta6nt, c/Mf-ti3nt, mats/Mf-a6nt,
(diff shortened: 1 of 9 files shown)


burgerrg:
We have `export MSYS_NO_PATHCONV=1` in the wininstall Makefile and mats Windows makefiles. Should we do the same thing for these c makefiles?

melted:
I tested it and that doesn't work on MSYS2 (but I saw that it works fine on the MSYS that comes with git, so that one is still useful). I have updated the patch to use exports as the previous one, I have added it to the test makefiles as well and added the `MSYS_NO_PATHCONV` to the ones in `c/`.

burgerrg:
Thanks for figuring this out! Please update LOG, and this pull request will be ready to merge into master.

melted:
Done!

burgerrg:
Thanks again!

PR #509, opened 2020-04-09
Fix build on Linux with musl libc

melted:
Distributions that use musl libc, such as Alpine, doesn't define
`__GLIBC__` yet still doesn't have xlocale.h.


melted:
Alpine is quite popular for making small Docker images.

burgerrg:
Thanks for the update!

amirouche:
To compile chez scheme on alpine linux, it is enough to do the following:

```sh
# apk add --update alpine-sdk util-linux-dev 
```

Issue #438, opened 2019-06-18
Compile/build problem on RasPi3


KenDickey:
Greetings,

First, thanks much for making Chez Scheme open source!!

I have been trying to get Chez up on a Raspberry Pi 3 under Raspian Linux, roughly following the guide at:

https://programmingpraxis.com/2017/09/15/compile-chez-scheme-on-android-arm/

The basic REPL is fine for small things, but some library loads fail with 
````
  Exception in +: #f is not a number
````

The test suite also shows this behavior.

Perhaps a codegen in FFI call?

Any suggestions?
Thanks much,
-KenD

[akku-compile-bug.txt](https://github.com/cisco/ChezScheme/files/3301985/akku-compile-bug.txt)
[mats-summary.txt](https://github.com/cisco/ChezScheme/files/3301987/mats-summary.txt)

**Comment by akeep** [comment]
It is very difficult to tell what is going on, except to say that the compiler is likely generating some bad code or the garbage collector is getting some corrupted data from the generated code which is causing it to overwrite something.

One thing to try is to compile the compiler and built-in libraries at optimize level 2:

```
$ make o=2 allx
```

If there is something wrong with the compiler backend for this target, it is likely you'll see some exceptions pop up during the compilation of the compiler and library code.

Debugging this kind of thing usually means trying to minimize to as small an example as possible and then running Chez Scheme under `gdb` (or `lldb` on macOS, though I've not used `lldb` much).

I do not have a Raspberry Pi 3 to test on, and currently the only ARM-based machine I have is not setup for this kind of development, but if I get some time I'll see if I can get an emulator running that I can try to replicate this in.  The failing `mats` are pretty indicative that things are going wrong.

I'm assuming this is the 32-bit ARM target that you are using on the Raspberry Pi 3?

KenDickey:
On 2019-06-25 20:16, Andy Keep wrote:
...
> One thing to try is to compile the compiler and built-in libraries at 
> optimize level 2:
> 
> $ make o=2 allx
> 
> If there is something wrong with the compiler backend for this target, 
> it is likely you'll see some exceptions pop up during the compilation 
> of the compiler and library code.

Great.  Thanks much.  I'll hook up my RasPi3 and check it out.

> I'm assuming this is the 32-bit ARM target that you are using on the 
> Raspberry Pi 3?

Yes.  Raspian Linux 32 bit.  I don't use it much.  My daily use machines 
are arm64 (LePotato and Chromebook Plus).

[...]

PR #510, opened 2020-04-27
Fix arm32le

cjfrisz:
This pull request attempts to fix `arm32` support. I'm not intending for it to be merged until all the mats are passing for `arm32le`, even though it should be in a mergeable state as is.

Currently, it fixes the original issue reported in #438 that was due to a bug in the handling of immediate operands.

Remaining are issues in three of the mats:

1. misc.ms: Errors in mat $read-time-stamp-counter
This is due to the time stamp counter on ARM having protected access that's generally enabled in user land via a kernel module. Whether the time stamp counter is accessible should be controlled by a parameter or a feature. I'm leaning toward a parameter, because it seems like it should be possible to load a kernel module that provides access to the time stamp counter after installing Chez Scheme and not need to rebuild the compiler. I can see an argument for making it a compiler feature, though, and input is welcome.

2. foreign.ms: Errors in mat structs
I don't have an answer to this one yet, but I've been investigating. Each clause of this mat performs a number of tests, but it seems to be the final test that's causing issues. Essentially, Scheme hands a foreign-callable to C (created via `make-ftype-pointer`) that C invokes to call `ftype-set!` on a memory location initialized on the C side. I provided a fairly minimal test case here: https://github.com/cisco/ChezScheme/issues/438#issuecomment-616293289.
Having done some time in GDB, it looks like everything is working through the invocation of the Scheme callback in C; the address points to code that matches the prelude generated for a foreign-callable (i.e., saving C state and setting up Scheme state). The problem occurs when jumping to the body of the foreign-callable code via`S_generic_invoke`. Mostly, this results in an "invalid memory reference error" where PC is set to a pretty high memory address. I increased the stack size provided by the operating system, and the error turned into "illegal instruction". Examining the PC when that occurs, its value is `0xffffffff` in my test case, which indeed is not a valid instruction. The surrounding memory values can be interpreted as valid instructions by GDB, but they look like garbage that's not meant to be treated as code.
I'm not stuck, but progress is slow as I slowly page effective ways to debug with GDB back into main memory.

3. foreign.ms: Bugs in mat varargs
I've been focused on errors in mat structs enough that I haven't seriously looked at this problem at all.


cjfrisz:
I also tried to replace the tabs in `arm32.ss` with spaces, since the code for `asm-foreign-call` and `asm-foreign-callable` were really hard to read. Sorry that inflates the diff.

It looks like some of the other backend files like `x86.ss` have the same problem.

s/arm32.ss
```diff
@@ -219,20 +219,20 @@
...
-                      (let ([u (make-tmp 'u)])
-                        (seq
-                          (build-set! ,u (asm ,null-info ,(asm-add #f) ,x0 (immediate ,imm)))
-                          (return u x1 0))))]
+                    ; NB: might should safe-assert x1 is %zero
+                    (let ([u (make-tmp 'u)])
+                      (seq
+                       (build-set! ,u (asm ,null-info ,(asm-add #f) ,x0 (immediate ,imm)))
+                       (return u x1 0)))]
                    [(funky12 (- imm)) =>
```
weinholt, line 229: Shouldn't this `=>` be removed?

cjfrisz: Yes--good catch

cjfrisz:
Added changes from @mflatt based on #438 as well as #528. Also added a new mat to `misc.ms` based on the test program from @weinholt in #438. Disabled the `varargs` mat in `foreign.ms` and the `read-time-stamp-counter` mat in `misc.ms` for `arm32le`.

dybvig:
@cjfrisz, I have looked through the changes that affect other machine types, and I don't see any issues there.  Though I haven't studied the arm32le-specific changes as closely, it seems clear we are better off with the new code than the old code, so I'm going to go ahead and do the pull.  Thanks for your efforts!  Thanks also to @mflatt and to everyone else who contributed to this request.

Issue #513, opened 2020-05-14
Lazy streams: GC retains memory for too long due to premature promotion of objects to maxgen


minad:
While working with lazy streams I found that the Chez GC retains objects for too long leading to a temporary space leak. The memory is reclaimed after a while, but the memory usage is huge.
Expected behavior would be constant memory usage.

In this example, the streams are implemented as pairs with a delayed cdr. Since the head is strict, these are odd streams according to [Wadler et al.](https://www.researchgate.net/publication/2646969_How_to_Add_Laziness_to_a_Strict_Language_Without_Even_Being_Odd). In contrast, srfi-40, 41 and 45 discuss even streams where the head is also delayed. But it turns out even streams are prone to space leaks. 

The odd streams here should not lead to leaks since all the functions are tail recursive. In particular, the stream head `(count 0)` is not retained. There is no need to keep many elements of the stream alive.

~~~
(define (count n)
  (cons n (delay (count (add1 n)))))

(define (filter p s)
  (if (p (car s))
      (cons (car s) (delay (filter p (force (cdr s)))))
      (filter p (force (cdr s)))))

(define (print n s)
  (when (> n 0)
    (display (car s))
    (display "\n")
    (print (sub1 n) (force (cdr s)))))

(print 100 (filter (lambda (n) (> n 50000000)) (count 0)))
~~~

jltaylor-us:
Can't do a whole lot with that code without the definitions for `delay` and `force`.

jltaylor-us:
oops, premature comment... please ignore.

jltaylor-us:
This "temporary space leak" behavior is inherent in using a generational garbage collector with data structures like this.  If you set the `collect-trip-bytes` parameter lower (so the garbage collector runs more often) then you'll see a corresponding reduction in memory used by this example.  (Alternately, if you were doing some interesting work on each item this would also allocate memory and cause the garbage collector to run more often.)

minad:
@jltaylor-us Thank you for your response! Setting `collect-trip-bytes` to a lower value helps - the memory usage is constant as it should. It also works if I set `collect-maximum-generation` to a smaller value (effectively disabling the generational gc). But using a high `collect-trip-bytes` value combined with many generations leads to the memory explosion (hundreds of MiB resident memory, while `collect-trip-bytes` seems to be 8MiB by default). Could you please elaborate a bit as to why that happens? 

[...]

PR #518, opened 2020-06-23
Change collector to promote objects only one generation at a time

dybvig:
This pull request addresses issue #513.  See LOG for detailed changes.

It also changes the x86 assembler to maintain 16-byte stack alignment (if incoming frames are 16-byte aligned) because newer versions of gcc count on 16-bit alignment for some generated sse instructions.  This properly belongs in a different pull request, but I discovered the alignment problem while testing the gc changes and I also didn't want to have to push two sets of boot files.


dybvig:
@burgerrg, I added you as a reviewer since I'm hoping you'll try building this on Windows.

burgerrg:
The following compiler error is raised on Linux:

burgerrg:
The following compiler error is raised on Windows:
```
gcwrapper.c(789): error C2065: 'S_child_processes': undeclared identifier
gcwrapper.c(789): error C2109: subscript requires array or pointer type
```

dybvig:
Thanks!  The gcwrapper.c problem on Windows is because I forgot to wrap the assignment in an "#ifndef Windows".

The gc.c problem on Linux is concerning.  By my reading of the code, there's no path to the reference to __to_g that doesn't assign __to_g first.  If I'm correct, this means the C compiler is issuing a bogus warning, possibly because it is using something other than live analysis to identify uninitialized values, because it's live analysis is broken, or because the code produced prior to live analysis is broken.  The version of gcc (10.1.1) on my machine doesn't complain.

burgerrg:
The Travis CI service provides Ubuntu 18.04, which has gcc (Ubuntu 7.5.0-3ubuntu1~18.04) 7.5.0 installed.

burgerrg:
Initializing __to_g to 0 (although it shouldn't be necessary) resolves the error on Ubuntu 18.04, and the rest of the system compiles without warnings.

burgerrg:
The Windows build failed with `gcwrapper.obj : error LNK2019: unresolved external symbol S_gc_011 referenced in function S_do_gc`

burgerrg:
I'll update the c/Makefile.* files and commit to this pull request.

burgerrg:
The automated tests all pass now!

PR #520, opened 2020-06-28
Change includes <Win[bB]ase.h> => <winbase.h>

Jarhmander:
When handling paths, Windows is case-insensitive, but unices are case-sensitive.

This simple patch facilitates cross-compiling the Windows version on a Unix system, and doesn't affect at all the native, Windows compilation.

To be more explicit, on a Linux system with an appropriate cross-compiler installed,
```
    ./configure -m=ta6nt --threads --toolprefix=x86_64-w64-mingw32-
    ./workarea ta6nt
    make -C ta6nt/c -f Mf-ta6nt cross=t o=o
```
will fail with:
```
segment.c:92:10: fatal error: WinBase.h: No such file or directory
   92 | #include <WinBase.h>
      |          ^~~~~~~~~~~
compilation terminated.
```
(and another similar error in windows.c) unless this present patch is applied.

### DIFF
--- a/LOG
+++ b/LOG
@@ -2077,3 +2077,5 @@
     gc.c
 - updated Windows makefiles
     c/Makefile.*nt
+- use lowercase for Windows include files
+    segment.c, windows.c
--- a/c/segment.c
+++ b/c/segment.c
@@ -89,7 +89,7 @@ void S_freemem(void *addr, iptr bytes) {
 #endif
 
 #if defined(USE_VIRTUAL_ALLOC)
-#include <WinBase.h>
+#include <winbase.h>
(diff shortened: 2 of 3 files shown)


Jarhmander:
(note: last Travis CI build failed for ti3nt because of a timeout in tests... it certainly has nothing to do with this patch)

burgerrg:
The change seems fine to me. Please update the LOG file.

Jarhmander:
> The change seems fine to me. Please update the LOG file.

Fair enough, I'll also update the other pending PR.

burgerrg:
Thank you!

PR #521, opened 2020-06-28
Small refactoring and proper unicode handling of error messages.

Jarhmander:
On a Windows system, error messages from the OS were not unicode-aware, and messages were not properly read back.

Example transcript from a french OS:
```
> (load-shared-object "not a dll")
Exception: (while loading not a dll) Le module sp�cifi� est introuvable
```

This patch fixes this, and with a bonus refactoring in windows.c to avoid duplicating code in `S_ntdlerror` and `s_ErrorString`.

### DIFF
--- a/c/windows.c
+++ b/c/windows.c
@@ -55,18 +55,11 @@ void *S_ntdlsym(void *h, const char *s) {
 
 /* Initial version of S_ntdlerror courtesy of Bob Burger, burgerrg@sagian.com
  * Modifications by James-Adam Renquinha Henri, jarhmander@gmail.com */
-char *S_ntdlerror(void) {
-    static char *s = NULL;
-    /*
-     * The caller does not expect to have to free the memory returned by this
-     * function (because normally, you shouldn't free the result of dlerror).
-     * But to properly support Unicode on Windows, we have to allocate memory to
-     * hold a utf-8 string. Hence, we have this semi-leak situation, where we
-     * always hold the last error string, and free the previous one, each time
-     * we call this function.
-     */
+ptr S_ntdlerror(void) {
+    ptr ret;
+    char *s = s_ErrorStringImp(GetLastError(), "unable to load library");
+    ret = Sstring_utf8(s, -1);
     free(s);
-    s = s_ErrorStringImp(GetLastError(), "unable to load library");
     return s;
 }
 
--- a/c/windows.c
+++ b/c/windows.c
@@ -53,7 +53,7 @@ void *S_ntdlsym(void *h, const char *s) {
     return (void *)GetProcAddress(h, s);
 }
 
-/* Initial version of S_ntdlerror courtesy of Bob Burger, burgerrg@sagian.com
+/* Initial version of S_ntdlerror courtesy of Bob Burger
  * Modifications by James-Adam Renquinha Henri, jarhmander@gmail.com */
 ptr S_ntdlerror(void) {
     return s_ErrorStringImp(GetLastError(), "unable to load library");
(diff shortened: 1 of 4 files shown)


Jarhmander:
Updated the LOG file, rebased and force-pushed. I anticipated that it would cause unnecessary conflicts, hence the force-push.

Maybe the most controversial "feature" of this PR is that some memory is allocated each time `FormatMessage` is called to fetch error messages that are converted from Windows `wchar_t` to utf-8 strings, and `S_ntdlerror` always hold the last string thus generated, freeing it only when a new error string is generated, which is called a "semi-leak" in the code. A comment in the code explain this situation in detail. It could probably be better done, but only with more invasive modifications in c/foreign.c, and I honestly think this isn't an issue, as memory usage will not grow over time, and that code path is only taken for error situations when using the Windows API.

burgerrg:
The static variable isn't thread safe, so I'd prefer a change that has S_ntdlerror returning the Scheme string directly. Then update foreign.c so that it calls Sstring_utf8(dlerror(), -1) on Unix systems, and S_ntdlerror() on Windows.

burgerrg:
s_ErrorString has the proper signature, so it would be straightforward to change foreign.c to call it with GetLastError() on Windows and eliminate S_ntdlerror altogether.

Jarhmander:
Effectively, the new implementation wasn't thread safe (so was the old one, but it might be less critical of a problem here), so I changed it to allocate Scheme strings from the start.

As for the proposed implementation, I respectfully disagree. Better be consistent: all implementations "calls" `s_dlerror` which always return a `ptr`. `s_dlerror` is a macro expanding to `Sstring_utf8(strerror(errno), -1)`, `Sstring_utf8(dlerror(), -1)` or `S_ntdlerror` (which now returns a `ptr`), depending on the target. I decided to keep `S_ntdlerror` just because its behavior is different from `s_ErrorString` when `FormatMessageW` fails to find the proper message for the code returned by `GetLastError()`: the former will return `"unable to load library"`, the latter will instead show the error code in hexadecimal. That was the exact behavior before my proposed changes, and I prefer not to change that aspect.

c/windows.c
```diff
@@ -242,33 +240,46 @@ static IUnknown *s_CreateInstance(CLSID *pCLSID, IID *iid) {
...
-        snprintf(hexerrbuf, HEXERRBUFSIZ, "0x%x", dwMessageId);
-        return Sstring(hexerrbuf);
+            result = malloc(HEXERRBUFSIZ);
+            snprintf(result, HEXERRBUFSIZ, "0x%x", dwMessageId);
+            return result;
 #undef HEXERRBUFSIZ
+        }
     }
+    result = Swide_to_utf8(lpMsgBuf);
+    endstr = result + len;
```
burgerrg, line 272: len is the number of wide characters in lpMsgBuf, not the number of 8-bit characters in result.

What do you think of pruning the trailing wide characters from lpMsgBuf before converting to UTF-8? Then you can use len in that loop and simply return `Swide_to_utf8(lpMsgBuf)`?

c/windows.c
```diff
@@ -242,33 +240,46 @@ static IUnknown *s_CreateInstance(CLSID *pCLSID, IID *iid) {
...
     DWORD len;
-    ptr result;
+    char *result;
+    char *endstr;
 
-    len = FormatMessage(FORMAT_MESSAGE_ALLOCATE_BUFFER | FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_IGNORE_INSERTS,
-                        NULL, dwMessageId, MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT), (LPTSTR)&lpMsgBuf, 0, NULL);
-    /* If FormatMessage fails, use the error code in hexadecimal. */
+    len = FormatMessageW(FORMAT_MESSAGE_ALLOCATE_BUFFER | FORMAT_MESSAGE_FROM_SYSTEM | FORMAT_MESSAGE_IGNORE_INSERTS,
+                        NULL, dwMessageId, MAKELANGID(LANG_NEUTRAL, SUBLANG_DEFAULT), (LPWSTR)&lpMsgBuf, 0, NULL);
```
burgerrg, line 256: Please add another space to indent this line.

Jarhmander: Fixed.

c/windows.c
Jarhmander: I can't believe I missed that! Fixed it like suggested.

Jarhmander:
Applied suggested changes, and also a tiny but important change too, which was laying there waiting for me in my working copy...

c/windows.c
```diff
@@ -60,7 +60,7 @@ ptr S_ntdlerror(void) {
     char *s = s_ErrorStringImp(GetLastError(), "unable to load library");
     ret = Sstring_utf8(s, -1);
     free(s);
-    return s;
+    return ret;
```
burgerrg, line 63: Good catch!

c/windows.c
```diff
@@ -268,18 +268,18 @@ static char *s_ErrorStringImp(DWORD dwMessageId, const char *lpcDefault) {
...
-        char c = *--endstr;
-        if (c == '.') {
+    while (lpMsgBuf != endstr) {
+        wchar_t c = *--endstr;
+        if (c == L'.') {
             *endstr = 0;
             break;
         }
-        else if (c != '\n' && c != '\r') break;
+        else if (c != L'\n' && c != L'\r') break;
```
burgerrg, line 279: This loop doesn't handle the case where the string doesn't end in a period but has trailing newlines and returns. You might simply use the original loop but with wchar_t to find the new end and then set it to nul.

c/windows.c
```diff
@@ -52,17 +53,14 @@ void *S_ntdlsym(void *h, const char *s) {
...
-    if (n == 0) return "unable to load library";
-  /* Strip trailing period, newline & return when present */
-    if (n >= 3 && s[n-3] == '.') s[n-3] = 0;
-    return s;
+/* Initial version of S_ntdlerror courtesy of Bob Burger, burgerrg@sagian.com
+ * Modifications by James-Adam Renquinha Henri, jarhmander@gmail.com */
+ptr S_ntdlerror(void) {
+    ptr ret;
+    char *s = s_ErrorStringImp(GetLastError(), "unable to load library");
+    ret = Sstring_utf8(s, -1);
```
burgerrg, line 61: You might simply `ptr ret = ...`

Jarhmander:
> With these improvements, I think that both uses of s_ErrorStringImp take the string, import it into Scheme, and then free it. s_ErrorStringImp could do this to simplify the code further.

Good idea. The last commit takes that approach. In doing so, with very little code duplication I could get rid of some C memory allocation.

c/windows.c
Jarhmander: I thought that wasn't C89-compliant, but I guess I was wrong. In any case, this hunk is no longer relevant…

c/windows.c
Jarhmander: Well somehow this wasn't added to the commit. As it is, it prints fixnums.

c/windows.c
Jarhmander: The fact that I wanted to rewrite that part of this code is that it wasn't obvious to me what it was supposed to do. But that wasn't very professional of my part to not test against the old code. I did it this time and I have something that matches 100% what the old code did, except on uninteresting cases (lines solely consisting of newlines and returns).

c/windows.c
```diff
@@ -242,33 +236,44 @@ static IUnknown *s_CreateInstance(CLSID *pCLSID, IID *iid) {
...
+            /* ... use the default string if provided... */
+            return Sstring_utf8(lpcDefault, -1);
+        } else {
+            /* ...otherwise, use the error code in hexadecimal. */
+            char buf[(sizeof(dwMessageId) * 2) + 3] = {0};
+            snprintf(buf, sizeof buf, "0x%x", dwMessageId);
+            return Sstring_utf8(buf, sizeof buf - 1);
         }
-        else break;
+    } else {
```
Jarhmander, line 260: Added an `else` clause to further delimit the scope of some variables.

burgerrg:
Thanks for the updates. I pushed a commit that uses the return value of snprintf to pass to Sstring_utf8 and strips trailing newlines and a period using the original logic.

burgerrg:

PR #523, opened 2020-07-16
Uniform (load-shared-object #f) handling across platforms

Jarhmander:
This is a request to make an undocumented feature… documented, and work correctly.

`load-shared-object` is a foreign function, so it accepts `#f` as well as strings as an argument. On e.g. Linux, `(load-shared-object #f)` does something very useful: it essentially calls `dlopen(NULL, RTLD_NOW)`, which makes especially easy to get regular C functions like `memcpy` available as foreign entries, but also _any_ linked shared objects and even external symbols for the executable itself. This can be a very effective simplification if one intends to make a portable, single file binary from Chez Scheme.

On Windows, `(load-shared-object #f)`… crashes. Obviously, `Sutf8_to_wide` doesn't like getting a NULL pointer. Rather than simply fix this, this PR goes quite a bit further by attempting to emulate the useful behavior of `dlopen(NULL, ....)` from UNIX. Doing so is not so simple apparently, as there no direct equivalent, and the code ends up enumerating the loaded DLLs of the current process and add all of them to the `S_foreign_dynamic` list.

I'm aware that the approach taken here might be against the core principle of being efficient in one particular case for a particular platform, but it's for the benefit of usability.

Also, I'm aware that this PR is likely not acceptable, as there is no tests written to the added functionality. This is an area I need a bit of guidance, as I'm not sure what other files besides `mats/foreign.ms` I have to modify.

Finally, note that I put some unrelated changes to the documentation in a separate commit. Review these carefully.

### DIFF
--- a/c/foreign.c
+++ b/c/foreign.c
@@ -38,6 +38,7 @@
 #define dlopen(path,flags) (void *)shl_load(path, BIND_IMMEDIATE, 0L)
 #define s_dlerror() Sstring_utf8(strerror(errno), -1)
 #elif defined(WIN32)
+#include <psapi.h>
 #define dlopen(path,flags) S_ntdlopen(path)
 #define dlsym(h,s) S_ntdlsym(h,s)
 #define s_dlerror() S_ntdlerror()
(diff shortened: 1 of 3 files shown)


c/foreign.c
```diff
@@ -228,9 +229,63 @@ static void load_shared_object(path) const char *path; {
...
+        DWORD ret;
+        HMODULE modules_stack[1024];
+        DWORD cur_num_bytes = sizeof modules_stack;
+        DWORD req_num_bytes;
+        HMODULE *modules = modules_stack;
+        HMODULE *dynamic_storage = NULL;
+
+        /* Always try to use the stack, allocate memory only if necessary */
+        ret = EnumProcessModules(GetCurrentProcess(), modules, cur_num_bytes, &req_num_bytes);
+        while (ret && req_num_bytes > cur_num_bytes) {
```
Jarhmander, line 256: A loop is used there, rather than a simple `if`, to be resilient in the face of weird foreign threads that loads DLL concurrently and independently of the Scheme system. This loop _ought_ to terminate quite fast, otherwise that means something bad happens anyway because it would mean the list of loaded modules grows very fast, and that's in itself the biggest concern, _not that allocation loop_.

Jarhmander:
Will recreate one later with smaller and slightly different changes.

PR #526, opened 2020-08-14
repair ephemeron handling with incremental promotion in the collector

mflatt:
Recent changes to add incremental promotion of objects to the collector was slightly incomplete for ephemerons.


dybvig:
Thanks for the fix!

PR #530, opened 2020-08-31
continuation repair for return-arity error

mflatt:
When the wrong number of values are returned to a multiple-value context, the stack pointer is left as-is while calling the exception handler. The intent is to keep the current procedure's frame in place for debugging. However, there's no guarantee that the return address is still on the stack (much less in a return register, when there is one); for example, the continuation starting with the current frame may have been captured and just restored. If the return address is not in place, then the current continuation is ill-formed in the exception handler.

For example, this program tends to crash:

```
(define (go)
  (call-with-values get
    (lambda (a b) (list a b))))

;; returns just one value after capturing the continuation:
(define (get)
  (call/cc
   (lambda (k)
     (collect) ; do something non-trivial
     k)))

(with-exception-handler
 (lambda (exn)
   (call/cc
    (lambda (k)
      (collect) ; detect broken `k`
      k)))
 go)
```

This patch repairs the problem by always reinstalling the return address before jumping to the mvlet error function.

### DIFF
--- a/LOG
+++ b/LOG
@@ -2132,3 +2132,6 @@
     wininstall/ta6nt.wxs wininstall/ti3nt.wxs
 - newrelease no longer logs as updated files with no actual changes
     newrelease
+- repaired continuation for exception handler of error for reurning
+  the wrong number of values to a multiple-value context
+    cpnanopass.ss, np-languages.ss, 3.ms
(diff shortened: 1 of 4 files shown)


dybvig:
Looks good.  Thanks for identifying and fixing this!

PR #536, opened 2020-10-18
Update nanopass-framework to v1.9.2

akeep:
Update to the latest nanopass-framework to allow `define-parser` use in
Chez Scheme, which previously caused problems because the resulting
parser had run-time dependencies on nanopass libraries.


dybvig:
Looks good.  I'm curious how / whether this manifests in an error in Chez Scheme and whether we should introduce a mat to check for that.

akeep:
I think it is tricky to test this one out, though maybe if we write a little test library and then get the revisit requirements for the library and see that it does not list any of the nanopass framework requirements.

Without this change if you define a parser for one of our languages from `np-language.ss` and exported it as a primitive, for instance by adding `($parse-L4.875 [flags])` to `primdata.ss` and then added the following:

```
(set! $parse-L4.875
  (let ()
    (import (nanopass))
    (define-parser parse-L4.875 L4.875)
    parse-L4.875))
```

to `cpnanopass.ss`, you'd produce a boot file that had a dependency on the `(nanopass nano-syntax-dispatch)` and `(nanopass implementation-helpers)` libraries, and you'd see:

```
Exception: library (nanopass nano-syntax-dispatch) not found
```

when you try to start Chez Scheme with the resulting binary.

dybvig:
Thanks for the update!

Issue #539, opened 2020-11-16
Use GOTPCRELX-like optimization for x86-64 code objects


fweimer:
Currently, 

```c
static void x86_64_set_jump(void *address, uptr item, IBOOL callp) {
  I64 disp = (I64)item - ((I64)address + 5); /* 5 = size of call instruction */
  if ((I32)disp == disp) {
    *(octet *)address = callp ? 0xE8 : 0xE9;  /* call or jmp disp32 opcode */
    *(I32 *)((uptr)address + 1) = (I32)disp;
    *((octet *)address + 5) = 0x90; /* nop */
    *((octet *)address + 6) = 0x90; /* nop */
    *((octet *)address + 7) = 0x90; /* nop */
    *((octet *)address + 8) = 0x90; /* nop */
    *((octet *)address + 9) = 0x90; /* nop */
    *((octet *)address + 10) = 0x90; /* nop */
    *((octet *)address + 11) = 0x90; /* nop */
  } else {
    *(octet *)address = 0x48; /* REX w/REX.w set */
    *((octet *)address + 1)= 0xB8;  /* MOV imm64 to RAX */
    *(uptr *)((uptr)address + 2) = item;
    *((octet *)address + 10) = 0xFF;  /* call/jmp reg/mem opcode */
    *((octet *)address + 11) = callp ? 0xD0 : 0xE0; /* mod=11, ttt=010 (call) or 100 (jmp), r/m = 0 (RAX) */
  }
}
```

It is possible to use a six-byte CALL or JMP instruction sequence for both 32-bit and 64-bit addressing. If a 32-bit displacement is sufficient, `0x67 xe8` (CODE32 CALL rel32) or `0xe9` (JMP rel32; with a 1-byte NOP after the displacement) work. Otherwise, use `0xff 0x15` (CALL m64)  or `0xff 0x25` (JMP m64) with a RIP-relative memory operand. This needs a relocation table at the end of the code object, but I think it already exists today. Code objects will be restricted to 2 GiB in size, but that's a limitation that exists today, I assume.

This technique is described under  [Linker Optimization](https://gitlab.com/x86-psABIs/x86-64-ABI/-/blob/master/x86-64-ABI/linker-optimization.tex) in the x86-64 psABI supplement.

[...]

PR #540, opened 2020-11-16
x86-64 fasl relocator: Use single 7-byte NOP after direct calls/jumps

fweimer:
See issue #539 for background.

`a6le/mats/summary` is still clean after this change.


dybvig:
I'm curious how this affects performance.  I assume it should reduce overall instruction decoding time, though such assumptions don't always hold.  Have you verified that it at least doesn't make code run slower?  If so, it looks like a great change.  Please add a LOG entry.

fweimer:
Fair enough. What would be an acceptable benchmark? I usually measure instruction counts because they are more stable than timings, but that would be cheating in this case.

The 7-byte NOP sequence is recommended in both the AMD and Intel programming manuals. Perhaps this is good enough?

dybvig:
Okay, I just ran a set of benchmarks on my mac laptop.  The average difference wasn't significant, so let's go ahead.  I'll pull the change once you add the LOG entry.

fweimer:
Fair enough. I added:

```
- use a 7-byte NOP in the x86_64 fasl relocator instead of 7 1-byte NOPs
```

dybvig:
Please add the file modified (fasl.c) in the LOG entry following the format of the earlier entries.  It might seem unnecessary given that we have the git log, but Chez Scheme tends to outlast revision control systems.

dybvig:
Also you'll have to rebase (and squash, please) since I've just pulled another request that modified LOG.

PR #541, opened 2020-11-16
NOTICE: Mention LZ4

fweimer:
Commit 8858b34bd92ac8d2b6511dc9ca17ebfa06a1bd93 ("Add LZ4 support and
use it by default for compressing files") added LZ4 support and the
lz4 submodule, but did not update the NOTICE file, unlike zlib support
is handled.


dybvig:
Thanks for catching and fixing this.

PR #543, opened 2020-11-20
Add special case in cpnanopass for (eq? (ftype-pointer-address x) 0)

gus-massa:
Add a special case in cpnanopass.ss for `(eq? (ftype-pointer-address x) 0)`. 

`eqv?`, `equal?` already have the same special case.

### DIFF
--- a/LOG
+++ b/LOG
@@ -2140,3 +2140,5 @@
     arm32.ss
 - repair remainder and modulo on flonums by using fmod
     prim5.c, 5_3.ss, 5_3.ms
+- add special case in cpnanopass.ss for (eq? (ftype-pointer-address x) 0)
+    cpnanopass.ss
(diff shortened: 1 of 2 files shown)


dybvig:
Thanks!   I've gone ahead and merged this, but here are a few notes just for edification:

- eq? isn't defined for numbers, so technically this case should never be exercised by correct code.  In fact, however, in Chez Scheme eq? behaves like eqv? when one of the arguments is a fixnum, including 0.
- The change also turns (eq? (ftype-pointer-address x) (ftype-pointer-address x)) into #t, even though would otherwise return #f in Chez Scheme when (ftype-pointer-address x) is a bignum.  No harm done, since eq? isn't defined for numbers.
- The text of a LOG entry doesn't need to mention the modified filename(s) since they are listed explicitly.

PR #546, opened 2021-01-14
Limit the line length

soegaard:
According to Matthew Butterick in "Practical Typography" [1] the
maximal line length should not be much higher than 90 characters.

This commit changes style of csug style to set the maximal line length.
This should only affect users with wide screens.

[1] https://practicaltypography.com/line-length.html

### DIFF
--- a/csug/csug.css
+++ b/csug/csug.css
@@ -1,4 +1,5 @@
-BODY {background-color: #FFFFFF}
+BODY { background-color: #FFFFFF; max-width: 40rem; margin:auto }
+
 
 a:link, a:active, a:visited { color:#005568; text-decoration:underline }
 a:hover { color:white; text-decoration:underline; background:#005568 }


jltaylor-us:
If a wide-screen user wants a narrower text they can make their web browser window narrower.  With this change a user who wants text wider than 640 pixels (on my system, using the default font size) has no recourse aside from manually overriding the style sheet.

soegaard:
@jltaylor-us 

>  With this change a user who wants text wider than 640 pixels (on my system, using the default font size) has no recourse aside from manually overriding the style sheet.

That was not the intention. The intention was to limit the line length on very wide screens.

Clearly, I have missed something. Can you help me reproduce this:
What is your OS, browser and normal screen resolution?

jltaylor-us:
I tested on both Safari in the latest MacOS on a 5K monitor and Firefox and Safari on MacOS Mojave on a 15-inch MBP at its native resolution (2880x1800).  The screen resolution isn't particularly relevant, though, since my browser windows tend to be about 1000px wide.

My default font size is 16px (which the internet tells me is the default for most browsers), and 16px * 40 is 640px.

The specific value of the maximum width aside, this change is about enforcing that preference in the CSS rather than allowing the user to choose their own page width.  Generally, I'd expect that to be done by making the window size whatever they want, but if the user really wants a full-screen window with huge swaths of whitespace on the sides of the text then they can do that with their browser's reader view.

All of that said, I probably wouldn't have even noticed if the maximum width were set to something larger than my usual window size... but I'm not sure what the point of that would be, either.

soegaard:
Different habits :-)  I usually have my browser set to full screen - so adjusting the width is cumbersome.

The image shows what I see on a 4k screen in Chrome on macOS using a resolution of 1920x1080.

The number of characters on the widest line is around 100. 
```
> (string-length "indicate the locations in the text where forms and procedures are formally defined. The summary of")
98
```
To me the result in Safari looks identical.

The recommendation from Pratical Typography is a max of 90 so I was afraid of complaints that the line width was too wide.

image

akeep:
Thanks for the pull request, but I think we are going to keep things as they are for now.  The PDF version of the docs provide a fixed width way to view the code, as do the reader modes in most modern browsers.

Issue #281, opened 2018-03-26
build failing on NetBSD


jessup:
Following #7, I generated boot files for M=a6nb on W=i3le using the method that is now described in BUILDING:

(cd $W ; make -f Mf-boot $M.boot)

This seemed to go well, so I moved boot/a6nb to the BSD machine, then called configure with no flags.

Building stats.c fails in this situation. I have done some tinkering with a6nb/c/Makefile, to see if this is just a problem with my system libraries being in strange places, but to no avail.

uname -prs: NetBSD 8.0_BETA x86_64
Here's the output from gmake --debug:

GNU Make 4.1
Built for x86_64--netbsd
Copyright (C) 1988-2014 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Reading makefiles...
Updating goal targets....
 File 'build' does not exist.
Must remake target 'build'.
(cd a6nb && gmake build)
GNU Make 4.1
Built for x86_64--netbsd
Copyright (C) 1988-2014 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Reading makefiles...
Updating goal targets....
 File 'build' does not exist.
Must remake target 'build'.
(cd c ; gmake)
GNU Make 4.1
Built for x86_64--netbsd
Copyright (C) 1988-2014 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
This is free software: you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
Reading makefiles...
Updating goal targets....
 File 'doit' does not exist.
   File '../bin/a6nb/scheme' does not exist.
     File '../boot/a6nb/kernel.o' does not exist.
         Prerequisite '../zlib/configure.log' is newer than target '../zlib/zconf.h'.
         Prerequisite '../zlib/configure.log' is newer than target '../zlib/zlib.h'.
       File 'stats.o' does not exist.
      Must remake target 'stats.o'.
[...]

PR #552, opened 2021-03-03
Fix missing include in externs.h

ismaell:
The time.h header is required to make the timespec struct available for
the prototypes.

Fixes: https://github.com/cisco/ChezScheme/issues/281

### DIFF
--- a/LOG
+++ b/LOG
@@ -2142,3 +2142,5 @@
     prim5.c, 5_3.ss, 5_3.ms
 - add special case in cpnanopass.ss for (eq? (ftype-pointer-address x) 0)
     cpnanopass.ss
+- fix missing include in externs.h for struct timespec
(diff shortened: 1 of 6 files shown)


burgerrg (changes requested):
Thank you for this fix! Now that externs.h includes time.h, please remove the include of time.h from the C files that already include externs.h. Then update LOG.

ismaell:
@burgerrg just to clarify the update: I've addressed both things (it required to make the inclusion unconditional) and squashed everything into a single commit.

burgerrg:
Thank you! Do we need the `#include <sys/time.h>` lines in scheme.c and stats.c?

burgerrg (approved):

burgerrg:
Thanks again for this fix!

ismaell:
> On macOS, the sys/time.h header is needed for stats.c.

Neither header requires the other's definitions, so I think that's universal.

PR #555, opened 2021-03-13
avoid hard-coded paths for utilities in build scripts

LiberalArtist:
These changes facilitate building Chez Scheme in environments where utility commands are not available at their usual absolute paths, such as Nix and GNU Guix.

Specific changes:
  - `cc` -> `$(CC)`
  - `/bin/rm` -> `rm`
  - `/bin/ln` -> `ln`
  - `/bin/cp` -> `cp`
  - `/bin/echo` -> `echo`
  - in `makefiles/installsh`, add a case to find `true` at an unusual path or as a shell builtin
    
Related to https://github.com/racket/racket/issues/3707
    
The first commit was merged into Racket's fork of Chez Scheme as https://github.com/racket/ChezScheme/commit/ade94d207592ad7ded079a675d1f2869f0a45a9a and https://github.com/racket/racket/commit/fdb2ad7892933aea1eedd5b199432b988a7df972

This pull request also includes a commit with two minor fixes to the `configure` script: updating the use of `curl` to reflect the update to Nanopass v1.9.2; and fixing the help description for the options `--installschemename`, `--installpetitename`, and `--installscriptname`.


mats/6.ms
```diff
@@ -2685,7 +2685,7 @@
       (begin
         (system "ln -s ../examples .")
         (load "examples/fatfib.ss" compile)
-        (system "/bin/rm examples")
+        (system "rm examples")
```
akeep, line 2688: This may need a `-f`.  Part of the reason for explicitly calling out `/bin/rm` is to avoid aliases of `rm` that change the defaults to prompt before removing files.

mats/unix.ms
```diff
@@ -72,8 +72,8 @@
   (mat system
     (eqv? (with-output-to-file "testfile.ss" void '(replace)) (void))
     (begin
-      (system (format "~:[~;/pkg~]/bin/rm testfile.ss" (embedded?)))
-      (system (format "~:[~;/pkg~]/bin/echo hello > testfile.ss" (embedded?)))
+      (system (format "~:[~;/pkg~]rm testfile.ss" (embedded?)))
+      (system (format "~:[~;/pkg~]echo hello > testfile.ss" (embedded?)))
```
akeep, line 76: This seems not quite right to me.  When `(embedded?)` is `#t` it is going to generate `/pkgrm` an `/pkgecho`, I think the intention was to get either `/pkg/bin/rm` or `/pkg/bin/echo`, if the intention is to always use `rm` or `echo` we could remove the `format` and just have `"rm tetfile.ss"` an `"echo hello > test file.ss"`, though we should probably consider adding `-f` for `rm`.

newrelease
```diff
@@ -92,19 +92,19 @@ sed -e "s/csv[0-9]\.[0-9]\(\.[0-9]\)*/csv$R/" ../makefiles/Mf-install.in > makef
 sed -e "s/csug[0-9]\.[0-9]/csug$MR.$mR/" -e "s/csug[0-9]_[0-9]/csug$MR""_$mR/" ../makefiles/Makefile-csug.in > makefiles/Makefile-csug.in
 set maybeupdatedfiles = ($maybeupdatedfiles makefiles/Mf-install.in makefiles/Makefile-csug.in)
 
-/bin/rm scheme.1.in
+rm scheme.1.in
```
akeep, line 95: We should add `-f` for consistency here.
```suggestion
rm -f scheme.1.in
```

newrelease
```diff
@@ -146,17 +146,17 @@ sed -e "s/Revised\(.*\)for Chez Scheme Version [^ ]*<br>/Revised\1for Chez Schem
   ../csug/csug.stex > csug/csug.stex
 set maybeupdatedfiles = ($maybeupdatedfiles csug/copyright.stex csug/csug.stex)
 
-/bin/rm bintar/Makefile
+rm bintar/Makefile
```
akeep, line 149: We should add `-f` for consistency here.
```suggestion
rm -f bintar/Makefile
```

newrelease
```diff
@@ -146,17 +146,17 @@ sed -e "s/Revised\(.*\)for Chez Scheme Version [^ ]*<br>/Revised\1for Chez Schem
...
 sed -e "s/^version = .*/version = $R/" \
     -e "s/csv[0-9][0-9][0-9]*/csv$ZR/g" \
   ../bintar/Makefile > bintar/Makefile
 set maybeupdatedfiles = ($maybeupdatedfiles bintar/Makefile)
 
-/bin/rm rpm/Makefile
+rm rpm/Makefile
```
akeep, line 155: We should add `-f` for consistency here.
```suggestion
rm -f rpm/Makefile
```

newrelease
```diff
@@ -146,17 +146,17 @@ sed -e "s/Revised\(.*\)for Chez Scheme Version [^ ]*<br>/Revised\1for Chez Schem
...
 sed -e "s/^version = .*/version = $R/" ../rpm/Makefile > rpm/Makefile
 set maybeupdatedfiles = ($maybeupdatedfiles rpm/Makefile)
 
-/bin/rm pkg/Makefile
+rm pkg/Makefile
```
akeep, line 159: We should add `-f` for consistency here.
```suggestion
rm -f pkg/Makefile
```

newrelease
```diff
@@ -175,7 +175,7 @@ foreach x ($maybeupdatedfiles)
   cmp -s ../$x $x || set updatedfiles = ($updatedfiles $x)
 end
 
-/bin/rm LOG
+rm LOG
```
akeep, line 178: We should add `-f` for consistency here.
```suggestion
rm -f LOG
```

akeep (changes requested):
Thanks for the pull request!  I noticed a few places we should probably clean-up around calls to `rm` and at least one place where it is possible the `mats` will generate `/pkgrm` and `/pkgecho` instead of `rm` and `echo`.

Once those are cleaned up, I think this looks good.

mats/6.ms
LiberalArtist: Would just `(delete-file "examples")` work here?

mats/unix.ms
LiberalArtist: I don't think I understand what's going on here. The only definition of `embedded?` I can find is: https://github.com/cisco/ChezScheme/blob/17fd84de28234e3a3e1dc29223a99ab466fc6b87/mats/mat.ss#L506-L507
for which, as with these lines in `mats/unix.ms`, the git blame points to “initial upload of open-source release”. I'm not thinking of a system or context when `/pkg` would usually exist: I guessed it might refer to the support for building Mac `.pkg` packages, but I'm not seeing anything likely there.

(For posterity, [apparently Common Lisp specifies](http://www.lispworks.com/documentation/HyperSpec/Body/22_cgb.htm) that the `format` control string “`~:[`*alternative*`~;`*consequent*`~]` selects the *alternative* control string if *arg* is *false*, and selects the *consequent* control string otherwise.”)

mats/6.ms
akeep: I think I'd jut add the `-f` in keeping consistent with the changes of `/bin/rm` to `rm` elsewhere.

mats/unix.ms
akeep: I don't know that `embedded?` is actually being used right now, but I've not spent enough time going through the code to see if this is something extraneous.  Either way, the change from the Racket leaves the possibility of generating the incorrect `/pkgrm` and `/pkgecho`.  I'm just suggesting that we should avoid the whole issue here and simply remove the formatting, which only exist to add `/pkg`.
```suggestion
      (system "rm -f testfile.ss")
      (system "echo hello > testfile.ss")
```

akeep (changes requested):
Thanks for getting back to me, I've responded to your comments and added the suggestion for the `/pkg` code.  As I mentioned, it is possible that `embedded?` is a vestige of something that is not used, and if so, we should probably come back and remove it, but in the meantime, it is better not to have code that is going to generate something incorrect here.

mats/unix.ms
LiberalArtist: Yes, if that seems good enough for you, it's certainly the simplest.

LiberalArtist:
Thanks for these improvements! I've made the changes and squashed. (I also have a version in 3 commits instead of 2, if you prefer: https://github.com/LiberalArtist/ChezScheme/tree/build-command-paths-3-commits)

akeep:
Looks good, I'm happy with the two commits.  Thanks again for the pull request!

Issue #557, opened 2021-03-28
FFI - Unable to use double pointer to ftype struct in foreign-procedure call


mspaulding06:
After spending a lot of time trying different things I have decided that this may be a bug.  Here is the relevant code example.

```
(define lib (load-shared-object "libssh2.so.1.0.1"))

(define-ftype ssh-agent* uptr)

(define-ftype ssh-agent-publickey
              (struct
                [magic unsigned-int]
                [node void*]
                [blob (* unsigned-8)]
                [blob_len size_t]
                [comment (* char)]))

(define c-ssh-agent-get-identity
  (foreign-procedure "libssh2_agent_get_identity"
                     (ssh-agent* (* (* ssh-agent-publickey)) (* ssh-agent-publickey)) int))
```

When loading this code in the REPL and get the following error:

```
Exception: invalid foreign-procedure argument type specifier (* (* ssh-agent-publickey)) at line 15, char 34 of test.scm
```

If I modify the call to `foreign-procedure` slightly to remove the double pointer like this:

```
(define c-ssh-agent-get-identity
  (foreign-procedure "libssh2_agent_get_identity"
                     (ssh-agent* (* ssh-agent-publickey) (* ssh-agent-publickey)) int))
```

Then everything loads just fine.

I've also tried using `define-ftype` to see if that changes anything with the following definitions:

```
(define-ftype ssh-agent-publickey* (* ssh-agent-publickey))
(define-ftype ssh-agent-publickey** (* ssh-agent-publickey*))

(define c-ssh-agent-get-identity
  (foreign-procedure "libssh2_agent_get_identity"
                     (ssh-agent* ssh-agent-publickey** ssh-agent-publickey*) int))
```

This still fails, but the error message changes slightly to this:

```
Exception: invalid (non-base) foreign-procedure argument ftype ssh-agent-publickey* at line 18, char 55 of test.scm
```

akeep:
It may seem a little odd, but this is actually the intended behavior.

The idea is basically this.  When you say:

```scheme
(define x (make-ftype-pointer ssh-agent-publickey (foreign-alloc (ftype-sizeof ssh-agent-publickey))))
```
[...]

PR #558, opened 2021-04-02
Doc updates

jltaylor-us:
Adds information on building docs to `BUILDING`; updates docs on ftype pointer types in `foreign-procedure` to be more precise (as inspired by #557).


akeep:
This looks good to me!

burgerrg (commented):
Thanks for the update! I have a few comma suggestions.

PR #559, opened 2021-04-04
fix mat that failed in bullyx with cmg=1

jltaylor-us:


mats/4.ms
```diff
@@ -3932,7 +3932,7 @@
    ;; ----------------------------------------
    ;; Check interaction of mutation and incremental generation promotion
 
-   (with-interrupts-disabled
+   (parameterize ([collect-request-handler void] [collect-maximum-generation (max (collect-maximum-generation) 2)])
```
burgerrg, line 3935: Alternate idea:
```
(or (< (collect-maximum-generation) 2)
    (with-interrupts-disabled ...))
```

jltaylor-us: That was my first impulse, but this is the way it's handled in `measure-time` a few dozen lines higher in the file.  Either way is fine with me.

burgerrg: Good find, then I now prefer your current implementation!

burgerrg (approved):

PR #562, opened 2021-04-21
repair bignum shift right

mflatt:
When a negative bignum is shifted right by a multiple of the bigit bit size (which is 32), when a shifted-off bit is non-0, and when the result would be a sequence of bigits with all 1 bits before rounding to deal with the dropped bits, a carry that should have been delivered to a new high bigit was dropped.

Old behavior:
```
> (bitwise-arithmetic-shift (- 307 (expt 16 240)) -32)
0
```

Corrected behavior:
```
> (bitwise-arithmetic-shift (- 307 (expt 16 240)) -32)
-2269007733883335972287082669296112915239349672942191252221331572442536403137824056312817862695551072066953619064625508194663368599769448406663254670871573830845597595897613333042429214224697474472410882236254024057110212260250671521235807709272244389361641091086035023229622419456
```

This bug was discovered by @samth and originally reported as racket/racket#3794, and the example above is by @gus-massa.

### DIFF
--- a/LOG
+++ b/LOG
@@ -2166,3 +2166,5 @@
     configure
 - maybe-compile-program now returns void
     compile.ss 7.ms
+- fixed right shift of a negative bignum by a multiple of 32
+    number.c 5_3.ms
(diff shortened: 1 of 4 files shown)


burgerrg (approved):
Thank you, @samth, for finding this bug and @mflatt for fixing it! I pushed a few minor text changes.

PR #563, opened 2021-04-22
Alternative (R7RS) spelling of booleans

Jarhmander:
This PR allow reading `#true` as `#t` and `#false` as `#f`.

The possible topics of bikeshedding:
- if the reader gets `#tralala`, it will complain that `#tra` is not a valid boolean (message: `read: invalid boolean #tra`); it doesn't try to "intelligently" read a full word to make an error message;
- in such error messages, case isn't retained, so reading `#FALACY` will result in `read: invalid boolean #fala`;
- this new spelling is disabled with `#!r6rs`; the exact message when `#true` is read is `read: alternative boolean syntax is not allowed in #!r6rs mode`.

_It ain't much but it's honest work_.


jltaylor-us:
The travis-ci jobs have failed because the `summary` file produced by the `*x` targets (`partialx`, `allx`, and `bullyx`) says that there are a different set of errors than was expected.  (The current travis-ci jobs don't actually run one of those targets, but the mechanism is the same.)  You need to update the files listing the expected errors for all of the various combinations of build options that can be set when running the mats.  There is a small set of "base" files and then several patch files used to build the files for other configurations.

I was going to write that I have not done this before, myself, but according to the git log, I have.  I don't remember it, though, and it may have been a case of Kent telling me what to type to get that bit done.  In any case, there are some instructions in #306 (although this change may involve updating more of the files).  If you can't figure out what to do from those instructions then post back here and someone will help.

Jarhmander:
I think I fixed it, and to be sure not to embarrass myself again with failed tests, I tried the travis script locally and it worked this time. It *should* work now. Sorry for the noise, and thanks for the very welcome support.

(I naively assumed that if `make 6.mo` worked (no "Bug" lines), then the tests will pass... I wasn't aware of the root-experr et al)

Jarhmander:
The [build #877 passed](https://travis-ci.org/github/cisco/ChezScheme/builds/768116931), but there's no checkmark in here for some reason.

burgerrg:
Please update the release notes and rebase to resolve merge conflicts.

Jarhmander:
Updated, and tests pass (unsurprisingly).

burgerrg:

PR #564, opened 2021-04-25
Mention an up-to-date DLL for Windows

Jarhmander:
The user guide mention an outdated DLL which does not necessarily exist in recent installments of Windows 10. [According to Wikipedia](https://en.wikipedia.org/wiki/Microsoft_Windows_library_files#MSVCRT.DLL,_MSVCP*.DLL_and_CRTDLL.DLL), `MSVCRT.DLL` exists since Windows 95, and I tested on a Windows XP VM that it does exist in that (unsupported) Windows version.

### DIFF
--- a/LOG
+++ b/LOG
@@ -2168,3 +2168,6 @@
     compile.ss 7.ms
 - fixed right shift of a negative bignum by a multiple of 32
     number.c 5_3.ms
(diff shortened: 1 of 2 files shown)


burgerrg:
Thank you for this update!

PR #566, opened 2021-04-29
Self-evaluating vectors

Jarhmander:
This is a *very* simple patch to make vectors self-evaluated.

The consequence of merging this PR is that it makes Chez Scheme no longer R6RS compliant, in the strict sense: section 11.13 of the revised report says, "Like list constants, *vector constants must be quoted*". On the other hand, the only programs that would be affected by that change are those that *specifically* check for violations arising from evaluating a non-quoted vector, which would only be possibly Scheme programs that test the R6RS implementation to the fullest; I believe that no *practical* program would be negatively affected by that change.


akeep:
Hey Jarhmander,

Thanks for the pull request, but as Chez Scheme is an R6RS implementation, I'd rather not move away from the required quoted vector literal syntax.  For what it is worth, I believe part of the motivation for making vector literals not self-evaluating is that unlike byte vectors and fxvectors, which can only contain self evaluating forms, vectors can contain any scheme datatype, meaning suddenly in the context of the self-evaluating vectors, these forms too are self evaluating.

In order to add this sort of support, I think we would need to have an R7RS specific mode, potentially controlled by a #!r7rs signifier, similar to the existing #!r6rs and #!chezscheme signifiers, controlling the reader and very likely setting a parameter to control evaluation of the expressions that follow it in R7RS mode.

I would also like to see the implementation function more like the traditional quoted vector syntax that currently exists.  For instance, if I write the macro:

```scheme
(define-syntax v
  (syntax-rules ()
    [(_ elem ...) #(elem ...)]))
```
Then I would expect to get vectors of the specified arguments, so calling `v` as:
```
> (v 1 2 a b)
```
should result in the syntax `#(1 2 a b)`, but the current implementation results in:
```
#(#<syntax 1> #<syntax 2> #<syntax a> #<syntax b>)
```

One solution would be to treat this as though it was a quoted vector.  A quoted vector is walked to strip syntax annotations:

```scheme
> (define-syntax qv
    (syntax-rules ()
      [(_ elem ...) '#(elem ...)]))
> (qv 1 2 a b)
#(1 2 a b)
```

The other solution is to follow what Chez Scheme currently does with bytevectors and fxvectors:

```scheme
> (define-syntax bv
    (syntax-rules ()
      [(_ n ...) #vu8(n ...)]))
----------------------------------------
read: invalid value n found in bytevector
----------------------------------------
> (define-syntax fxv
    (syntax-rules ()
      [(_ n ...) #vfx(n ...)]))
-----------------------------------
read: non-fixnum found in fxvector
-----------------------------------
```

Unfortunately, the later is tricky to do with general vectors, because nothing is restricted from their contents, so ruling them out in the reader isn't really possible.  It is another motivation for why R6RS decided not to allow self-evaluating vectors.  They seem fine at the service, but the moment you start to use them in more advanced settings, like macros, they start to fall apart.

Jarhmander:
Ahah! I knew that it couldn't be that easy. In fact, I was surprised that it didn't affect at all quasiquote. I should have looked at the effect of syntax forms...
You suggest having an #!r7rs comment designator. As far I can see (but I might be wrong!), current comment designators only affect the reader, and these have no effect on the evaluation of forms. It seems that the "reader mode" (#!chezscheme or #!r6rs) is an information attached to the port where datums are read. So it can be weird that it affects compilation, too.
Maybe the best is then not pretend that vectors are self-evaluated and have a parameter that change how non-quoted vectors are treated (error or "self-quoting") during compile/eval. Something like `self-quoting-vector,` maybe?

akeep:
Yeah, you are correct that the `#!chezscheme` and `#!r6rs` designators only impact the reader.  The other differences between Chez Scheme and R6RS are controlled through exported identifiers, for instance R6RS's `fx+` function expects exactly two arguments, while the Chez Scheme's version of this function can accept any number of arguments, so `(import (r6rs))` imports the `r6rs:fx+` identifier as `fx+` while `(import (chezscheme))` imports the Chez Scheme version's identifier.

Unfortunately, changing which forms self-evaluate is trickier since it falls outside of these categories.  I don't know what other differences would be required to support R7RS, but I think we'd want to collect them up to enable/disable all at once, though you may be correct that `self-quoting-vector` is a better option name for this.

Jarhmander:
Yeah, nothing prevent having a "low-level" parameter (`self-quoting-vectors`) for controlling the specific bit of evaluation of vectors, and having something more high-level that deals with R7RS support, which would then control low-level parameters like `self-quoting-vectors`.

Putting momentarily aside the parameter question, I tried a different approach for the evaluation of vectors. I reset back to master and rather than messing with "what is self-evaluating?", I fiddle with how the forms are interpreted. My approach here is, "if you ever see a vector literal, act as if it is quoted". `syntax-type` seems to be what is closer to what is being read, and deals with macro expansion. So, (pardon my very unscientific method), I tried this: 

```diff --git a/s/syntax.ss b/s/syntax.ss
index 75aee67..fe7a2ca 100644
--- a/s/syntax.ss
+++ b/s/syntax.ss
@@ -1861,6 +1861,8 @@
             (syntax-type (chi-macro (binding-value b) e r w ae rib)
               r empty-wrap ae rib)]
            [else (values type (binding-value b) e w ae)]))]
+      [(vector? e)
+       (syntax-type #`(quote #,e) r w ae rib)]
       [(self-evaluating? e) (values 'constant #f e w ae)]
       [else (values 'other #f e w ae)])))

```

Non-quoted vectors doesn't work in the REPL (attempt to reference unbound identifier quote), *but* if we try your macro:
```
> (define-syntax qv
    (syntax-rules ()
      [(_ elem ...) #(elem ...)]))
> (qv 1 2 a b)
#(1 2 a b)
```

This works, and even better, `quote` is not taken from the environment of the evaluated forms, so this also works:
```
> (let ((quote (lambda (_) (display "FAIL\n"))))
    (qv 1 a 2)
    "OK")
"OK"
```
So it puzzles me a bit that quote isn't visible at top-level, but it is in a macro. On the other hand, if in `syntax-type`, we change syntax-quasiquote and syntax-unquote by simple quasiquote and unquote, the unquoted vector works in the REPL and doesn't in the `qv` macro.

Jarhmander:
Sorry for the double reply. It just occurred to me that the syntax stripping of quoted vectors is done by `strip`, and it is clearly seen in the value of `(global-extend 'core 'quote ...)`; I just have to strip the form myself to get the desired effect.

With this:
```diff
diff --git a/s/syntax.ss b/s/syntax.ss
index 75aee67..822b9eb 100644
--- a/s/syntax.ss
+++ b/s/syntax.ss
@@ -1861,6 +1861,8 @@
             (syntax-type (chi-macro (binding-value b) e r w ae rib)
               r empty-wrap ae rib)]
            [else (values type (binding-value b) e w ae)]))]
+      [(vector? e)
+       (values 'constant #f (strip e w) w ae)]
       [(self-evaluating? e) (values 'constant #f e w ae)]
       [else (values 'other #f e w ae)])))
```
, it seems to work both with simple unquoted vectors and unquoted vectors in macros:
```
> #(1 a 2)
#(1 a 2)
> '#(1 a 2)
#(1 a 2)
> `#(1 ,+ 2)
#(1 #<procedure +> 2)
> (define-syntax qv
    (syntax-rules ()
      ((_ a ...) #(a ...))))
> (qv 1 a 2)
#(1 a 2)
> (let ((quote (lambda (_) (display "FAIL\n"))))
    (qv 1 a 2)
    "OK")
"OK"
```
Is this adequate? In other words, if I update the tests (modify/add new tests), and add a new parameter for controlling the evaluation of unquoted vectors (so the test in `syntax-type` reads `(and (self-quoting-vectors) (vector? e))`), along with its documentation, will this PR be a candidate for merging? If the new parameter is really desired, I can add it and document it, but I'd rather postpone anything bigger that deals with R7RS at large, to limit the scope of this PR.

Jarhmander:
I updated the branch with the last suggested implementation above + rebase on main, to ease reviewing and testing. It's not complete because it needs a bit of documentation and proper testing, but you can pull and play with this.

If the change looks OK to you, I can proceed to add the parameter, add a few tests and update the docs.

akeep:
Hey @Jarhmander,

Apologies for the slow response, a little bit of me getting busy and a little bit of me wanting to consult with the other maintainers before accepting a change this large.  We are still chatting a bit about the details, but overall I think the changes you've commented about make sense and seem consistent with how the r7rs documentation writes about the template language for `syntax-rules`.

As I mentioned, we'd wan to make sure this is only enabled when a parameter is set to enable self-evaluating vectors.

One of the details we've talked a bit about is the potential for having a command-line argument that specifies which standard should be used when processing code.  Something along the lines of `--compat r7rs` (or `r7rs-preview` since we only have a few features added at this point), would enable this parameter, and in the long run select any r7rs specific reader or library options.

Jarhmander:
I added the new parameter `self-evaluating-vectors`, along with some tests. I wasn't too sure of where to document the parameter; I put it in section 7.5, "Vectors", in "Operations on Objects".

As it is, I believe the PR is complete in its scope. High-level compiler controls for R7RS would be a nice addition to Chez Scheme, but it's best suited in another PR in my opinion. Also, the new parameter wouldn't be unlike `internal-defines-as-letrec*`, which control a very specific

akeep (commented):
Thanks for the update.  I wanted to test using this in an r6rs library, since in the long run, I expect support for r7rs libraries will be based on the r6rs model.  Anyway, I was a little surprised by the result:

```scheme
(self-evaluating-vectors #t)
(library (t)
  (export v)
  (import (chezscheme))
  (define v #(a b c)))
```

When I imported this at the repl and looked a `v` instead the expected `#(a b c)`, I see:

```
> (import (t))
> v
#(#<annotation /tmp/t.ss[93:94] a>
  #<annotation /tmp/t.ss[95:96] b>
  #<annotation /tmp/t.ss[97:98] c>)
```

I've not looked into why the annotations are showing through, but this should be fixed before we incorporate it.  We probably should add something like this to the test suite and try macro generating macros, etc. to make sure we've covered the corner cases.

Let me know if you need some help looking into why this is happening.

Jarhmander:
I addressed the error in the documentation as well as the last defect. What's left is to correctly test for that bug.

This is what I used to help me figure out what's going on:

```scheme
(source-directories '("ta6le/s" "."))
(library-directories "nanopass:.")
(for-each
  (lambda (e) (load (symbol->string e)))
  '(cmacros.so priminfo.so primvars.so env.so setup.so))
(load "syntax.ss")
(subset-mode #f)
```

... and then `trace`ing `strip` and parts of it. I saw that quoted expressions are annotated as a whole, whereas self evaluating vectors like what I implemented have each elements individually annotated. There are some parts of `strip` that I don't understand; for instance, it seems that it can iterate in vectors to strip them, but I don't know when that happens. Do you have a clue?

Jarhmander:
Sorry if the last comment wasn't clear enough. I'm not quite sure if `strip` itself is working as it should or if it should be fixed, because in some cases it still check pairs but not other structures like vectors, and for me it's not clear why. If `strip` needs fixing, the only consequence is to simplify `syntax-type` and make it cleaner.

This morning I tried nested structures of vectors like `#(a (b #(d) c) #(e #(f #(g) (h #(i)))))` and it seems to work well, even inside crazier macros like these:
```scheme
 (define-syntax qv2
   (syntax-rules ()
     ((_ a ...) #(#(a (a #(a))) ... #(aa (cc #(dd () ee)) bb cc)))))
```

This is not too surprising, because only the elements themselves are annotated, not the elements' content.

I'll write new tests and update this PR soon.

Jarhmander:
New tests written. I tested crazier macro-generating macro to ensure that the revised implementation is correct, and I verified that I got "Bug in clause" messages when using the previous implementation, at least when directly testing misc.mo with `make misc.mo`.

I *hope* this is it.

akeep:
Thanks for the changes, merged!

mnieper:
A sensible approach to control how datums like (unquoted) vectors are evaluated (and, in my opinion, a much better approach than using a parameter) is Racket's implicit `#%datum` syntax ([1]), suitably made compatible with R6RS. `#%datum` special forms are inserted by the expander.

In order to make this fully compatible with R6RS, the `#%datum` identifier must be special in that it is not a regular R6RS identifier. In particular, `(datum->syntax #'#%datum (string->symbol "#%datum"))` must not be `bound-identifier=?` or `free-identifier=?` to `#%datum`. 

The `#%datum` identifier will be bound in the top-level environment of each library and program. Here, we can control whether we get R6RS or R7RS semantics depending on how this identifier is bound.

In order to allow the user to change this aspect, a procedure like `(special->syntax ID NAME)` can be introduced, which creates a special identifier with the marks and substs of `ID` and with name `NAME` (e.g. `#%datum`). These special identifiers are *not* wrapped symbols.

In the same manner, the other similar forms of Racket like `#%app` can be integrated. From a language designer's point of view, this has the advantage that section 9.1 of R6RS can be moved into the base library (documenting the default binding to `#%datum` and `#%app`).

--

[1] https://docs.racket-lang.org/reference/quote.html

PR #567, opened 2021-05-05
x86_64: fix corner case of foreign calls for signed 8-bit or 16-bit integers with &

mflatt:
For x86_64, it seems that integer values passed in registers should be sign-extended to at least 32 bits. In the case that a foreign-call argument is passed via `(& integer-8)` or `(& integer-16)` — which no one is likely to do in practice! — the sign extension didn't happen.

Clang on macOS apparently only takes advantage of the sign extension when compiling with optimization, but  "foreign1.mo" is compiled without optimization, so the existing test didn't find the problem. This patch changes the "s/Mf-..." files for x86_64 macOS and Linux to include `-O2` when compiling, but a more consistent change probably would be better. (For the Racket branch of Chez Scheme, the makefiles have been revised so that flags used for compiling the kernel are also used to compile `foreign1.mo`, and that's how the bug was exposed.)

### DIFF
--- a/LOG
+++ b/LOG
@@ -2185,3 +2185,6 @@
     mats/{5_4,6,7,8,bytevector,examples,foreign}.ms
     mats/{ftype,hash,io,misc,primvars,profile,record}.ms
     mats/Mf-base mats/Mf-*nt mats/mat.ss mats/patch-interpret*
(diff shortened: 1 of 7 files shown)


burgerrg:
Thank you for finding and fixing this bug!

burgerrg (approved):

Issue #565, opened 2021-04-26
Creating a compile-time-value for one of the native basic ftypes (char, unsigned-32, etc.) breaks ftype-related macros


lemvik:
Hi!

I was trying to build a set of FFI helpers on top of `ftype-...` macros and I encountered the following problem.

I wanted to add a [compile-time-value](https://cisco.github.io/ChezScheme/csug9.5/syntax.html#./syntax:s15) to some identifiers that correspond to primitive types to be able to generate a more efficient expansion for some of my macros if they were to work on such primitives. However, adding a cpv to any identifier (for example, `char`) of the primitive type breaks `make-ftype-pointer`/`ftype-set!`/`ftype-ref`/etc for that identifier.

I guess that the reason behind it is in the order checks are performed [here](https://github.com/cisco/ChezScheme/blob/9b4fff71852a0b7a6d758a0fd4ed74e4eb9dde02/s/ftype.ss#L386). I'm not sure what would be implications of changing the order of `cond` clauses there (or if that is even the correct fix).

Example from REPL:
```
> (make-ftype-pointer char 0)
#<ftype-pointer char 0>
> (define-syntax char (make-compile-time-value "I'm a character"))
> (make-ftype-pointer char 0)
Exception: unrecognized ftype name char
Type (debug) to enter the debugger.
```

lemvik:
1. `char`, for example, is not a compile-time value:
```
> (define-syntax get-compile-time-value
    (lambda (x)
      (lambda (lookup)
        (syntax-case x ()
          [(_ id)
           #`'(compile-time-value -> #,(lookup #'id))]))))
> (get-compile-time-value char)
(compile-time-value -> #f)
> (define-syntax char (make-compile-time-value "I'm a char"))
> (get-compile-time-value char)
(compile-time-value -> "I'm a char")
```
2. If it was a compile-time value then it's not really clear from documentation. The snippet you've referenced says nothing about ftypes being compile-time-values (even though ones defined via `define-ftype` are indeed cpv).
3. I could try to go with properties (and I tried, actually), but defining a property requires the existence of a binding for the identifier (in this case, `char`), since properties are tied to bindings, not identifiers ("`attaches a property to an existing identifier binding`"). I would have to do something like defining my own bindings for primitive values (which is not currently done in Chez, as evaluating `char` in REPL returns `variable char is not bound`) and **making sure they are not shadowed** by client code, which can happen by just introducing some `let ([char #\a])` somewhere in client code.

[...]

PR #569, opened 2021-05-10
Create compile-time values for base ftypes.

akeep:
Converted built-in foreign types to work like other ftypes using a
compile-time value to specify the value of the foreign type.

This change is intended to enable the base-types to be treated like other ftypes which are bound as compile time values to address #565.


akeep:
Updating with an alternative solution.  Instead of introducing new top-level bindings for the native ftypes, this change simply ignores compile-time environment bindings to non-ftd data for the native types.  The upside of this is that it avoids creating a bunch of bindings for things that might interfere with existing code, but still addresses the issue in #565.

Additionally, it is now possible to create bindings like the following:

```
(define-ftype integer-32 integer-32)
```
Which adds a normal ftype binding for the native ftype `integer-32`, essentially allowing users to add binding for the native ftypes if they so desire.  The latter change works by treating the native ftypes the same way as normal ftypes to detect recursive bindings, essentially looking up the ftd from the native ftype list if the binding doesn't already exist as a normal ftype.

akeep:
Thanks for the feedback, @burgerrg!  I believe I've addressed it and also tested out the code examples and realized I had a bug in the `(endianness <endian> <ftype>)` instead of `(endian <endian> <ftype>)`.

burgerrg (approved):

PR #570, opened 2021-05-12
fix misnamed pattern variables in bytevector-*-ref

laqrix:
This fixes misnamed pattern variables in bytevector-*-ref. It appears to be a cut & paste bug since before open-sourcing.

Practically, this has no effect as the affected code is avoided by `unaligned-integers` which appears to be `#t` for all currently supported platforms.  Flipping `unaligned-integers` to `#f` appears to emit the proper code, but there are other compile errors down the chain.  There may be a larger issue here of whether `unaligned-integers` should be supported option.


cjfrisz:
Thanks for the update!

PR #572, opened 2021-05-17
add workaround for Rosetta 2 bug in [t]a6osx builds

jltaylor-us:
From the description in S_bytevector_read in c/new-io.c:
  /* If we are running on Apple Silicon under Rosetta 2 translation, work around
     a bug (present in 11.3.1 at least) in its handling of memory page protection
     bits.  One of the tasks that Rosetta handles is to appropriately twiddle the
     execute and write bits based on what's happinging to the memory in order to
     preserve the illusion that the pages have RWX permissions, whereas Apple
     Silicon enforces a W^X (write XOR execute) model.  For some reason, this
     bit-twiddling sometimes fails when the bytevector passed to `read` extends
     onto a page that's currently R-X, causing the `read` to fail with EFAULT
     ("bad address").  By writing to each subsequent page, we force Rosetta to
     do the right magic to the protection bits.  (Or at least it makes the error
     go away and all the mats pass.)
     */

This workaround can be removed completely by building with
NO_ROSETTA_CHECK defined.

If NO_ROSETTA_CHECK is not defined and we are building a6osx or ta6osx
then a new global variable `is_rosetta` is added and intitialized during
s_init.  This variable is consulted in S_bytevector_read.

### DIFF
--- a/LOG
+++ b/LOG
@@ -2190,3 +2190,9 @@
     x86_64, s/Mf-[t]a6{le,osx}
 - fixed misnamed pattern variables in bytevector-*-ref
     bytevector.ss
+- add workaround for Rosetta 2 bug in [t]a6osx builds.  The overhead
+  during native (x86) execution is the addition of a single int compare
+  and branch in S_bytevector_read, plus a small amount of work at startup
+  to determine whether we are running under Rosetta translation.
+    c/new-io.c c/scheme.c c/version.h
+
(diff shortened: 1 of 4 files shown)


jltaylor-us:
One would hope this is a temporary workaround, but the change for separate code allocation segments (#573) that I would have expected to solve this problem did not.  It reduced the frequency of occurrences during `bullyx` (in my two data points) from 10 to 3, but did not eliminate them completely.  (This workaround does.)


Does this warrant an entry in the release notes?

PR #573, opened 2021-05-17
Separate code allocation segments

jltaylor-us:
Backport from Racket - distinguish code vs. non-code allocation segments.  This will be required to port to architectures that enforce W^X semantics (e.g., Apple Silicon).  At least I'm pretty sure it's required.  We can always wait until we have the rest of arm64osx implemented to see if it really is required before merging this in.

Requires through review.  All `bullyx` mats pass, but I'm not entirely sure that the code in `gcwrapper.c` does the right thing (since one of the comments says `space_code` is not supported in whatever that section of code is doing).  The Racket GC appears to have been largely rewritten, so comparing to that was not helpful.

### DIFF
--- a/LOG
+++ b/LOG
@@ -2190,3 +2190,5 @@
     x86_64, s/Mf-[t]a6{le,osx}
 - fixed misnamed pattern variables in bytevector-*-ref
     bytevector.ss
+- distinguish code vs non-code allocation segments.  This will be
+  required to port to architectures that enforce W^X semantics.
--- a/c/externs.h
+++ b/c/externs.h
@@ -341,7 +341,7 @@ extern INT matherr PROTO((struct exception *x));
 
 /* segment.c */
 extern void S_segment_init PROTO((void));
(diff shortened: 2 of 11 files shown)


jltaylor-us:
I should mention that the GitHub diff does a particularly poor job identifying the actual changes in `gcwrapper.c`.  You may want to do the diff yourself with the "ignore whitespace changes" option (`-b`).

michaellenaghan:
> I should mention that the GitHub diff does a particularly poor job identifying the actual changes in `gcwrapper.c`. You may want to do the diff yourself with the "ignore whitespace changes" option (`-b`).

In case you weren't aware, you can ignore whitespace in GitHub diffs: 

    https://github.blog/2018-05-01-ignore-white-space-in-code-review/

Look under the gear icon in the Files Changed view for the setting. 

And, as noted in the post, you can also manually add "w=1" to the URL parameters; in fact, that's exactly what the setting does.

PR #575, opened 2021-05-23
Fix the release notes so `make docs` succeed

Jarhmander:
Commit 0d5ec1da7 broke the release notes, so a fix is needed in order to generate the docs.


burgerrg:

PR #594, opened 2021-11-25
github workflow update

jltaylor-us:
Always run the archive/upload workspace steps (even if previous
steps failed or timed out).

Set an explicit timeout for the "Run tests" step.  (Avoid waiting
for the default job-wide 6 hour timeout if the tests hang.)


burgerrg:
Thanks for this helpful update!

PR #596, opened 2021-12-04
define-record-type methods and interfaces

dybvig:
This pull request extends `define-record-type` with methods, effectively converting `define-record-type` into a class-definition form in support of object-oriented programming.   It adds a `define-interface` form and a `define-record-type` `implements` clause to support multiple interface inheritance.  While object-oriented programming obscures control flow, tends to encourage an overly imperative style of programming, and is often overused when it's available, it has its uses when used judiciously.  This particular OOP subsystem, at least, supports a functional style of programming: allocation and initialization is inexpensive as well as functional, and fields are immutable by default.  The essentials of the implementation come from a `define-class` form that Oscar Waddel and I designed and implemented many years ago, the code for which had been sitting in mats/oop.ss; only its recasting into an extension of `define-record-type` is new.

The pull request adds a couple of other things that are useful independently of the support for methods and interfaces; these are described in the LOG.


s/record-defn.ss
```diff
@@ -0,0 +1,1204 @@
+;;; record-defn.ss
+;;; Copyright 1984-2017 Cisco Systems, Inc.
```
jltaylor-us, line 2:
```suggestion
;;; Copyright 1984-2021 Cisco Systems, Inc.
```

jltaylor-us (commented):
Cool!  CSUG updates coming?

I suppose theoretically all of the edited files that have the standard copyright / license header should have the copyright date updated.

gwatt:
Will there be a way to ask if an object implements an interface without walking the result of `record-type-interfaces`? An auto-generated procedure named `interface-name?` like what `define-record-type` does would be nice.
Also, when specifying either a `methods` or `implements` clause, I can't seem to retrieve rtd information with `record-type-descriptor`; it says `unrecognized record: Foo`. Is that intentional?

s/record-defn.ss
dybvig: All fixed.

steelhopper:
已收到

dybvig:
@jltaylor-us I will add csug updates and a release note when and if we decide to proceed with this.  Thanks for the fixes, which will be incorporated in my next commit.

dybvig:
@gwatt `record-type-descriptor` should work for record types with methods but didn't due to a missing call to unwrap-drtinfo.  Thanks for pointing that out and for the suggestion for `interface-name?` predicates.  Both changes will be in my next commit.

SumanSunderroy:
Would the facilities for custom readers, writers, hash functions, and equality tests that are currently present in Chez Scheme be made part of this interface facility?

dybvig:
@sunderroy Yes, there's nothing preventing operations on records in general from working on records with methods.

LOG
```diff
@@ -2301,3 +2301,67 @@
...
 - updated comments describing the new features
     record-defn.ss
+- fixed newly improved generate-temporaries so it doesn't loop
+  indefinitely on cycles in the elemenents of the input list.
+    syntax.ss,
+    8.mo, root-experr-compile-0-f-f-f
+- cp0 record-construct handlers now keep record-type wrappers when
+  propagating constants and refs.
+    cp0.ss
+- added immutable-vector primitiive along with cp0 handling of
```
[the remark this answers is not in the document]
dybvig: Fixed, thanks.

s/record-defn.ss
```diff
@@ -416,6 +424,56 @@ products:
...
+    (lambda (x)
+      (syntax-case x ()
+        [(_ iface-name ([local-mname iface-mname] ...) obj-expr)
+         (lambda (env)
+           (let ([local-mname* #'(local-mname ...)])
+             (unless ($distinct-bound-ids? local-mname*)
+               ($invalid-ids-error local-mname* x "local method name"))
+             (let* ([diinfo (unwrap-diinfo (env #'iface-name))]
+                    [iface-rtd ($diinfo-rtd diinfo)])
+               (unless ($diinfo? diinfo) (syntax-error #'iface-name "unrecognized interface"))
```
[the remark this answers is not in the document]
dybvig: Yup.  Fixed, thanks.

mats/record.ms
```diff
@@ -9459,11 +9459,36 @@
...
+  (error? ; invalid syntax
+    (define-interface (rat (israt?))))
+  (error? ; invalid syntax
+    (define-interface ((rat) israt?)))
+  (error? ; invalid syntax
+    (define-interface (rat israt?) (methods . none)))
+  (error? ; invalid syntax
+    (define-interface (rat israt?) (methods a b c)))
+  (error? ; invalid syntax
+    (define-interface rat (mehtods)))
```
genericist, line 9479:
```scm
(define-interface rat (mehtods))
```
should be
```scm
(define-interface rat (methods))
```

genericist (commented):
Just a little typo I spotted :)

PR #603, opened 2022-01-12
Fix mats on Windows with MSYS

melted:
The absolute path for SCHEMEHEAPDIRS set in Mf-Base might be unix-style
on MSYS, which will not be understood by scheme. Setting it to a relative
path will work regardless of the environment used to build.

Tested with Cygwin and MSYS2.


melted:
Hmmm, it doesn't work on the CI setup. Let's not do that then.

PR #607, opened 2022-02-07
Apple Silicon support

cjfrisz:
This changeset adds an `arm64mac` machine type, with draft implementations for the `define-registers` and section 2 `define-instructions` in `arm64.ss`. The `arm64.ss` file is based off of `arm32.ss`, and so the section 3 assembler is unchanged from `arm32.ss`. The code included compiles, but (unsurprisingly) fails to bootstrap.


cjfrisz:
I've left changes in the `arm64mac` workarea, so the diff is larger than the actual changeset. Changes are primarily in `arm64.ss` and `cpnanopass.ss`.

michaelmesser:
Will this be merged? If so, about how long will it take for this to be merged?

cjfrisz:
It can't be merged in its current state because it effectively has no assembler and doesn't pass the mats. There's no set timeline for merging it, but I see the mentioned issue that Idris can't be installed on Apple Silicon via Homebrew without it.

Until this PR is complete, is it possible to install the Intel version of Idris via Homebrew and run through Rosetta 2?

michaelmesser:
Idris can use the racket fork of chez scheme on M1 to run natively. If this will take awhile, I can try adapting the brew formula to do that instead.

mflatt:
Apple Silicon support is included in v9.9.9 merge 47daa9b34016de84fd111801d9d589d15a523abe.

chenrui333:
Can you cut off a new release per this issue close?

relates to https://github.com/Homebrew/homebrew-core/pull/152760

burgerrg:
Yes, the next release will include this support.

PR #610, opened 2022-02-14
Fix rational-valued? for numbers with an exceptional real part

gus-massa:
The result of

    (rational-valued? +inf.0+0.0i)

is `#t` but it should be `#f`.

### DIFF
--- a/LOG
+++ b/LOG
@@ -2259,3 +2259,5 @@
     BUILDING, c/vs.bat, wininstall/locate-vcredist.bat
 - make threaded foreign mats more robust
     mats/foreign.ms
+- fix rational-valued? for numbers with an exceptional real part
+    s/5_3.ss mats/5_1.ms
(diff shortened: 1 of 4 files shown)


jltaylor-us:
Looks good to me.  Please update the Bug Fixes section of release_notes.stex with something like this

```
\subsection{\protect\scheme{rational-valued?} and exceptional flonums (9.5.8)}

The \scheme{rational-value?} function returned incorrect results when called on
a value with an inexact zero imaginary part and real part that is an exceptional
floating point value (i.e., an infinity or NaN).  For example,
\scheme{(rational-valued? +inf.0+0.0i)} incorrectly returned \scheme{#t}, but now
returns \scheme{#f}.
```

gus-massa:
Done. I hope I copied the text to the correct spot in the file.

jltaylor-us:
Thanks for the great find!

PR #611, opened 2022-02-16
fix exact/inexact comparisons where fixnum->flonum had loss of precision

### LINKED ISSUE #606: = can be not transitive
R6RS requires `=` to be transitive.

However, with a case from [this page](https://codewords.recurse.com/issues/one/when-is-equality-transitive-and-other-floating-point-curiosities): `(= 9007199254740992.0 9007199254740993)` => `#t`.

This is not conforming. The requirement is same from R5RS to R7RS, and there is a note in R7RS stating "the implementation  approach of converting all arguments to inexact numbers if any argument is inexact is not transitive".

(I'm not sure whether this is intended in R6RS, since [this ticket](https://small.r7rs.org/ticket/367/) suggests inexact contagion is allowed generally.)

jltaylor-us:
makes <=, =, and >= transitive as required by R6RS by changing behavior
of mixed exact/inexact comparisons in the range where fixnums have
greater precision than flonums.  For example, the flonum representation
of 9007199254740992.0 and 9007199254740993.0 is identical,
but obviously the fixnums 9007199254740992 and 9007199254740993
are not.  In order for = and the inclusive inequalities to be transitive,
exact/inexact comparisons must behave as if the inexact number was replaced by
its exact representative.  A similar change was made for the strict
inequalities as well (even though they were already transitive).

Fixes #606

### DIFF
--- a/s/5_3.ss
+++ b/s/5_3.ss
@@ -34,6 +34,10 @@
       (float-type-case
          [(ieee) -1023]))
 
+   (define-constant float-precision-bits
+      (float-type-case
+         [(ieee) 53]))
+
 )
 
 (let ()
@@ -660,6 +664,13 @@
          [($exactnum?) (and (floatable? (real-part x))
                              (floatable? (imag-part x)))]
          [else #t])))
+
+(define fixnum-floatable-wlop?
+  ;; floatable without loss of precision
+  (lambda (x)
+    (if (<= (fixnum-width) (constant float-precision-bits))
+        #t
+        (fx<= (- (expt 2 (constant float-precision-bits))) x (expt 2 (constant float-precision-bits))))))
 
 (define exact-inexact-compare?
    ; e is an exact number, i is a flonum
@@ -2004,7 +2015,7 @@
           (type-case y
              [(fixnum?) (fx= x y)]
              [(bignum? ratnum? $exactnum?) #f]
-             [(cflonum?) (cfl= (fixnum->flonum x) y)]
+             [(cflonum?) (if (fixnum-floatable-wlop? x) (cfl= (fixnum->flonum x) y) (exact-inexact-compare? = x y))]
              [else (nonnumber-error who y)])]
          [(bignum?)
           (type-case y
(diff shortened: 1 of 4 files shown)


s/5_3.ss
```diff
@@ -661,6 +665,13 @@
                              (floatable? (imag-part x)))]
          [else #t])))
 
+(define fixnum-floatable-wlop?
+  ;; floatable without loss of precision
+  (lambda (x)
+    (if (<= (fixnum-width) (constant float-precision-bits))
```
luke21923, line 671: Let's assume `(fixnum-width)` is equal to `24`. Then fixnums are in the range [-2^23, 2^23 - 1].

Let's assume `(constant float-precision-bits)` is equal to `23`. Then integers in the range [-2^23, 2^23] can be stored in the mantissa of a float without loss of precision.

Therefore, I would replace the line 671 of the file s/5_3.ss with:

    (if (<= (- (fixnum-width) 1) (constant float-precision-bits))

The width of a fixnum includes the sign bit, whereas the width of the mantissa of a float excludes it.

Also, I have suggestions to improve the names of two variables:

* Replace the name of the constant `float-precision-bits` with the name `float-mantissa-width`, because it is not clear what "precision" means (for instance, the precision of a float could be 64 in the case of IEEE 754 binary64).
* Replace the name of the procedure `fixnum-floatable-wlop?` with the name `fixnum-fits-in-mantissa?`. Even though there is an edge case in which this new name is misleading (the integer 2^53 is 54 bits wide, and therefore does not fit in a 53 bits mantissa), I think it is a better name because this procedure rejects a lot of fixnums that **are** floatable without loss of precision, for instance 2^55:

      > (= (expt 2 55) (exact (inexact (expt 2 55))))
      #t

luke21923 (changes requested):

luke21923:
I would replace the line 671 of the file s/5_3.ss with:

    (if (<= (- (fixnum-width) 1) (constant float-precision-bits))

Let's assume `(fixnum-width)` is equal to `24`. Then fixnums are in the range [-2^23, 2^23 - 1].

Let's assume `(constant float-precision-bits)` is equal to `23`. Then integers in the range [-2^23, 2^23] can be stored in the mantissa of a float without loss of precision.

The width of a fixnum includes the sign bit, whereas the width of the mantissa of a float excludes it.

Also, I have suggestions to improve the names of two variables:

* Replace the name of the constant `float-precision-bits` with the name `float-mantissa-width`, because it is not clear what "precision" means (for instance, the precision of a float could be 64 in the case of IEEE 754 binary64).
* Replace the name of the procedure `fixnum-floatable-wlop?` with the name `fixnum-fits-in-mantissa?`. Even though there is an edge case in which this new name is misleading (the integer 2^53 is 54 bits wide, and therefore does not fit in a 53 bits mantissa), I think it is a better name because this procedure rejects a lot of fixnums that **are** floatable without loss of precision, for instance 2^55:

      > (= (expt 2 55) (exact (inexact (expt 2 55))))
      #t

PR #619, opened 2022-04-17
Update all K&R function declarations to modern style

melted:


melted:
I made this a while back when I learned that K&R style functions will be removed from C23. Obviously they are never going to stop being supported by C compilers, but there's no upside to keeping them around and there is a risk that some modern tooling will drop support at some point.

burgerrg:
Thank you for this helpful update! I took the liberty of adding a commit to remove the PROTO macro from scheme.h, since it's no longer needed.

burgerrg:
Thank you for this helpful update! I took the liberty of adding a commit to remove the PROTO macro from scheme.h, since it's no longer needed.

PR #621, opened 2022-04-26
arm32le: only use cache sysconf constants if they are defined

nmeum:
This fixes compilation on musl libc which presently does not implement
these sysconf constants [\[0\]][0]. Please note though that this only fixes
compilation, there is still a separate run-time issue (see #622).

[0]: https://www.openwall.com/lists/musl/2017/07/24/1


cjfrisz:
This change looks pretty straightforward and passes the tests, though please update the ‘LOG’. For my own curiosity, I may test this on my Alpine Linux system once I manage to get it running.

PR #625, opened 2022-05-01
Don't rely on signed integer overflow in symhash

### LINKED ISSUE #622: arm32le: segfault in lookup_static() with musl libc
Hello, I am a contributor to the [Alpine Linux](https://alpinelinux.org) distribution (which uses [musl libc](https://musl.libc.org)). I am currently trying to improve our ChezScheme package. As part of this effort, I am trying to make the package available for more architectures (currently it is only supported on x86/x86_64). I started working on a port for armv6. While working on this, I ran into the following segfault during the build process on armv6:

```
Program received signal SIGSEGV, Segmentation fault.
0x4002d22c in lookup_static (s=0x4005d584 "(cs)lookup_foreign_entry") at foreign.c:101
101	    if (strcmp(s, (char *)&BVIT(Scar(Scar(p)),0)) == 0)
(gdb) bt
#0  lookup_static (s=0x4005d584 "(cs)lookup_foreign_entry") at foreign.c:101
#1  lookup (s=0x4005d584 "(cs)lookup_foreign_entry") at foreign.c:141
#2  0x4002d39c in Sforeign_symbol (s=0x4005d584 "(cs)lookup_foreign_entry",
    v=0x4002d304 <lookup_foreign_entry>) at foreign.c:170
#3  0x4002d6a0 in S_foreign_init () at foreign.c:323
#4  0x40035fe8 in main_init () at scheme.c:77
#5  Sbuild_heap (
    kernel=kernel@entry=0x3ffff951 "/var/tmp/abuild.XXXXEdCBcm/tmp/src/csv9.5.8/arm32le/bin/arm32le/scheme", custom_init=custom_init@entry=0x0) at scheme.c:1121
#6  0x40008168 in main (argc=1, argv=0x3ffff7b4) at main.c:292
```

To me, it looks like it segfaults when iterating through the `S_G.foreign_static` vector in `lookup_static` for the `(cs)lookup_foreign_entry` symbol:

https://github.com/cisco/ChezScheme/blob/e47faca3171eeeb5210ada0171d56df29e4e0fd6/c/foreign.c#L100-L102

Since `lookup_static` is called from `S_foreign_init` here I would have assumed the vector to be only filled with `Snil` but for some reason `Scar(p)` ends up creating a pointer which causes a segmentation fault on dereference. Since I am not an expert on the ChezScheme code base I am presently not sure what is causing this and was hoping someone with more advanced knowledge of the ChezScheme code might have an idea what could be causing the segfault (it is likely musl-related since I assume that arm32le is well tested in conjunction with glibc).

The full build log: [alpine-edge-chez-scheme-armhf.txt](https://github.com/cisco/ChezScheme/files/8566755/alpine-edge-chez-scheme-armhf.txt)

---

PS: I also had to slightly patch the code to address #527 and #621.

nmeum:
Without this patch, the loop body can easily cause an integer overflow. In this case, it is possible for the local variable `h` to take a negative value. The bitwise-and with `0x7fffffff` is supposed to handle this case by setting the sign-bit to zero. However, GCC may optimize out the bitwise-and if char is an unsigned value since it assumes that signed integer overflow does not take place and in this case arithmetic on `h` would only be performed on unsigned values. This is, for example, the case on the ARMv6 architecture. On ARMv6, symhash might thus return negative values, which can lead to a segmentation fault when they are used to index vectors etc.

This commit attempts to address this issue by rewriting the code in a way that it doesn't rely on undefined behavior (i.e. signed integer overflow). Alternatively, it might also be possible to cast the local variable `s` explicitly to a signed type to prevent the bitwise-and with `0x7fffffff` from being optimized out. However, such a change would still rely on undefined behavior.

Fixes #622


c/foreign.c
```diff
@@ -85,16 +85,16 @@ static ptr bvstring(const char *s) {
...
 /* multiplier weights each character, h = n factors in the length */
-static iptr symhash(const char *s) {
-  iptr n, h;
+static uptr symhash(const char *s) {
+  uptr n, h;
 
   h = n = strlen(s);
-  while (n--) h = h * multiplier + *s++;
-  return (h & 0x7fffffff) % buckets;
+  while (n--) h = h * multiplier + (unsigned char)*s++;
```
nmeum, line 92: Not what happens if `char` is signed and `*s` takes a value which cannot be represented in `unsigned char`.

Issue #627, opened 2022-05-08
Bug in char-


alewmoose:
```
;; should be -1
(char- #\a #\b) => 72057594037927935
```
I'm using Chez Scheme Version 9.5.8, machine type is ta6le

akeep:
Thanks for the bug report!  It looks like we have a logical shift right where we needed an arithmetic one.  I have a fix, but I need to add some tests for `char-`.  Apparently we do not currently have any tests for this (at least that I can find)!

PR #629, opened 2022-05-09
Fix to char-

akeep:
Small fix to #627 so that `char-` will return a negative number instead of a large positive integer when the second argument is `char>?` the first argument.

Also added a couple of tests for `char-` to the mats.


LOG
```diff
@@ -2309,3 +2309,8 @@
 - added install-docs target for make and corresponding configuration
   options: --installdoc, --installcsug, and --installreleasenotes
     configure makefiles/Makefile.in
+- fixed a bug in `char-` that lead to returning a large positive integer
```
burgerrg, line 2312: `lead` => `led`

burgerrg (commented):
Thanks for this fix! Please update the release notes as well.

akeep:
Thanks for having a look @burgerrg!  I've updated the release notes and fixed the log.  If it all looks good I'll squash and merge.

burgerrg (approved):
Thanks for fixing this bug!

PR #632, opened 2022-05-24
small repairs for old platforms

mflatt:
Three commits for small repairs backported from the Racket branch:

 * repair ppc32 code generation when a branch displacement is exactly 32,764 bytes
 * repair register handling — in a way that matters only for x86 — for a foreign callable float argument
 * repair too-narrow signatures for `multibyte->string` and `string->multibyte`

The commits don't include new tests. It must be possible to construct an expression that would try a 32,764-byte branch on ppc32, but it's not obvious and doesn't seem worth the trouble; there was such a branch generated by a Racket library at the time of the repair. For x86, existing tests could already fail and would sometimes fail in the Racket CI setup (but I don't know how to probe for x87 stack misuse more directly). For the signature changes, I think "primvars.ms" will use the new signature (but "primvars.ms" is ok with signatures that are too narrow); with the old signature, "windows.ms" tests could fail on Windows in the Racket branch due to the way cptypes uses the signature for optimization.

### DIFF
--- a/LOG
+++ b/LOG
@@ -2319,3 +2319,5 @@
     s/ppc32.ss
 - fix callable floating-point argument allocation on x86
     s/cpnanopass.ss
+- corrected signature of multibyte->string and string->multibyte
+    s/primdata.ss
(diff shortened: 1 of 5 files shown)


burgerrg:
Thank you for these repairs!

PR #634, opened 2022-06-02
Unicode 14.0

burgerrg:
unicode-charinfo.ss fails to build with the Unicode 14.0 files:
```
% make Scheme=scheme    
echo | scheme -q extract-info.ss
Exception in name->wbprop: unexpected property ZWJ
```


burgerrg:
@dybvig, do you know what needs to be done to support the ZWJ category?

dybvig:
Bob, I don't remember now much about the code, but what you've done to add support for ZWJ makes sense to me.

omasanori (approved):
LGTM. Thanks a lot!

PR #635, opened 2022-06-04
Propagate immutable-"" in cp0

gus-massa:
Also, propagate the immutable versions of `#()`, `#vu8()` and `#vfx()`.

And a similar change in cpnanopass, that includes also

    (define (eqok-help? obj)
      (or ...
          (eq? obj '#6=#6#)))


gus-massa:
Done

PR #649, opened 2022-07-23
Fix `Ssymbol_to_string` doc

lambdadog:
`Ssymbol_to_string` is a function, not a macro.

If it'd be better for me to combine this, #647, and #648 into one PR, let me know and I'll do so. I've just been authoring these as I've been running through and writing some bindings for the Chez Scheme kernel and noticing errors.


jltaylor-us:
Changes need an entry in `LOG`, and if each PR has its own edit to LOG then there will be merge conflicts between them.  So, yes, one PR (with an entry in LOG) would be good.

Thanks for the corrections!

lambdadog:
@jltaylor-us I see, give me just a moment.

PR #650, opened 2022-07-24
Fix foreign docs in several places

lambdadog:
Consolidation of #647, #648, and #649 + a `LOG` entry.

 - `Slocked_objectp` is now named correctly, as opposed to the previous incorrect `Sunlocked_objectp`.
 - Correctly marked `Sforeign_callable_{entry_point,code_object}` as macros, rather than functions.
 - Correctly marked `Ssymbol_to_string` as a function, rather than a macro.

### DIFF
--- a/LOG
+++ b/LOG
@@ -2336,5 +2336,6 @@
     c/foreign.c
 - Propagate immutable versions of "", #(), #vu8() and #vfx() in cp0.
     s/cp0.ss s/cpnanopass.ss
-- Fix several issues in foreign docs (section 4.8)
+- correct mislabeling of some functions as macros and vice versa in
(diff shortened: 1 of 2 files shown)


lambdadog:
Does this look correct, @jltaylor-us?

jltaylor-us:
_I_ would probably be a little more specific in the LOG, like

    fix some C functions mislabeled as macros and vice-versa,
    Slocked_objectp misspelled as Sunlocked_objectp

but the only real issue is the missing newline at the end of LOG. (Might as well keep the tools from whining about it.)

lambdadog:
I can absolutely be more specific.

jltaylor-us:
Thanks again!

PR #656, opened 2022-08-25
Fix configure arm64 error

Syzygies:
On arm64 (M1/M2 MacOS) ./configure throws nonfatal errors

./configure: line 338: [: =: unary operator expected
./configure: line 338: [: =: unary operator expected

because $m is not defined. This patch handles arm64, advising user that arm64 is not supported.

### DIFF
--- a/configure
+++ b/configure
@@ -133,6 +133,11 @@ case "${CONFIG_UNAME}" in
       m64=a6osx
       tm32=ti3osx
       tm64=ta6osx
+    else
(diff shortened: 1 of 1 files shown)


akeep:
Thanks for the update!  I think the right way to address the core issue with the script is to surround the references to `$m` with `"`.  Inspired by your message #657 also includes a message on how to build Chez Scheme for Rosetta2 while we continue to work on the arm64 port.

PR #657, opened 2022-08-26
fix missing quotes in configure, add message about building on arm64 OSX

jltaylor-us:


akeep:
Looks good, thanks for the update!

PR #660, opened 2022-09-06
Allow 0 arguments for `max` and `min`.

NoahStoryM:
See https://github.com/racket/racket/issues/4429 and https://github.com/racket/rhombus-prototype/issues/112.

`-inf.0` should be the identity element of `max`, and we can get more accurate results by ignoring the identity elements.

Before:
```scheme
Chez Scheme Version 9.5.8
Copyright 1984-2022 Cisco Systems, Inc.

> (max)
Exception: incorrect argument count in call (max)
> (min)
Exception: incorrect argument count in call (min)
> (max -inf.0 3)
3.0
> (min +inf.0 3)
3.0
> (flmax)
Exception: incorrect argument count in call (flmax)
> (flmin)
Exception: incorrect argument count in call (flmin)
```

After:
```scheme
Chez Scheme Version 9.5.8
Copyright 1984-2022 Cisco Systems, Inc.

> (max)
-inf.0
> (min)
+inf.0
> (max -inf.0 3)
3
> (min +inf.0 3)
3
> (flmax)
-inf.0
> (flmin)
+inf.0
```


burgerrg:
This is a breaking change for these procedures. I would instead recommend defining these variants in a library that you can then import when you want this behavior.

NoahStoryM:
As for allowing 0 arguments, I think there isn't compatibility issues.

The modified `max` will ignore the `-inf.0`s in the arguments, then return more accurate results (no loss of precision). I guess this change is advocated by r6rs:

> Note: If any argument is inexact, then the result is also inexact (unless the procedure can prove that the inaccuracy is not large enough to affect the result, which is possible only in unusual implementations). If min or max is used to compare number objects of mixed exactness, and the numerical value of the result cannot be represented as an inexact number object without loss of accuracy, then the procedure may raise an exception with condition type &implementation-restriction.

Of course if Chez Scheme thinks this change is not recommended, feel free to close this PR. :)

burgerrg:
Yes, I think your proposed changes would fit the R6RS specification. However, the changes would break compatibility with Chez Scheme's min & max procedures. Since you can define your own variants of these functions, I do not see a compelling reason to break compatibility.

PR #667, opened 2022-10-16
--installprefix requires an equal sign

gqqnbig:
If a space is used as the doc instructed, `configure` gives error "option '--installprefix' unrecognized or missing an argument"


burgerrg:
Thanks for the correction!

PR #685, opened 2023-03-01
Fix symbol-hashtable? docs in csug

geoffhill:
Looks like a copy-paste error from eq-hashtable? docs.


geoffhill:
FYI: Haven't actually tested rendering this.

burgerrg:

PR #686, opened 2023-03-19
Fix flags in signature of remove, member and assoc

gus-massa:
They can not be discarded in case they are comparing records with a custom equal procedure that has side effects.


gus-massa:
Something like this commit? I couldn't write a shorter version. (Note that if you test it in the current version of Chez Scheme, it fails for `#3%remove`,  but it's ok `#2%remove`.)

If you like the example, I can copy it to the mats of `assoc` and `member` and squash the commits.

jltaylor-us:
Seems like a reasonable test to me, and I can confirm that without the fix in this PR you get different results at optimize levels 2 and 3.

burgerrg:
Thanks, the short example tests whether remove calls the equality predicate. You don't need the box:
```
(let ()
  (define called #f)
  (define-record-type r)
  (record-type-equal-procedure (record-type-descriptor r)
   (lambda (x y eql?) (set! called #t) #t))
  (remove (make-r) (list (make-r)))
  called)
```
Please use this kind of example for assoc and member.
Please update the release notes by explaining that an optimization bug with these predicates was fixed when using custom record equality predicates with side effects in optimize-level 3.

gus-massa:
Done.

release_notes/release_notes.stex
```diff
@@ -1955,6 +1955,14 @@ in fasl files does not generally make sense.
 %-----------------------------------------------------------------------------
 \section{Bug Fixes}\label{section:bugfixes}
 
+\subsection{Optimization bug in \protect\scheme{remove},
+  \protect\scheme{member} and \protect\scheme{assoc}}
+
+When \scheme{remove}, \scheme{member} or \scheme{assoc} were used in
+\scheme{optimize-level} 3 in a position where the result was discarded,
+a bug in the source code optimizer could drop the call in spite first argument
+may be a records with a custom equality predicate with side effects.
```
jltaylor-us, line 1964:
```suggestion
a bug in the source code optimizer could drop the call even though the first argument
may be a record with a custom equality predicate that has side effects.
```

gus-massa: Thanks, fixed.

LOG
```diff
@@ -2350,3 +2350,7 @@
     mats/thread.ms
 - fix documentation of symbol-hashtable?
     csug/objects.stex
+- fix flags in signature of remove, member and assoc
+  they can not be discarded in case they are comparing records
+  with a custom equal procedure that has side effects.
+    s/primdata.ss
```
burgerrg, line 2356: add 5_2.ms and release_notes.stex

burgerrg (approved):

PR #690, opened 2023-03-31
fix a character comparison bug in S_intern_gensym

symingz:


cjfrisz:
Thank you for the pull request and making sure to document it in the `LOG`. The CI failure seems to be due to CI and nothing with your patch.

Do you happen to have a test case that exhibits the issue without your patch? It would be nice to add that to the `mats`.

symingz:
Added a test case.

cjfrisz:
Thanks for the patch!

PR #691, opened 2023-04-05
fix gensym->unique-string to leave symbol unchanged on failure

symingz:


burgerrg:
Thank you for finding and fixing this bug!

PR #693, opened 2023-04-12
fix live-pointer mask issue in np-expand-primitives

owaddell:


burgerrg (approved):
Thanks for your careful, detailed work to hunt down the problematic raw values!

owaddell:
Thanks for the quick review, Bob!

PR #699, opened 2023-05-23
repair for overflow checking and `raw` form

mflatt:
In 102b2c1082, the auto-generated case in `np-place-overflow-and-trap` for `raw` was not the correct treatment of the form.


jltaylor-us:
Can confirm that, at optimize-level 3, the expression added in the mats errors (specifically with `Exception: illegal instruction.  Some debugging context lost` on my system) on the code in `main` and works as expected with single line addition from this PR.

I'm not familiar with this bit of code, so I guess I'll give it a day or two to make sure nobody else wants to raise any objection or issue.

PR #703, opened 2023-07-06
follow xdg spec for history file on non-Windows

LiberalArtist:
Change the default location for the expeditor history file to `$XDG_STATE_HOME/chezscheme/history` on non-Windows systems, but fall back to `~/.chezscheme_history` when it already exists. Also, document the existing locations used on Windows (by default, `%APPDATA%\Chez Scheme\History`).

### DIFF
--- a/LOG
+++ b/LOG
@@ -2373,3 +2373,7 @@
 - fix repair to invalid live-pointer mask, where a pass did not handle the
   new raw-data form correctly
     mats/fx.ms s/cpnanopass.ss
+- change the default expeditor history file on non-Windows systems to
+  "$XDG_STATE_HOME/chezscheme/history" and document the existing Windows
+  default, "%APPDATA%\Chez Scheme\History"
+    scheme.1.in s/expeditor.ss csug/csug.bib csug/use.stex
--- a/csug/csug.bib
+++ b/csug/csug.bib
@@ -566,3 +566,16 @@ year = 2008}
    url = {https://doi.org/10.1145/263700.263733},
    year = {1997}
 }
+
+@techreport{xdg-base-directory,
+author = {Waldo Bastian and
+          Allison Karlitskaya and
+          Lennart Poettering and
+          Johannes L\"{o}thberg},
+title = {{XDG Base Directory Specification}, version 0.8},
+type = {{Cross-Desktop Group} specification},
+institution = {Freedesktop.org (X.Org Foundation)},
+url = {https://specifications.freedesktop.org/basedir-spec/0.8/},
+month = may,
+year = 2021
+}
--- a/csug/use.stex
+++ b/csug/use.stex
@@ -345,12 +345,34 @@ The expression editor does not run if the TERM environment variable is not
 set (on Unix-based systems), if the standard input or output files have
 been redirected, or if the \scheme{--eedisable} command-line option
 (Section~\ref{SECTUSECOMMANDLINE}) has been used.
-The history is saved across sessions, by default, in the file
-``.chezscheme\_history'' in the user's home directory.
+By default, the history is saved across sessions.
 The \scheme{--eehistory} command-line option
-(Section~\ref{SECTUSECOMMANDLINE}) can be used to specify a different
-location for the history file or to disable the saving and restoring of
-the history file.
+(Section~\ref{SECTUSECOMMANDLINE}) can be used to
+disable the saving and restoring of the history file or to specify an
+alternate location for the history file.
+The default location depends on the operating system:
+\begin{itemize}
+\item
+On Windows, the default file is
+``\scheme{%APPDATA%\Chez Scheme\History}'', where the value of
+\scheme{%APPDATA%} is typically
+``\scheme{C:\Users\\var{username}\AppData\Roaming}''.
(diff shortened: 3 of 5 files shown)


jltaylor-us:
... why?  What advantage does this added complexity have for Chez Scheme?  (Changing decades-old behavior just for adherence to some specification made up by four guys with apparently no community input does not sound like a persuasive argument to me.)  What problem is this trying to solve that can't already be solved by making an alias to invoke scheme with the `--eehistory` flag to put the file exactly where you want it?

LiberalArtist:
> What advantage does this added complexity have for Chez Scheme?

I will concede that the intended advantage of this change is for the users of Chez Scheme, rather than "for Chez Scheme" per se.

> (Changing decades-old behavior just for adherence to some specification made up by four guys with apparently no community input does not sound like a persuasive argument to me.)

I probably didn't do a particularly good job at introducing the XDG Base Directory Specification, because I would not characterize it this way. The spec dates from [at least 2003](https://web.archive.org/web/20170710174001/https://specifications.freedesktop.org/basedir-spec/basedir-spec-0.6.xml); it was extended in [2010](https://specifications.freedesktop.org/basedir-spec/0.7) and most recently in 2021 to define base directories for additional purposes. My perception is that it has been enthusiastically embraced by the community and widely adopted. To highlight just a few programs that also adjusted decades-old behavior:

  - GDB has used `$XDG_CONFIG_HOME/gdb/gdbinit` in preference to `~/.gdbinit` since [GDB 11.1 (2021)](https://lists.gnu.org/archive/html/info-gnu/2021-09/msg00007.html).
  - `curl` has [looked in `$XDG_CONFIG_HOME`](https://github.com/curl/curl/pull/5837) for its `curlrc` file since [`curl` 7.73.0 (2020)](https://curl.se/changes.html#7_73_0).
  - `less` has used `$XDG_CONFIG_HOME` and `$XDG_DATA_HOME` since [version 590 (2021)](https://www.greenwoodsoftware.com/less/news.590.html) and `$XDG_STATE_HOME` since [version 600 (2022)](https://www.greenwoodsoftware.com/less/news.600.html).
  - `nano` has used `$XDG_CONFIG_HOME` and `$XDG_DATA_HOME` since [`nano` 2.9.0 (2017)](https://www.nano-editor.org/news.php).
  - All of the tools from `util-linux` that support colorized output, such as `dmesg`, `fdisk`, and `hexdump`, have used `$XDG_CONFIG_HOME/terminal-colors.d` since [release 2.25 (2014)](https://git.kernel.org/pub/scm/utils/util-linux/util-linux.git/commit/?id=570b321).
  - [Emacs 27.1 (2020)](https://www.gnu.org/savannah-checkouts/gnu/emacs/news/NEWS.27.1) adopted `$XDG_CONFIG_HOME` with a particularly conservative compatibility strategy:
    ```
    Emacs can now use the XDG convention for init files.
    The 'XDG_CONFIG_HOME' environment variable (which defaults to
    "~/.config") specifies the XDG configuration parent directory.  Emacs
    checks for "init.el" and other configuration files inside the "emacs"
    subdirectory of 'XDG_CONFIG_HOME', i.e. "$XDG_CONFIG_HOME/emacs/init.el"

    However, Emacs will still initially look for init files in their
    traditional locations if "~/.emacs.d" or "~/.emacs" exist, even if
    "$XDG_CONFIG_HOME/emacs" also exists.  This means that you must delete
    or rename any existing "~/.emacs.d" and "~/.emacs" to enable use of
    the XDG directory.

    If "~/.emacs.d" does not exist, and Emacs has decided to use it
    (i.e. "$XDG_CONFIG_HOME/emacs" does not exist), Emacs will create it.
    Emacs will never create "$XDG_CONFIG_HOME/emacs".

    Whichever directory Emacs decides to use, it will set
    'user-emacs-directory' to point to it.
    ```

If some different strategy for compatibility&mdash;for example, preferring an existing `~/.chezscheme_history` even if `$XDG_STATE_HOME/chezscheme/history` also exists&mdash;seemed more acceptable, I wouldn't object to that.

> What problem is this trying to solve that can't already be solved by making an alias to invoke scheme with the `--eehistory` flag to put the file exactly where you want it?

It's a collective action problem: from the perspective of an individual tool, it's easy to argue that one should just use tool-specific configuration options, but much of the benefit as a user comes from being able to manage the scores of tools one has installed in a consistent manner. Probably everyone has their own perspective, but some of the advantages of the XDG Base Directories are:

  - Organizing the scores of dotfiles that otherwise proliferate in `~`;
  - A well-defined mechanism for changing default locations, suitable for individual users, sysadmins, distributions, and isolated temporary environments; and
  - Meaningfully grouping the different kinds of files tools generate, so that, for example, a user might check `$XDG_CONFIG_HOME` into version control, exclude `$XDG_CACHE_HOME` from backups, and share `$XDG_DATA_HOME`, but not `$XDG_STATE_HOME`, across multiple machines.

jltaylor-us:
I have polled the other maintainers, and of those who responded with any opinion at all, all were against.  So, thanks for your efforts, but we'll pass on this one.

If a downstream packager of Chez Scheme wants to implement the behavior described in the XDG Base Directory Specification, they can simply create a wrapper script to do so, using the long-established command line argument to specify the history file location.

Issue #273, opened 2018-03-01
build system doesn't support gcc 7 yet


cartazio:
```
/usr/local/bin/gcc-7  -m64 -Wpointer-arith -Wall -Wextra -Werror -O2 -I/opt/X11/include/  -c -DX86_64 -I../boot/a6osx -I../zlib schsig.c
schsig.c: In function 'handle_signal':
schsig.c:590:13: error: this statement may fall through [-Werror=implicit-fallthrough=]
             S_abnormal_exit();
             ^~~~~~~~~~~~~~~~~
schsig.c:592:9: note: here
         case SIGILL:
         ^~~~
schsig.c:594:13: error: this statement may fall through [-Werror=implicit-fallthrough=]
             S_error_reset("illegal instruction");
             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
schsig.c:595:9: note: here
         case SIGFPE:
         ^~~~
schsig.c:597:13: error: this statement may fall through [-Werror=implicit-fallthrough=]
             S_error_reset("arithmetic overflow");
             ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
schsig.c:599:9: note: here
         case SIGBUS:
         ^~~~
schsig.c:603:16: error: this statement may fall through [-Werror=implicit-fallthrough=]
             if (S_pants_down)
                ^
schsig.c:607:9: note: here
         default:
         ^~~~~~~
cc1: all warnings being treated as errors
make[2]: *** [schsig.o] Error 1

```
is the  relevant bit of errors. 

i triggered this with both recent clang (well, apple's current gcc aliased clang) and gcc 7

gwatt:
@cartazio Ah, I fixed that for linux. I didn't expand that to non-linux systems as older versions of gcc don't accept arbitrary `-W` flags.
https://github.com/cisco/ChezScheme/pull/183 if you want to adapt it for OS X.

PR #704, opened 2023-07-07
Prevent compiler warnings related to abnormal exit and fallthrough.

eric-s-raymond:
This PR was motivated by a failed attempt to build go-interpreter/chezgo, which sets -Werrors..

It fixes several trivial warnings.  There are a few others that look nontrivial, which I will investigate further.

No logic changes.


burgerrg (commented):
Thanks for this change. Please put the pacify comment on every unneeded break, and add a line to LOG.

eric-s-raymond:
Done.

With this change you're pretty close to compiling clean with -Werrors.  I recommend pushing it the rest of the way.

jltaylor-us:
Not that this should prevent us from merging this PR, but I thought we already built with `-Werror`...

On macOS, we use `-Wall -Wextra -Werror -Wno-implicit-fallthrough`, and from a quick glance it looks like the same in the linux makefiles.  This PR will allow us to remove the `-Wno-implicit-fallthrough` – I've verified that make in `a6osx/c` succeeds after removing the flag (on macOS, obviously); other builds would have to be verified (but maybe we can assume the GitHub runners will cover it).

It looks like go-interpreter/chezgo is using a copy of the code already configured for a6osx from 6 years ago, which is before the a6osx makefile was updated to add `-Wno-implicit-fallthrough` (see #273).

Are there any _other_ errors preventing a clean compile with `-Werror` with some other OS/compiler combination?

LOG
```diff
@@ -2373,3 +2373,5 @@
 - fix repair to invalid live-pointer mask, where a pass did not handle the
   new raw-data form correctly
     mats/fx.ms s/cpnanopass.ss
+- avoid sime trivial warnings related to abnormal exit and fallthrough.
```
burgerrg, line 2376: `sime trivial` => `some false-positive`

burgerrg (commented):
Thanks! Please remove the -Wno-implicit-fallthrough since it shouldn't be required any more.

burgerrg (approved):

Issue #700, opened 2023-06-06
Final stage in riscv64 and loongarch64 support


maoif:
Hello everyone,

I've been porting ChezScheme to riscv64 and loongarch64 (an architecture by Loongson Technology in China) and so far the majority of the tests are passed. However, there are still some tests that stuck and I don't have a clue why they exist. You can check out my latest code [here](https://github.com/maoif/ChezScheme-riscv64-loongarch64).

This is rv64le's mats report:
```
-------- o=0 --------
> cp0.mo:Bug in mat cp0-partial-folding clause 1 at line 1149, char 3 of ../cp0.ms
> cp0.mo:Bug in mat cp0-partial-folding clause 2 at line 1161, char 3 of ../cp0.ms
> cp0.mo:Bug in mat cp0-partial-folding clause 5 at line 1193, char 3 of ../cp0.ms
> cp0.mo:Bug in mat cp0-partial-folding clause 6 at line 1205, char 3 of ../cp0.ms
> cp0.mo:Bug in mat cp0-partial-folding clause 7 at line 1217, char 3 of ../cp0.ms
> cp0.mo:Bug in mat cp0-partial-folding clause 8 at line 1229, char 3 of ../cp0.ms
> cp0.mo:Bug in mat cp0-partial-folding clause 13 at line 1291, char 3 of ../cp0.ms
> cp0.mo:Bug in mat cp0-partial-folding clause 14 at line 1303, char 3 of ../cp0.ms
> cp0.mo:Bug in mat cp0-partial-folding clause 15 at line 1315, char 3 of ../cp0.ms
> cp0.mo:Bug in mat cp0-partial-folding clause 16 at line 1327, char 3 of ../cp0.ms

-------- o=3 --------
> cp0.mo:Bug in mat cp0-partial-folding clause 1 at line 1149, char 3 of ../cp0.ms
> cp0.mo:Bug in mat cp0-partial-folding clause 2 at line 1161, char 3 of ../cp0.ms
> cp0.mo:Bug in mat cp0-partial-folding clause 5 at line 1193, char 3 of ../cp0.ms
> cp0.mo:Bug in mat cp0-partial-folding clause 6 at line 1205, char 3 of ../cp0.ms
> cp0.mo:Bug in mat cp0-partial-folding clause 7 at line 1217, char 3 of ../cp0.ms
> cp0.mo:Bug in mat cp0-partial-folding clause 8 at line 1229, char 3 of ../cp0.ms
[...]

PR #706, opened 2023-07-13
Support for 64-bit RISC-V and LoongArch

maoif:
This PR adds support for riscv64 and loongarch64 architectures.
Only 64-bit Linux is supported right now.

RISC-V is the popular free and open-source ISA.
The current backend uses the G extension (M, A, F, D and Zifencei).

LoongArch is a new ISA developed by a Chinese company and is similar to RISC-V.

The machine names for the two architectures are `[t]rv64le` and `[t]la64le`.

I thank @mflatt for completing the FFI code when he merged the RISC-V code into Racket.
The two backends' FFI do not support varargs yet.

## Representation of `+nan.0`

There was one error (as noted [here](https://github.com/cisco/ChezScheme/issues/700)) that was
related to the different representations of `+nan.0` on different archs.

`(fl/ 0.0 0.0)` gives `+nan.0` on all archs. But on x86_64, the hex representation of the result
is `0xfff8000000000000`, yet on RISC-V and LoongArch the result is `0x7ff8000000000000`.

Both are correct. `0xfff8000000000000` is x86_64's QNaN Floating-Point Indefinite
(Intel® 64 and IA-32 Architectures Software Developer’s Manual, Vol. 1, 4-17)
and `0x7ff8000000000000` is the ordinary QNaN.

So it is not recommended to make the `[t]rv64le` and `[t]la64le` boot files on x86_64
because x86_64's QNaN will be baked into the boot file, which would lead to wrong
results on RISC-V and LoongArch in code like `(eqv? (fl/ 0.0 0.0) +nan.0)` (which should be `#t`).
Here `(fl/ 0.0 0.0)` will be computed by the instructions and give `0x7ff8000000000000`,
and `+nan.0` will be the baked-in (wrong) value of `0xfff8000000000000`.

## Test result

Ever since the RISC-V port was just bootstrapped, there were lots of errors whenthe tests were run. I tried hard to eliminate most of them. Some of them are duenot to the wrong code in the port, as detailed below.

### Test environemnt

The RISC-V backend is tested on qemu 8.0.2 and on a Sifive Unmatched.

The LoongArch backend is tested only on qemu 8.0.2 because, well, I don't have access to a LoongArch
machine.

### Tests that are influenced by the speed of execution

Since nowadays these two types of processors are not so fast, andsome tests are (I think) influenced by the speed of execution, soI bypassed them. They include:

- A `(current-time)` test in `5_1.ms`
- A cost-center test `mics.ms`

### One last probablistic error

There's still one error that sometimes appears and sometimes does not.

For RISC-V, the errors are:

```
> foreign.mo:Bug in mat foreign-ftype clause 76 at line 1921, char 3 of ../foreign.ms
> 5_8.mo:Bug in mat box-cas! clause 4 at line 41, char 3 of ../5_8.ms
> 5_6.mo:Bug in mat vector-cas! clause 4 at line 1265, char 3 of ../5_6.ms
```

For LoongArch, the error is:

```
> 5_6.mo:Bug in mat vector-cas! clause 4 at line 1265, char 3 of ../5_6.ms
```

The reason I call them *one* error is because they all have to do with atomics.
Specifically, the `foreign.mo` error compiles to `asm-lock`, and the `5_8.mo` and `5_6.mo` errors compile to `asm-cas`.
Both asm templates use the `lr.d` (load-reserved) and `sc.d` (store-conditional) instructions.

If I enter the respective test code directly in REPL, everything's fine.
Please help check if there's anything wrong with the backend.


maoif:
Anyone give me some feedback? 😅

cjfrisz:
Thanks for the PR! It will take some time to review the commits, though I know it was merged for Racket, which could help on this end.

We may also need to have a discussion about how to test these backends. Typically, whether it’s possible to run CI for a target determines whether it’s officially or unofficially supported.

mflatt:
The RISC-V half here (as authored by @maoif) is included in v9.9.9 merge 47daa9b34016de84fd111801d9d589d15a523abe.

burgerrg:
Please open a separate pull request if you'd like to add support for LoongArch.

maoif:
> Please open a separate pull request if you'd like to add support for LoongArch.

Sure, I'll be preparing for that.

PR #708, opened 2023-07-25
provide control over transcoded-port buffer sizes

owaddell:
This pull request adds two new parameters that provide control over the size of the string buffer and the internal codec bytevector buffer associated with transcoded ports. It also includes changes to reduce allocation and copying in certain cases for `bytevector->string`, `get-bytevector-all`, `get-bytevector-n`, `get-string-all`, and `get-string-n`.

One use case for the new parameters is to allow transcoded custom binary ports to take advantage of larger buffer sizes. Previously, such ports were effectively constrained by a fixed 1024-byte internal codec buffer even if created with a large `custom-port-buffer-size`.

Resolved questions:
- [X] Could / should we avoid allocating the `ioffsets` fxvector if the underlying `bp` does not support `port-position`? (Yes. Updated to avoid allocating `ioffsets` fxvector in this case.)
- [X] Should `make-codec-buffer` procedure take `bp` and `min-size` instead of just `bp`? (For now keeping `bp` as the only argument.)
  - con: For existing codecs it will always be 4, so why pass it in. (Perhaps it could be less for `latin-1-codec`, but that's seems pedantic.)
  - con: immediate 4 is compact in the code stream
  - con: if users know the minimum is 4 they could skip the check in cases where they know they have something of length >= 4 available.
  - pro: having a `min-size` argument is a reminder that there is in fact a minimum length
  - pro: avoids ugly occurrences of "four" in documentation and 4 in user code
- [X] Should I merge the two error checks on the result from `make-codec-buffer` into one? (Yes. Done.)


mats/io.ms
```diff
@@ -2135,6 +2144,166 @@
...
+                  "\x24B6;\x24B7;\x24B8;\x24B9;\x24BA;\x24BB;"
+                  ;; utf-8 4-byte
+                  "\x1F311;\x1F312;\x1F313;\x1F314;\x1F315;\x1F316;\x1F317;\x1F318;")
+                 tx)]
+               [min-codec-buffer-size 4]
+               [bp
+                (parameterize ([make-codec-buffer
+                                (lambda (bp)
+                                  (make-bytevector
+                                   (max min-codec-buffer-size codec-buffer-size)))])
```
jltaylor-us, line 2188: This function is never called, which is good since `codec-buffer-size` is unbound.  (It's spelled `codec-buf-size` here.)

Is this supposed to be part of the `parameterize` in the next binding in this `let*` (which actually does have a call to `transcoded-port` in its body)?

mats/io.ms
```diff
@@ -5006,3 +5192,98 @@
...
+  (let* ([test-str "\x1F311;\x1F312;\x1F313;\x1F314;\x1F315;\x1F316;\x1F317;\x1F318;"]
+         [bv (string->utf8 test-str)]
+         [str-len (string-length test-str)]
+         [bv-len (bytevector-length bv)]
+         [one-block-buffer-len (* 2 (+ str-len 1))]
+         [two-block-buffer-len (* 2 str-len)]
+         [tx (native-transcoder)])
+    (define (check-b->s buffer-size)
+      (parameterize ([file-buffer-size buffer-size])
+        (parameterize ([file-buffer-size buffer-size])
```
jltaylor-us, line 5279: duplicate line

csug/io.stex
```diff
@@ -420,6 +420,36 @@ Its initial value is the value of the \scheme{native-transcoder} procedure.
...
+
+\noindent
+\scheme{transcoded-port-buffer-size} is a parameter that determines the size
+of the string buffers associated with newly created transcoded ports.
+In addition to the string buffer, a transcoded input port maintains an fxvector
+of this size that aids \scheme{port-position}.
+When called with no arguments, the parameter returns the
+current buffer size.
+When called with a positive fixnum \var{k},
+it sets the current buffer size to \var{k}.
```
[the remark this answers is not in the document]
owaddell: I agree. Pushing a proposed fix as WIP commit at head.

mats/io.ms
owaddell: Indeed it is. Thanks for spotting this. Force pushing a fix for this and the next one.

mats/io.ms
owaddell: Thanks for spotting this.

jltaylor-us:
Looks good to me.

burgerrg (approved):

burgerrg (approved):

owaddell:
Thanks for the reviews, folks!

PR #710, opened 2023-07-30
Replace naive complex division by Algorithm 116

### LINKED ISSUE #709: (/ 1.0 0+inf.0i) produces NaN, which also affected result of atan and atanh
I was [investigating a bug in arctan implementation used in J](https://github.com/jsoftware/jsource/issues/180)

And I used Chez Scheme and Clozure CL for comparison, and find there is also a bug (but a different one) in Chez.

```
> (atan +inf.0+0.0i)
1.5707963267948966+nan.0i
> (atan +inf.0+1.2i)
1.5707963267948966+nan.0i
> (atanh 0+inf.0i)
+nan.0+1.5707963267948966i
> (atanh 0+1.0e18i)
0.0+1.5707963267948966i
```

That Clozure CL gives correct result

```
CL-USER> (atanh #C(0D0 1D++0))
#C(0.0D0 1.5707963267948966D0)
```

Where the issue itself is in https://github.com/cisco/ChezScheme/blob/8c5f5cc27cf8f2b432aafc4f2fc640435294b74a/s/5_3.ss#L556

since `clfatan` calls `clfatanh` to do the actual job.

I know Chez copied algorithm from [Kahan's paper](https://people.freebsd.org/~das/kahan86branch.pdf),
I checked and seems the bug is not from the original algorithm.

LdBeth:
Using Robert L. Smith's "Algorithm 116: Complex division", which fixes nan produced by original naive division method and potential float overflows.

fix #709

### DIFF
--- a/LOG
+++ b/LOG
@@ -2373,3 +2373,5 @@
 - fix repair to invalid live-pointer mask, where a pass did not handle the
   new raw-data form correctly
     mats/fx.ms s/cpnanopass.ss
+- fix cfl/ produces unwanted nan if denominator contains +/-infinity
+    s/library.ss
--- a/s/library.ss
+++ b/s/library.ss
@@ -204,7 +204,7 @@
           (fl- ($inexactnum-imag-part x) ($inexactnum-imag-part y)))]))
 
 (define-library-entry (cfl/ x y)
-   ;; spurious overflows, underflows, and division by zero
+   ;; Robert L. Smith, Algorithm 116
    (cond
       [(flonum? y)
        ;; a+bi/c => a/c + (b/c)i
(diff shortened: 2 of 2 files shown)


LdBeth:
Some test cases are added and this is ready for code review. Please let me know if there is more test case I can add or code styling issue.

burgerrg:
Thank you for finding this! Numerical Recipes in C uses the same algorithm because it avoids overflows, underflow, and loss of precision.

burgerrg:
@LdBeth, I made minor changes to the code, added a couple tests, and updated the documentation. Please let me know if you find any mistakes. Thanks!

jltaylor-us:
fwiw, Smith has (or is co-author on) an "improved" algorithm that handles even more edge cases, but at the cost of some additional complexity (and thus lower performance).  (I'm not suggesting that we need to implement this algorithm, just pointing out that it exists.)
https://arxiv.org/abs/1210.4539

LdBeth:
@burgerrg I have no problem with the changes. Thank you for the documentation update!

PR #711, opened 2023-08-02
minor documentation fixes

owaddell:


burgerrg (approved):
Thanks for finding and fixing these!

PR #714, opened 2023-08-22
repair for weak pairs and `collect` target generations

mflatt:
This is a backport of https://github.com/racket/ChezScheme/commit/e2cb8363a491692c55ef5739033ed94ec7c0de02.

When sweeping weak pairs and the max target generation is greater than the minimum target generation, then an updated car may need to be marked as dirty.


burgerrg:
Thank you very much for this fix!

PR #715, opened 2023-08-24
repair collector handling of ephemerons

mflatt:
When an ephemeron pair refers to an object in younger generation, and when a collection would move that object to a generation that's less the maximum target generation, the collector could sometimes mishandle the ephemeron pair by trying to register it multiple times for later handling.


burgerrg:
Thank you very much for fixing this!

PR #717, opened 2023-08-28
Add instructions on partial clone to README.md

jltaylor-us:


jltaylor-us:
I agree that structurally it would make more sense in BUILDING, but... I also think being jarring is part of the point, to keep people from needlessly cloning GB of repo.  I have a strong opinion either way.

burgerrg:
Thanks, Jamie! Matthew Flatt provided a similar update.

PR #728, opened 2023-10-13
Optimize _ pattern in syntax-case clauses

Jarhmander:
Plain `_` patterns in `syntax-case` occurs fairly frequently in code (more than 80 occurrences in Chez Scheme itself), and these patterns never fail. There are used like `else` clauses for `syntax-case`.

An easy optimization is to avoid calling `$syntax-dispatch` for such patterns and directly embed the `()` return value that would result for such a call. A similar optimization already existed for patterns that consists of a single identifier.

With that optimization, `cp0` can easily optimize out the last call to `syntax-error` that is never reached.

Before:
```
> (expand/optimize '(syntax-case x () (_ 'ok)))
(let ([tmp x])
  (let ([tmp (#3%$syntax-dispatch tmp '_)])
    (if tmp 'ok (#3%syntax-error tmp))))
```

After:
```
(expand/optimize '(syntax-case x () (_ 'ok)))
(begin x 'ok)
```

----
Note: there is something fishy with the GitHub workflows, and one random job would fail. It happened on the main in my fork, as well as with my branch.

### DIFF
--- a/LOG
+++ b/LOG
@@ -2455,3 +2455,5 @@
     boot/*/scheme.h c/externs.h c/number.c c/scheme.c csug/foreign.stex
     mats/foreign1.c mats/foreign.ms release_notes/release_notes.stex
     s/mkheader.ss
+- optimize simple _ patterns in syntax-case
+    s/syntax.ss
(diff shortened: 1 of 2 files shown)


burgerrg:
Thank you for this useful optimization!

PR #730, opened 2023-10-18
Add reboot pathway to CI

jryans:
As [discussed a while ago](https://github.com/racket/racket/issues/4601#issuecomment-1462903695), it would be good to have CI coverage of the reboot pathway.

This adds a Linux variant to CI which covers this additional bootstrapping dimension.


burgerrg:
Thanks for this helpful addition!

PR #731, opened 2023-10-18
update docs and tests for weak `cas!` and `lock!` operations

### LINKED ISSUE #720: Semantics of {vector, box}-cas! and ftype-lock!
I investigated the error related to `{vector, box}-cas!` and `ftype-lock!` in loongarch and riscv backend, and found they also exist in Racket's arm64 backend. I ran the following code:

```
(compile-interpret-simple #f)
;; Run the given forms n times,
;; stop if one of them returns #f
(define-syntax run-n
  (lambda (stx)
    (syntax-case stx ()
      [(_ n e* ...)
       #`(let loop ([count 0])
           (if (= n count)
               #t
               ;; run each form
               #,(let f ([e* #'(e* ...)])
                   (let ([e (car e*)] [rest (cdr e*)])
                     (if (null? rest)
                         #`(let ([res #,e])
                             (if res
                                 (loop (add1 count))
                                 (printf "error at ~s: ~s~n" count '#,e)))
                         #`(let ([res #,e])
                             (if res
                                 #,(f rest)
                                 (printf "error at ~s: ~s~n" count '#,e))))))))])))

(begin
  (define bx (box 1))
  (run-n 99999999
         (box-cas! bx 1 4)
         (box-cas! bx 4 1)))
```

The seesion on loongarch is something like:

```
> (run-n 9999999
         (ftype-init-lock! A (g *) x)
         (ftype-lock! A (g *) x)
         (not (ftype-lock! A (g *) x))
         (eq? (ftype-unlock! A (g *) x) (void)))
error at 110315: (ftype-lock! A (g *) x)
> (run-n 9999999
         (ftype-init-lock! A (g *) x)
         (ftype-lock! A (g *) x)
         (not (ftype-lock! A (g *) x))
         (eq? (ftype-unlock! A (g *) x) (void)))
error at 1245134: (ftype-lock! A (g *) x)
> (run-n 9999999
         (ftype-init-lock! A (g *) x)
         (ftype-lock! A (g *) x)
         (not (ftype-lock! A (g *) x))
         (eq? (ftype-unlock! A (g *) x) (void)))
error at 501949: (ftype-lock! A (g *) x)
> (run-n 9999999
         (ftype-init-lock! A (g *) x)
         (ftype-lock! A (g *) x)
         (not (ftype-lock! A (g *) x))
         (eq? (ftype-unlock! A (g *) x) (void)))
error at 26899: (ftype-lock! A (g *) x)
> (run-n 9999999
         (ftype-init-lock! A (g *) x)
         (ftype-lock! A (g *) x)
         (not (ftype-lock! A (g *) x))
         (eq? (ftype-unlock! A (g *) x) (void)))
error at 533811: (ftype-lock! A (g *) x)
```

If we decrease the number of times, we may succeed:

```
> (run-n 99999
         (ftype-init-lock! A (g *) x)
         (ftype-lock! A (g *) x)
         (not (ftype-lock! A (g *) x))
         (eq? (ftype-unlock! A (g *) x) (void)))
#t
> (run-n 99999
         (ftype-init-lock! A (g *) x)
         (ftype-lock! A (g *) x)
         (not (ftype-lock! A (g *) x))
         (eq? (ftype-unlock! A (g *) x) (void)))
#t
> (run-n 99999
           (ftype-init-lock! A (g *) x)
           (ftype-lock! A (g *) x)
           (not (ftype-lock! A (g *) x))
           (eq? (ftype-unlock! A (g *) x) (void)))
error at 27123: (ftype-lock! A (g *) x)
```

The situation is similar on other two architectures.

The generated machine code under optimization level 3 for `(box-cas! bx 1 4)` on riscv is:

```
lr.d           %scratch0, %x28, ()   # load-reserved from the box
bne            %scratch0, %x25, 16   # compare with old value
sc.d           %scratch1, %x28, %x26 # store-conditional
sltiu          %cond, %scratch1, 1   # if %scratch1 == 0, then write 1 into %cond
jal            %real-zero, 8
xor            %cond, %cond, %cond
```

Offsets are in bytes and all instructions are 4 bytes wide.

The code is run is a single thread, how can it fail? I checked some sites ([1](https://www.appsloveworld.com/cplus/100/159/c11-how-to-produce-spurious-failures-upon-compare-exchange-weak) [2](https://stackoverflow.com/questions/25199838/understanding-stdatomiccompare-exchange-weak-in-c11) and [3](https://en.cppreference.com/w/cpp/atomic/atomic/compare_exchange)), and found that the above phenomenon is called the "spurious failure" of weak compare-and-swap, which is how our `cas!` is implemented. So even we run the code in a single thread and the given value is equal to the boxed value, and no other cores have touched the addresses reserved by the paired load-reserved, store-conditional will still fail.

For weak CAS, if the store-conditional fails, it will just quit and the set the status flag. In the strong CAS, if the store-conditional fails, it will loop back to the load-reserved/linked, and retry, until the store-conditional succeeds. In x86, CAS is implemented using a single `locked-cmpxchg`, hence no spurious failure.

In the mats tests `5_6.ms line 1268` and `5_8.ms line 41`, the two tests fail/succeed randomly. I suppose the tests assume that the `cas!` is strong, but the implementation is weak (also in arm32, powerpc and Racket's arm64).

In addition to `{vector, box}-cas!`, the test `ftype.ms line 2790` and `2796` also assume that `ftype-lock!` acquires the lock "strongly", but in fact the implementation (using store-conditional without retry loop) is "weak". Yet if we change the `asm-lock!` implementation to contain a retry loop, we are inventing another `ftype-spin-lock!`.

So maybe we should 1) make the CAS assembler primitives strong by adding loops, and 2) assume that `ftype-lock!` is weak and modify the tests.

Any other nice ideas?

mflatt:
This PR is intended to close #720.

The `vector-cas!`, `box-cas!`, and `ftype-lock!` operations are "weak" in the sense that they may spuriously fail on an architecture (like AArch64) with a weak memory model. Also, those operation do not necessarily imply any overall memory ordering, so using `memory-order-acquire` and `memory-order-release` may be needed. The documentation did not make this clear, and some tests did not handle spurious failure.

As noted in the discussion for #720, a good case could be made that these operations should be "strong", which could mean both avoiding spurious failure and ensuring more memory ordering. The approach here is conservative: it keeps the current implementation and allows access to the underlying machine's behavior, but it's also compatible with the possibility that we decide to strengthen guarantees in the future.


mats/mat.ss
```diff
@@ -615,3 +615,20 @@
...
+            (memq (substring mt 0 3) '("ta6" "ti3")))
+        ;; no retry loop needed on x86
+        (lambda (stx)
+          (syntax-case stx ()
+            [(_ e) #'e]))
+        ;; add retry loop
+        (lambda (stx)
+          (syntax-case stx ()
+            [(_ e) #'(let loop ([n 10])
+                       ;; 10 sprious failures in a row is vanishingly unlikely?
```
burgerrg, line 631: sprious => spurious

csug/objects.stex
```diff
@@ -820,9 +820,16 @@ If the \var{n}th element of \var{vector} that would be replaced
 is not \scheme{eq?} to \var{old-obj}, then
 \var{vector} is unchanged.
 
+On an architecture with a weak memory model, \scheme{vector-cas!} can
+``spuriously'' fail, leaving \var{vector} unchanged and returning
```
burgerrg, line 824: ``spuriously'' => spuriously since you're not quoting anything here.

csug/objects.stex
```diff
@@ -1852,9 +1859,16 @@ if the replaced content is \scheme{eq?} to \var{old-obj}.
 If the content of \var{box} that would be replaced is not \scheme{eq?} to \var{old-obj}, then
 \var{box} is unchanged.
 
+On an architecture with a weak memory model, \scheme{box-cas!} can
+``spuriously'' fail, leaving \var{box} unchanged and returning
```
burgerrg, line 1863: ``spuriously'' => spuriously

csug/threads.stex
```diff
@@ -414,9 +417,13 @@ to the use of any of the other operators; if this is not done, the
 behavior of the other operators is undefined.
 
 \scheme{ftype-lock!} can be used to lock the lock.
-If it finds the lock unlocked at the time of the operation, it locks
+If it finds the lock unlocked at the time of the operation, it (normally) locks
 the lock and returns \scheme{#t}; if it finds the lock already locked,
-it returns \scheme{#f} without changing the lock.
+it returns \scheme{#f} without changing the lock. On an architecture with a weak memory model,
+\scheme{ftype-lock!} can ``spuriously'' fail, leaving a lock unchanged and returning
```
burgerrg, line 423: ``spuriously'' => spuriously

burgerrg (changes requested):

burgerrg:
One of the tests triggered the error near line 1176 of mats/misc.ms. The call `(error 'compute-size-increments` should be `(errorf 'compute-size-increments`

mflatt:
> One of the tests triggered the error near line 1176 of mats/misc.ms. The call (error 'compute-size-increments should be (errorf 'compute-size-increments

Thanks - I'll work on that in another PR in the near future, since it's unrelated to the change here.

burgerrg:

PR #734, opened 2023-10-19
fix record? primitive at optimize-level 3

owaddell:


burgerrg (approved):
Thanks for this bug fix!

owaddell:
Thanks for the review, Bob!

PR #737, opened 2023-10-20
repair tests for cmg=1

mflatt:
In attempting to improve some tests, commit a4654e4ee lost two adjustments to `collect-maximum-generation` to ensure that tests work when the parameter is set to 1 (as in some modes that are tried by `make test-more`).


burgerrg (approved):
Thanks for finding and fixing this test bug!

PR #738, opened 2023-10-21
restore simpler Windows CI

mflatt:
Mostly reverts 01b5bb60cc, but also ~~changes "/c" to "//c"~~ uses `MSYS_NO_PATHCONV=1` --- which seems to be the right repair to the original problem based on the way Windows `bash` variants treat arguments that start with `/`.

### DIFF
--- a/.github/workflows/build.sh
+++ b/.github/workflows/build.sh
@@ -2,8 +2,7 @@
 set -e -o pipefail
 export ZUO_JOBS="$(getconf _NPROCESSORS_ONLN)"
 if test "$TOOLCHAIN" = vs ; then
-    # cmd.exe /c "build.bat $TARGET_MACHINE"
-    echo assuming built previously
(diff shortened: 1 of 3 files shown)


.github/workflows/build.sh
```diff
@@ -2,8 +2,7 @@
 set -e -o pipefail
 export ZUO_JOBS="$(getconf _NPROCESSORS_ONLN)"
 if test "$TOOLCHAIN" = vs ; then
-    # cmd.exe /c "build.bat $TARGET_MACHINE"
-    echo assuming built previously
+    cmd.exe //c "build.bat $TARGET_MACHINE"
```
burgerrg, line 5: Try `MSYS_NO_PATHCONV=1 cmd.exe /c build.bat $TARGET_MACHINE`

mflatt: Thanks — I see how that can be better. Change pushed.

PR #740, opened 2023-10-22
repair `char-foldcase` for some single-character special cases

mflatt:
For example, `(char-foldcase #\x1E9E)` should be `#\xDF`, in contrast to the `(string-foldcase (string #\x1E9E))` result of "ss". The character foldings that that are different here correspond to cases that are "S" status in "CaseFolding.txt".

Thanks to Marc Feeley for noticing and reporting the bug.

Also, repair some dependencies in "s/build.zuo".


mats/5_4.ms
```diff
@@ -590,6 +590,9 @@
     (error? (char-foldcase 'a))
     (eqv? (char-foldcase #\a) #\a)
     (eqv? (char-foldcase #\Z) #\z)
+    (eqv? (char-foldcase #\Z) #\z)
```
burgerrg, line 593: The `#\Z` case is a duplicate.

burgerrg (changes requested):

mats/5_4.ms
mflatt: Fixed in a new push - thanks!

burgerrg (approved):

PR #744, opened 2023-10-30
Update BUILDING -- fixed small typo

cel7t:
There is a small typo in the document, where existing is misspelled as exsisting.


burgerrg:

PR #748, opened 2023-11-02
Update revision file generation

burgerrg:
- In version 9, the revision file has a second line naming the revision control system. This pull request adds "git" in the second line.
- The build process now properly identifies the .git file when used as a submodule.
- BUILDING is updated to reflect the unconditional inclusion of boot/pb.


mflatt:
This generally looks good to me.

The reason for the old loop to search for `.git` in ancestor directories (instead of assuming `..`) was to work when the `ChezScheme` directory is vendored into another repo, such as `racket/racket`. I don't feel strongly about this, though, and the `racket/racket` vendored `ChezScheme` could just have a difference here.

burgerrg:
OK, I think the loop does the same logic that `git describe` uses internally, so see what you think now.

burgerrg:
@mflatt, is it possible for c/build.zuo to catch an error if `git describe` fails and then write the literal string? That would avoid the loop entirely.

mflatt:
@burgerrg The revised loop looks right to me.

The `process-status` function gets the exit code for a completed process, so that could be used to detect whether a `git` command succeeded. I prefer not trying to run `git` if the source does not appear to be in a Git repository, but I can see either choice.

PR #750, opened 2023-11-06
fix dependencies for cross compilation

mflatt:
Changes in the "unicode" directory were not tracked correctly. The missing dependency affects rebuilding via pb, for example.


burgerrg:

PR #752, opened 2023-11-07
add $hashtable-cells that skips immutable check

owaddell:
Provide a way to work around the 10.x breaking change in `hashtable-cells`.


burgerrg (approved):

owaddell:
Thanks for the review, Bob!

PR #753, opened 2023-11-13
Make (load-shared-object #f) portable

Jarhmander:
_Am I late to the Chez Scheme 10 party?_

This is an old PR that's now reworked. Basically, an non-documented feature of `load-shared-object` is that it accepts `#f`. On UNIX it does something useful, but on Windows it just makes an invalid memory reference. 

The behavior on UNIX systems is to make the exported symbols of the executable as well as all shared objects linked to it available as foreign entries (what essentially `dlopen(NULL, ...)` does). It's very useful to get standard C library functions in a portable way; moreover, when creating a custom executable based on Chez Scheme, all dynamic libraries linked to the executable are readily available with `(load-shared-object #f)` to have access to the linked foreign entries. It helps decoupling the shared objects from their paths.

### DIFF
--- a/c/externs.h
+++ b/c/externs.h
@@ -445,6 +445,7 @@ extern ptr *S_get_call_arena(ptr tc);
 /* windows.c */
 extern INT S_getpagesize(void);
 extern ptr S_LastErrorString(void);
+extern HMODULE *S_enum_process_modules(void);
 extern void *S_ntdlopen(const char *path);
 extern void *S_ntdlsym(void *h, const char *s);
 extern ptr S_ntdlerror(void);
--- a/c/foreign.c
+++ b/c/foreign.c
@@ -66,6 +66,9 @@ static ptr bvstring(const char *s);
 
 #ifdef LOAD_SHARED_OBJECT
 static void load_shared_object(const char *path);
+#ifdef WIN32
+HMODULE *base_modules;
+#endif /* WIN32 */
 #endif /* LOAD_SHARED_OBJECT */
 
 #ifdef HPUX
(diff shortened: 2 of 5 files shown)


mflatt:
Using `#f` on Unix platforms would also find symbols in shared libraries loaded after startup (particularly since `RTLD_LOCAL` is lot specified with `dlopen`), but it looks like the implementation here would only check DLLs loaded at startup on Windows. Would it be better to enumerate modules on each use of `#f`? (FWIW, that's what Racket does.)

For calling `EnumProcessModules`, maybe avoid a hard limit by allocating the array and resizing if the actual number of needed byes is larger than a hardwired limit like 256?

In the documentation, I think it would be better to say `#f` works for getting C-library functions *if* Chez Scheme is dynamically linked, instead of making the linking strategy part of Chez Scheme's specification.

Jarhmander:
> Using `#f` on Unix platforms would also find symbols in shared libraries loaded after startup (particularly since `RTLD_LOCAL` is lot specified with `dlopen`), but it looks like the implementation here would only check DLLs loaded at startup on Windows. Would it be better to enumerate modules on each use of `#f?` (FWIW, that's what Racket does.)

Yes, the way I described it was as if I tried to replicate `dlopen(NULL, ...)`, but what I documented in CSUG is deliberately vague, and what I implemented is technically consistent with the documentation. That being said, I'm quite OK to change the code to enumerate all modules each time a NULL handle is seen in the "foreign dynamic list", if it's really useful. I guess it could be, if some external C code uses `LoadLibrary` and we would like Chez Scheme to snoop on those libraries too.

(Also, if it's done on Racket, why it's not in Chez?)

> In the documentation, I think it would be better to say #f works for getting C-library functions if Chez Scheme is dynamically linked, instead of making the linking strategy part of Chez Scheme's specification.

My initial PR had such provisions, but I removed them from the text because I thought it was unusual and perhaps would make the text uselessly heavier. I guess I was right in the beginning and I should not have changed my mind after all!

Updated. I can squash a few commits if you want.

jltaylor-us (approved):

PR #755, opened 2023-11-14
find boot files relative to executable by default

mflatt:
This PR builds on #732 by changing the default boot-file search path to
    
```
     "%x:%x/../lib/csv%v/%m:%x/../../boot/%m"
```

which searches in the executable's directory, through a relative path that works post-installation, and through a relative path that works within the workarea. (More precisely, the middle path in the default takes into account the configured bin and lib target directories.)
    
After the change, it's much easier to run a just-built executable before installing. Also, installing produces something that you can move around in the filesystem.
    
As a further step to make things work in the default way, `configure` is changed to recognize `--prefix` as an alias for `--installprefix`. The `--installabsolute` flag uses the old default search path, which is absolute relative to the target installation directory.

Potential discussion points:

 * I'm not sure about putting `%x` as the first search path. The intent is to be consistent with Windows and to make it easy for someone to gather files into a single directory and still have `scheme` run.

 * Instead of changing the default configuration while adding `--installabsolute` to get the old bevahior, the new behavior could be opt-in with `--installabs`. I think the new behavior is easier to work with, but the extra path operations and searching aren't useful or unnecessary in a conventional installation.

 * The `main` program communicates `argv[0]` to `%x` resolution through a new `Sregister_argv0` function. That function isn't needed on platforms and configurations where the process executable's path can be discovered, but using `argv[0]` supports a general and portable fallback.


c/self-exe.c
```diff
@@ -0,0 +1,234 @@
...
+  RETURN VALUE
+    On success, `S_get_process_executable_path()` returns a pointer
+    to the resolved path string. Otherwise, it returns `NULL`.
+
+  NOTES
+    If `SELF_EXE_MAIN` is defined, a `main()` is defined to call and
+    print the result from `S_get_process_executable_path`, which is
+    useful for testing.
+
+  Parts of the implementaiton here are from the LLVM Project under
```
jltaylor-us, line 25:
```suggestion
  Parts of the implementation here are from the LLVM Project under
```

csug/foreign.stex
```diff
@@ -3103,6 +3104,10 @@ values of \var{v} and off when \var{v} is zero.
 In verbose mode, the system displays a trace of the search process
 for subsequently registered boot files.
 
+\scheme{Sregister_argv0} registers \scheme{argv[0]} as supplied to an
+executable, which needed on some platforms to search for boot files
+relative to the executable where the search path uses ``\scheme{%x}''.
```
jltaylor-us, line 3109:
```suggestion
executable, which is needed on some platforms to search for boot files
relative to the executable when the search path uses ``\scheme{%x}''.
```

c/self-exe.c
mflatt: Fixed in #732

csug/foreign.stex
mflatt: Fixed - thanks!

ur4t:
> * I'm not sure about putting `%x` as the first search path. The intent is to be consistent with Windows and to make it easy for someone to gather files into a single directory and still have `scheme` run.
> * Instead of changing the default configuration while adding `--installabsolute` to get the old bevahior, the new behavior could be opt-in with `--installabs`. I think the new behavior is easier to work with, but the extra path operations and searching aren't useful or unnecessary in a conventional installation.

Consistency across platforms is great. Maybe we can go further to uniform `DEFAULT_HEAP_PATH` with `"%x" SEARCHPATHSEP "%x" PATHSEP ".." PATHSEP "lib" PATHSEP "csv%v" PATHSEP "%m" SEARCHPATHSEP "%x" PATHSEP ".." PATHSEP ".." PATHSEP "boot" PATHSEP "%m"` in `scheme.c`.

> * The `main` program communicates `argv[0]` to `%x` resolution through a new `Sregister_argv0` function. That function isn't needed on platforms and configurations where the process executable's path can be discovered, but using `argv[0]` supports a general and portable fallback.

With `Sregister_argv0()` and a global `process_argv0`, argument `execpath` in `find_boot()`, `next_path()` and `Sbuild_heap()` is redundant, and related logic in `main()` can be improved as well.

Changing signature of `Sbuild_heap()` will introduce a API breakage, but it is also not a good idea for two APIs to affect the same global state in different ways.

mflatt:
> Consistency across platforms is great. Maybe we can go further to uniform ...

That path is not quite right in general, though, because the actual middle search path depends on the relative path from `InstallBin` to `InstallLib`. Also I don't think it's common arrange files in that way on Windows. So, I doubtful that we should go further in that direction.

> it is also not a good idea for two APIs to affect the same global state in different ways.

On reflection, `Sregister_argv0()` seems like a bad idea, and I was using `process_argv0` in unnecessary places. I now think it's better to add `Sregister_boot_executable_relative_file()` as a variant of `Sregister_boot_file` to be preferred when the boot file path might be resolved relative to the executable. That way, `execpath` is consistently passed as an argument when it is useful, including for `Sbuild_heap`. I rewrote the documentation on the `Sregister_` functions, and I renamed and added documentation for a variant that I had added earlier but not documented.

csug/foreign.stex
```diff
@@ -3093,6 +3093,8 @@ program.
 \cfunction{void}{Sscheme_init}{void (*\var{abnormal}_\var{exit})(void)}
 \cfunction{void}{Sset_verbose}{int \var{v}}
 \cfunction{void}{Sregister_boot_file}{const char *\var{name}}
+\cfunctTwo{void}{Sregister_boot_executable_relative_file}{const char *\var{execpath},}{const char *\var{name}}
+\cfunction{void}{Sregister_boot_relative_file}{const char *\var{name}}
 \cfunction{void}{Sregister_boot_file_fd}{const char *\var{name}, int \var{fd}}
 \cfunctTwo{void}{Sregister_boot_file_fd_segment}{const char *\var{name}, int \var{fd},}{iptr \var{offset}, iptr \var{len}, int \var{close}_\var{after}}
```
ur4t, line 3099: `Sregister_boot_file_fd_segment()` is actually `Sregister_boot_file_fd_region()` in other places.

And `check_boot_file_state("Sregister_boot_file_fd");` in body of `Sregister_boot_file_fd_region()` might be refined as `check_boot_file_state("Sregister_boot_file_fd_region");`.

mflatt: Thanks! Fixed.

ur4t (approved):

PR #758, opened 2023-11-17
adjust timing tests

mflatt:
For tests that involve time: use CPU time whenever possible, retry tests that unavoidably involve real time, and and scale an expected upper limit on CPU time by timing a baseline calculation.

Also, scale down a test that pushes memory use to 1.7GB on a 64-bit machine, bringing peak memory down below 500MB.


jltaylor-us (approved):

mflatt:
An issue with switching to CPU time: less granularity on Windows, and various tests go wrong by measuring 0 for computations that take too little time. I'll try again.

PR #759, opened 2023-11-19
repair for `system` in thread that can be deactivated

mflatt:
The `system` primitive is intended to be like a `__collect_safe` foreign call, allowing other threads to perform a collection while waiting for the shell command to return. Its shell-command argument is allocated, however, so `system` needs a copy of its argument while its thread is deactivated.

### DIFF
--- a/c/prim5.c
+++ b/c/prim5.c
@@ -711,16 +711,30 @@ static void s_showalloc(IBOOL show_dump, const char *outfn) {
 
 static ptr s_system(const char *s) {
   INT status;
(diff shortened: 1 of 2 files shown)


jltaylor-us (commented):
I suppose this probably _should_ check that `malloc` was able to allocate memory... presumably with error handling similar to that is `S_malloc` but with a different error message.

mflatt:
Thanks — `malloc` check added.

jltaylor-us (approved):

PR #760, opened 2023-11-19
avoid some C functions that are not in C99

mflatt:
Commit ff97b9003aa065c4ac2b1bf09f4876ec654ece9b introduces "self-exe.c" using C99-and-later forms, which were not previously used in Chez Scheme. But it also uses `strdup` and `strtok_r`, which are widely available but not C99. Those functions turn out not to be available on some old environments that are still relevant to Racket, for example.

It always seems dumb to write yet another alternative to `strdup` and `strtok`, but the trade-off here seems worthwhile to me.

### DIFF
--- a/c/self-exe.c
+++ b/c/self-exe.c
@@ -61,6 +61,9 @@ static char *wide_to_utf8(const wchar_t *arg) {
     return NULL;
   }
   char *arg8 = (char *)malloc(len * sizeof(char));
+  if (arg8 == NULL) {
+    return NULL;
+  }
   if (0 == WideCharToMultiByte(CP_UTF8, 0, arg, -1, arg8, len, NULL, NULL)) {
     free(arg8);
     return NULL;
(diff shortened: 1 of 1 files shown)


ur4t:
How about using implementations from musl libc?

- http://git.etalabs.net/cgit/musl/tree/src/string/strdup.c
- http://git.etalabs.net/cgit/musl/tree/src/string/strtok_r.c

And it might reduce a bit of size to define those helpers only if they are not available.

jltaylor-us (approved):

mflatt:
Ok, switched to the musl implementations. I did not try to figure out detecting when `strcpy` and/or `strtok_r` are available.

jltaylor-us:
Why would we want to borrow code with _yet another_ license and have to figure out what our obligations are under that license when you already had perfectly good code that does exactly what you want?

mflatt:
The most recent push has the old-but-improved code.

I also looked for calls to `malloc` that need a check for a `NULL` result and added checks. I'm not sure that's really useful when allocating something the size that `_NSGetExecutablePath` reports, say, since it seems like a catastrophic situation. But I might be wrong, and uniformity seems like the best policy.

ur4t (approved):

PR #768, opened 2023-11-23
Correct typos and reformat code

maoif:
Corrected some typos,
Changed many tabs to 4-whitespaces,
Reformatted some backend code.

Some changes might be unnecessary. Take them as you see fit.

### DIFF
--- a/IMPLEMENTATION.md
+++ b/IMPLEMENTATION.md
@@ -350,7 +350,7 @@ type word indicates a vector. The fixnum value is the vector's length
 in words/objects, but shifted up by 1 bit, and then the low bit is set
 to 1 for an immutable vector.
 
-Most kinds of Scheme values are represented records, so the layout is
+Most kinds of Scheme values are represented as records, so the layout is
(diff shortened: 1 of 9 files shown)


mflatt:
Overall, I'm ok with reformatting to remove tabs, but it does complicate tools like `git blame`, so I would normally leave it alone. Let's see what others think.

There were a few places where tabs has been replaced by 4 spaces but needed 8, so I pushed a small revision here.

samth:
You can use `-w` with `git blame` to ignore whitespace changes. It's also possible to create a repository-level `config` file that notes particular revisions to ignore for the purposes of `git blame`.

shenlebantongying:
> It's also possible to create a repository-level config file that notes particular revisions to ignore for the purposes of git blame.

There is `.git-blame-ignore-revs` for this purpose. Github support it while local git need one line config. https://docs.github.com/en/repositories/working-with-files/using-files/viewing-a-file#ignore-commits-in-the-blame-view

Issue #263, opened 2018-01-18
ChezScheme startup time


xfbs:
Chez scheme is a real pleasure to use, I'm especially fond of the expression editor on the command line for some interactive coding. But I have one question — I understand that chez is known for being quite efficient, why does it take so long to start up?

```
$ echo "(exit)" | time chez -q
chez -q  0.17s user 0.04s system 54% cpu 0.383 total
$ time ruby -e 'exit'
ruby -e 'exit'  0.08s user 0.02s system 93% cpu 0.108 total
$ time python3 -c 'exit()'
python3 -c 'exit()'  0.05s user 0.01s system 76% cpu 0.083 total
```

Is this possibly a bug?

atticus0:
You are comparing the python and ruby interpreter with the chez scheme compiler. Interpreters have generally a faster startup time. Try it with the chez scheme interpreter :  `time petite -q`.

akeep:
It is worth noting that the ``scheme`` and ``petite`` are actually the same binary, and both do (roughly) the same thing at startup.

When you start ``scheme`` (or ``petite``), after looking through the command line arguments and determining where the boot file(s) are, it loads the boot file(s), then runs a maximum generation collection into the static generation, since the core library is not touched during normal garbage collection.

Loading the boot files consists of uncompressing, reading the fasl-format binary, and executing the expressions in the boot file, which set-up the top-level environment that Chez Scheme programs (including the REPL) execute from.

In the case of ``scheme`` this loads both the ``petite.boot`` which contains the core library and interpreter and ``scheme.boot`` which contains the compiler.  In the case of ``petite`` it only loads ``petite.boot``, and this does load faster, since it both executes less code and requires less time to collect, since there is less in the Scheme heap when it is called.

So, that is what it is doing during that first ~0.10s to ~0.20s.  Could it be faster?  Possibly.  We could break the boot file into smaller pieces.  For instance, the ``petite.boot`` file contains things like the expression editor code and the interpreter, which you do not need if you are just executing a pre-compiled Scheme program, and not using the REPL or ``eval``.  We have talked, in the past, about breaking this down into smaller piece so that, for instance, if you are running a small script it could be run against a ``minuscule`` Scheme session that doesn't bother loading the expression editor or maybe even the interpreter.

[...]

PR #769, opened 2023-11-24
buffer more reads from a boot-file file descriptor

mflatt:
Before:

```
$ echo "(exit)" | strace -c scheme -q
% time     seconds  usecs/call     calls    errors syscall
------ ----------- ----------- --------- --------- ----------------
 98.87    0.008833           1      4626           read
  0.87    0.000078           1        68           mmap
  0.22    0.000020           2         8           rt_sigaction
  0.03    0.000003           0         6           close
  0.00    0.000000           0         5           mprotect
....
```

After:

```
$ echo "(exit)" | strace -c scheme -q
% time     seconds  usecs/call     calls    errors syscall
------ ----------- ----------- --------- --------- ----------------
 68.42    0.001443          51        28           munmap
 24.56    0.000518           3       150           read
  5.17    0.000109           1        68           mmap
  0.95    0.000020           2         8           newfstatat
....
```

The overall difference in startup time turns out to be very small on Linux and macOS — just a a couple of milliseconds, if anything. But buffering is generally a good idea, and the change makes a startup system-call trace look less suspicious. (See, for example, the discussion in #263.)

I tried this kind of change a couple of times before over the years, but the result always seemed like too much for little benefit. This time, I found a path that involves fewer changes, involves simpler local changes that are easier to check through a diff, and tends to make the code nicer (collapsing two `uptr` decodings into one) instead of more complicated; so, I think it's worthwhile.

In addition to buffering changes, this commit adds `Sregister_boot_file_bytes`, so that an embedding application can point to bytes that are already loaded into memory. I experimented with that by linking a program with `-sectcreate` on macOS and using `getsectiondata` to find embedded boot files at run time. It doesn't make startup any faster, but that kind of approach might be useful for tools that create a single-file executable including boot files.


c/scheme.c
```diff
@@ -1107,20 +1106,37 @@ extern void Sregister_boot_file_fd_region(const char *name,
...
+  boot->close_after = close_after;
+  add_boot(boot, name);
+}
+
+extern void Sregister_boot_file_bytes(const char *name,
+                                      void *data,
+                                      iptr len) {
+  struct boot_desc *boot;
+
+  check_boot_file_state("Sregister_boot_bytes");
```
samth, line 1129: Shouldn't this be `Sregister_boot_file_bytes`?

mflatt: Yes, thanks.

c/fasl.c
```diff
@@ -327,109 +312,59 @@ void S_fasl_init(void) {
...
+  ffo->f.next = ffo->f.end = ffo->f.buf = ffo->buf_space;
 }
 
-ptr S_bv_fasl_read(ptr bv, int ty, uptr offset, uptr len, ptr path, ptr externals) {
-  ptr tc = get_thread_context();
-  ptr x; struct unbufFaslFileObj uffo;
+void S_fasl_init_bytes(faslFile ffo, ptr path, void *data, iptr len) {
+  ffo->uf.path = path;
+  ffo->uf.type = UFFO_TYPE_BV;
+  ffo->uf.fd = 0;
```
burgerrg, line 329: Should this be -1, since file descriptor 0 is STDIN?

mflatt: That's a good idea.

PR #773, opened 2023-11-30
update release notes on `/` and exact `0`

mflatt:
The release notes mentioned a change to the treatment of exact 0 as a divisor, but not as a dividend. The release notes now also point out that this change is not consistent with R6RS.

### DIFF
--- a/release_notes/release_notes.stex
+++ b/release_notes/release_notes.stex
@@ -362,11 +362,14 @@ polling.
 
 \subsection{Exact zeros, transcendentals, and exponentials (9.9.9)}
 
-Division of an inexact number by an exact \scheme{0} now raises
-exception instead of producing \scheme{+inf.0} or \scheme{-inf.0}.
-This makes division more consistent with with the existing behavior of
+Division of an exact \scheme{0} by an inexact number now always
+produces exact \scheme{0}, and
+division of an inexact number by an exact \scheme{0} now raises an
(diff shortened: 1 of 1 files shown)


release_notes/release_notes.stex
```diff
@@ -362,11 +362,14 @@ polling.
...
 \subsection{Exact zeros, transcendentals, and exponentials (9.9.9)}
 
-Division of an inexact number by an exact \scheme{0} now raises
-exception instead of producing \scheme{+inf.0} or \scheme{-inf.0}.
-This makes division more consistent with with the existing behavior of
+Division of an exact \scheme{0} by an inexact number now always
+produces exact \scheme{0}, and
+division of an inexact number by an exact \scheme{0} now raises an
+exception instead of producing \scheme{+inf.0} or \scheme{-inf.0},
+This change makes division more consistent with with the existing behavior of
```
burgerrg, line 369: `with with` => `with`

johnwcowan:
I should think the Right Thing, if R6RS is wrong, to add this to the R6RS errata.  In that case, the `(rnrs)` and `(chezscheme)` libraries can do the same thing.

release_notes/release_notes.stex
mflatt: Fixed - thanks!

burgerrg (approved):

PR #774, opened 2023-11-30
yet another attempt to fix an unreliable test

mflatt:
Commit 612b7adaa7 did not adjust a `compute-size-increments` test in the intended way, because it used the current thread's parameter vector instead of the measured thread's parameter vector. This new attempt adds an extra thread parameter, so it checks the intended repair without having to be lucky with a "thread.ms" test.


burgerrg (approved):
Good find!

PR #775, opened 2023-12-05
Syntax quoting

mnieper:
This contribution adds the `quote-syntax` form to the language. This form, which can also be found in Racket and Guile, is like `syntax` but does not substitute pattern variables.  (In fact, `quote-syntax` is like the `syntax` form from the R<sup>4</sup>RS appendix.)

The `quote-syntax` can be helpful for macro writers when the quoted datum's contents are not controlled by the macro and unwanted capture of pattern variables has to be avoided.

### DIFF
--- a/csug/syntax.stex
+++ b/csug/syntax.stex
@@ -335,15 +335,12 @@ For example:
 \endentryheader
 
 A \scheme{quote-syntax} expression is like a \scheme{quote}
-expression except that contextual information associated with
+expression, except that contextual information associated with
 the datum is retained.
 
 Unlike \scheme{syntax}, \scheme{quote-syntax} does not substitute
 pattern variables bound by \scheme{with-syntax} or
 \scheme{syntax-case}.
-
-(This form is identical to the Revised$^4$ Report
-\scheme{syntax}.)
 
 \schemedisplay
 (let-syntax ([m (lambda (stx)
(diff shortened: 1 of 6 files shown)


csug/syntax.stex
```diff
@@ -327,6 +327,32 @@ For example:
...
 
+%----------------------------------------------------------------------------
+\entryheader
+\formdef{quote-syntax}{\categorysyntax}{(quote-syntax \var{datum})}
+\returns a syntax object
+\listlibraries
+\endentryheader
+
+A \scheme{quote-syntax} expression is like a \scheme{quote}
+expression except that contextual information associated with
```
mflatt, line 338: Add a comma before "except"?

csug/syntax.stex
```diff
@@ -327,6 +327,32 @@ For example:
...
+the datum is retained.
+
+Unlike \scheme{syntax}, \scheme{quote-syntax} does not substitute
+pattern variables bound by \scheme{with-syntax} or
+\scheme{syntax-case}.
+
+(This form is identical to the Revised$^4$ Report
+\scheme{syntax}.)
```
mflatt, line 346: I'd leave out the R4RS reference, just because R4RS has generally be referenced in the manual. (I don't know for certain that there are other places where a similar comment would apply, but it seems likely.)

mflatt (approved):
I have two small suggestions on the doc changes, but this is a nicely complete PR, and the addition of `quote-syntax` is fine by me.

csug/syntax.stex
mnieper: I am not a native speaker, so I followed your advice and added the comma. :) (The CSUG seems to be slightly inconsistent.)

PR #776, opened 2023-12-07
initialize a thread's exception state to a default state

### LINKED ISSUE #452: Child thread should not inherit the exception handler stack
As discussed in #451:
```
> (with-exception-handler
      (lambda (e) (printf "parent~%") (exit))
      (lambda () (fork-thread (lambda () (error 't "")))))
#<thread 1>
parent
```
The exception handler stack is inherited by the child thread, which usually lead to surprising behaviors and memory leaks.
To start a thread with a fresh exception handler stack, I must call `(current-exception-state (create-exception-state))` explicitly inside fork-thread.
It's really counter-intuitive and error-prone.

mflatt:
Fixes #452.

The former behavior was to inherit the exception state of the creating thread, but that's confusing and unlike the treatment of winders, continuation, and continuation marks.

The implementation moves the handler stack from a thread parameter to a dedicated tc slot so that the kernel can initialize the slot for a new thread.


burgerrg (approved):
Thank you for making this important change!

PR #777, opened 2023-12-07
 change default install of bin files to be soft links to lib files

mflatt:
Unless configured with `--installabsolute`, install "scheme", "petite", and "scheme-script" as soft links to executables in the library directory that contains boot files. That way, multiple versions can be installed easily by referencing executables inthe library directory.
    
Configuring with `--installabsolute`, which already restored the old boot search-path default, also restores the old installation behavior.

### DIFF
--- a/configure
+++ b/configure
@@ -624,7 +624,7 @@ if [ "$help" = "yes" ]; then
   echo "  --installschemename=<schemename>  install scheme as ($installschemename)"
   echo "  --installpetitename=<petitename>  install petite as ($installpetitename)"
   echo "  --installscriptname=<scriptname>  install scheme-script as ($installscriptname)"
-  echo "  --installabsolute                 disable relative boot-file search"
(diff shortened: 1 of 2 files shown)


burgerrg:
Do you think it would be better to use the fixed names scheme, petite, and scheme-script in the Lib directory, and use the specified names for the symbolic links? I'm thinking about the code I'll write to find the executables for a particular version in the lib tree.

mflatt:
I don't think it works to use different names for the "lib" files and "bin" links, because whichever name you use to start the process will be the name used to find the corresponding boot file(s).

My guess is that your script to find executables will need to make some assumptions about the configuration, anyway. Or maybe there's a place we could record a configuration so that the script can find things?

burgerrg:
You're right, I didn't think about the boot file naming. It's fine to use the same name. The systems I use don't change the default names.

burgerrg (approved):

PR #784, opened 2023-12-21
ensure write barrier in last step of `vector-sort!`

mflatt:
The old `$vector-copy!` could use `$ptr-copy!`, which doesn't have an associated write barrier. As a result, using `$vector-copy!` at the end of `vector-sort!` is incorrect.

The new test is arranges for one half of a vector to have generation-0 objects and the other to have immediate objects; the two halves are already sorted, so they end up being swapped via a temporary vector and `$vector-copy!`. As a result, a collection afterward can fail to scan the half of the original vector that it should scan.

The solution here is to rename the internal `$vector-copy!` function to `$vector-fill-copy!` (by analogy to `$stencil-vector-fill-set!`) and use keep using it for things like `vector-copy`. But the new `$vector-copy!` as used by `vector-sort!` avoids `$ptr-copy!`.

### DIFF
--- a/mats/5_6.ms
+++ b/mats/5_6.ms
@@ -1340,6 +1340,46 @@
                 (fprintf (console-output-port) "\n~s\n" ls)
                 (errorf #f "failed")))))))
     (make-string 200 #\.))
+
+  (let ([Ns '(10 100 1024 2048 4096)])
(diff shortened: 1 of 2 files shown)


burgerrg (approved):
That was a great find!

PR #790, opened 2024-01-07
Add add1, sub1 and abs to cptypes

gus-massa:
Add special cases for `add1`, `sub1` and `abs` to cptypes. In particular, that they are closed for flonums, reals and exact-integers.


jltaylor-us:
needs more mats to cover all of the clauses in the new specializations

gus-massa:
I added more tests.

jltaylor-us:
Still doesn't test the bignum case for `abs`, but looks ok to me.  I don't think I've actually done any work in cptypes before, though, so maybe one of the other maintainers will give it a quick glance before it goes in.

mflatt:
The changes in `cptypes.ss` look good to me. I agree that it would be good to cover the bignum case of `abs` with a test (which is the one where I had to pause to believe that it's right).

gus-massa:
I don't care too much about `bignum`s because they don't trigger too many optimizations.  I care more about `fixnum`s. 

A long time ago, I added `bignum`s  so `(and (integer? x) (exact? x))` is equivalent to `(or (fixnum? x) (bignum? x))`. This is a reduction in cpnanopass that is not visible by cptypes, but the reduction may be moved to an early pass in the future, or the expression may be in the code written by the programmer. 

Now, I added the test for `bignum`s in `abs` and also negative tests in the others. I also added the negative tests for `fixnum`s. They are sanity checks, because it's too easy to be confused and add a wrong rule for `fixnum`s in the future to cptypes.

PR #795, opened 2024-01-23
Unicode 15.1

burgerrg:


unicode/Makefile
```diff
@@ -1,6 +1,9 @@
-Scheme=../bin/scheme
+Scheme=../bin/zuo .. run
```
melted, line 1: Is that intended? It seems unrelated to unicode.

unicode/Makefile
```diff
@@ -1,6 +1,9 @@
...
 
-doit: unicode-char-cases.ss unicode-charinfo.ss
+doit: update unicode-char-cases.ss unicode-charinfo.ss
+
+update:
+	./get-UNIDATA
```
jltaylor-us, line 6: I'm guessing that the intent of this is to run `update` (to completion) before the other two targets?  Listing them this way allows all three to run in parallel (or any arbitrary order).

Maybe
```
doit:
    ${MAKE} update
    ${MAKE} unicode-char-cases.ss unicode-charinfo.ss
```
(with tabs substituted appropriately)

...although that will still allow parallel execution if someone explicitly names the targets together on the command line.  (I'm not too concerned about that scenario... I say that falls under "play stupid games, win stupid prizes")

unicode/Makefile
burgerrg: Yes, there is no ../bin/scheme any more.

mflatt (approved):
Thanks for this update! I took the liberty of pushing two commits (to squash when merging): the makefile update suggested by @jltaylor-us and an extra test as a sanity check that the 15.1 upgrade worked.

burgerrg:
Thanks, @mflatt and @jltaylor-us!

PR #796, opened 2024-01-24
cp0: combine dropping a binding and `begin` rotation

mflatt:
A follow-up to c081296367, this commit adjusts the cp0 change to prefer an existing case instead of the new one. This order still passes the new test, it passes old ones with a small adjustment, and it passes Racket tests that are similar to "cp0.ms" tests.

### DIFF
--- a/mats/record.ms
+++ b/mats/record.ms
@@ -9008,10 +9008,9 @@
                        (new q x)))))))
            (make-foo 3))))
     `(let ([ctr 0])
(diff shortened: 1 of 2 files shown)


mflatt:
Er... looking one more time, the new example suggests that the updated cp0 order is worse, not better! I'll take this from the top on the Racket test-suite side and figure out where I got confused.

mflatt:
Trying again, this time with @owaddell's repair, a test that reflects the original issue for Racket, and a corrected version of the broken test that showed why my first attempt was bad.

PR #801, opened 2024-02-01
Version 10.0.0

burgerrg:
- `(scheme-version)` now includes the patch version number even when it's 0.
- make install now includes the patch version in the directory name even when it's 0.
- All 9.9.9 instances in the release notes are now 10.0.0.
- The Windows install no longer includes the Microsoft Visual C++ redistributable merge modules. This change was made in Version 9.6.0 but didn't get ported into the Version 10 branch until now.


mflatt:
Thanks for the repair to `(scheme-version)`. The original intent was to follow Semantic Versioning, and I think I probably got confused started mixing up conventions by dropping a `0` as the patch number.

The `makefiles/install.zuo` scripts still drops a `0` as the patch number when building an install directory name. For example, boot files are installed by default into `/usr/local/lib/csv10.0/`. Should that change, too, to make it `/usr/local/lib/csv10.0.0/`? I think the right change would be at line 73:

```
  (define Version (let ([l (get-version)])
                    (~a "csv" (list-ref l 0) "." (list-ref l 1) "." (list-ref l 2)
                        (if (= (list-ref l 3) 0)
                            ""
                            (~a "-pre-release." (list-ref l 3))))))
```

mflatt (approved):

Issue #797, opened 2024-01-24
mat failures with macOS 14's iconv


burgerrg:
Using c08129636735db7f4703875094a44d42a48d59d1 on arm64osx on macOS 14.3 and a6osx on macOS 14.2.1, the following mats fail:
```
Bug in mat iconv-codec clause 6 at line 710, char 7 of ../../mats/io.ms
Bug in mat iconv-codec clause 14 at line 1004, char 7 of ../../mats/io.ms
Bug in mat iconv-codec clause 14 at line 1004, char 7 of ../../mats/io.ms
Bug in mat iconv-codec clause 6 at line 710, char 7 of ../../mats/io.ms
Bug in mat iconv-codec clause 14 at line 1004, char 7 of ../../mats/io.ms
Bug in mat iconv-codec clause 14 at line 1004, char 7 of ../../mats/io.ms
```

The mat for codec GB18030 on line 710 of io.ms fails when make-codec-buffer returns a 4-byte buffer instead of the default 1024-byte buffer. Apparently this iconv-based codec needs more bytes. At 17 bytes it passes.

The mat for codec UTF-8 on line 1004 of io.ms fails unconditionally. It also fails using Chez Scheme 9.5.8 on a6osx, so I think the problem may be due to a change in macOS 14's iconv.

These mats pass on macOS 12, so I suspect the problem involves the newer iconv in macOS 14.

When I install libiconv using Homebrew on macOS 14 and configure Chez Scheme to link against it, all the mats pass.

ncihnegn:
There are many reports about macOS Sonoma `iconv` compatibility issues.
For example, https://github.com/orgs/Homebrew/discussions/4884.

PR #805, opened 2024-02-19
adjustment and workaround for macOS iconv

mflatt:
Intended to address #797.

History note: the iconv library in macOS 14.0 was especially broken. By 14.2.1, it has been mostly fixed, but I think the Chez Scheme test suite uncovers a remaining problem. That problem persists in 14.3.1. 

The two different failing cases have different reasons:

* The test at line 1004 fails because iconv encodes a BOM to UTF-8 as 0 bytes. That's arguably correct behavior (a BOM encoding isn't supposed to appear in UTF-8), and the test suite can accommodate it by just avoiding a BOM.

* The test at line 710 fails due to behavior that seems to be a lingering bug in iconv, where the library incorrectly reports too little decoding success (by not advancing the output pointer) than it should when failing to convert a lambda character to a Chinese encoding. The commit here works around that by not trusting the pointer advances when an a specific error is reported. Unless I have it wrong, the change is ok for a working iconv, too, in part due to the way the workaround is specific to an `E2BIG` result.

I'm not sure the bug-workaround change is a good idea. Workarounds for OS bugs are sometimes necessary, but it's usually easier to get a sense of the bug's extent so that a workaround feels worthwhile. This workaround is fairly specific to a random corner of a library, and I can't help thinking there are likely other random corners with bugs. Still, I lean toward having the workaround, for now. An alternative is to disable the failing test on macOS.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,6 +15,10 @@ jobs:
             os: macos-12
           - machine: ta6osx
             os: macos-12
+          - machine: arm64osx
+            os: macos-14
+          - machine: tarm64osx
+            os: macos-14
           - machine: i3le
             os: ubuntu-22.04
           - machine: ti3le
--- a/c/prim5.c
+++ b/c/prim5.c
@@ -2273,6 +2273,50 @@ static void s_iconv_close(uptr cd) {
   ICONV_CLOSE((iconv_t)cd);
 }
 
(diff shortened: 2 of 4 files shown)


mflatt:
Forgot to mention: with the workaround in place, there's a commit here to enable arm64osx in the CI action.

jltaylor-us (approved):

burgerrg (approved):
Thanks for figuring this out!

PR #807, opened 2024-02-27
Repairs and improvements for building with external dependencies

LiberalArtist:
This patch series repairs and improves support for configuring Chez Scheme to use external copies of some dependencies normally included as submodules. Overall, these changes let Chez Scheme be built in ways that already worked to build older versions of Chez Scheme and the copy of Chez Scheme distributed with Racket (because the configure script in the Racket sources tried less hard to make sure the dependencies were really present).

 1. The first patch fixes the checks for the `zlib` and `lz4` submodules, which are currently noops, to actually perform checks unless configured with `ZLIB=<zlib>` or `LZ4=<lz4>`, respectively.
 2. The second patch adds an analogous `ZUO=<zuo>` argument. Rather than simply skip the submodule check and require the user then to build with `make ZUO=zuo` (or to use Zuo directly), this patch records the configuration choice in `Mf-config` and slightly adjusts `Makefile.in` to adapt more gracefully to this mode.
 3. The final patch similarly adds an `STEXLIB=<stex>` argument to `configure` and updates `build.zuo` as I had considered in https://github.com/racket/racket/pull/4203#issuecomment-1107692530 after https://github.com/racket/racket/commit/d897379ef64178347ba7af3f0314317433d00410:
    > The only way to use an installed Stex up to this point has been running `make STEXLIB=` with GNU Make or some other `make` that will propagate it as needed. I have only adjusted the Zuo script to recognize an `STEXLIB` passed that way: I haven't tried to support setting `STEXLIB` via `configure` or other mechanisms. There would be some uniformity in doing so, but it wouldn't be useful for Guix packaging unless it were also supported upstream, since we try to share as much as possible between the package definitions for the two Chez Schemes.

In fact, with these changes cherry-picked as downstream patches, my draft update of the [Guix package](https://packages.guix.gnu.org/packages/chez-scheme) to Chez Scheme 10 is able to share its *entire* build recipe exactly with our [`chez-scheme-for-racket`](https://packages.guix.gnu.org/packages/chez-scheme-for-racket) package (which we used for the architectures newly supported in version 10 and will continue to use for bootstrapping).

### DIFF
--- a/BUILDING
+++ b/BUILDING
@@ -149,7 +149,8 @@ information on the supported options.
 The generated makefile mostly just ensures that a `zuo` executable is
 built in a `bin` directory, and then it defers the actual build work
 to `zuo`, which uses the "main.zuo" file. If you have `zuo` installed,
-you can use `zuo` directly instead of `make`. In general, instead of
+you can use `zuo` directly instead of `make`: in that case, you may
+wish to use `./configure ZUO=<zuo>`. In general, instead of
 the command `make X` to build target `X` as described below, you can
 use `zuo . X` (or `bin/zuo . X` after `bin/zuo` is built).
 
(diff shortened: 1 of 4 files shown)


mflatt (approved):

mflatt:
This looks good to me. I'll wait a day or two before merging to see whether others detect an issue, though.

PR #812, opened 2024-02-28
Update libraries.stex. Fixes a typo

olopierpa:
Only a typo.


jltaylor-us:
Thanks for fixing this!

Issue #809, opened 2024-02-28
Intermittent crash in `make test`


LiberalArtist:
Quoting @mflatt in https://github.com/cisco/ChezScheme/issues/808#issuecomment-1967671535:

> Meanwhile, the test failure appears to be a crash in the test related to signal handling. I have seen this on occasion, and I suspect an issue with the test, but I didn't figure it out the last time I tried.

I've hit this problem several more times while building 10.0.0 and 9.9.9-pre-release.23 (from Racket 8.12). I don't really know what's going on, but I thought I should open this to track it.

LiberalArtist:
I've hit the issue with `make test-some` in addition to `make test`. IIRC, every time it's happened the last line has included `eval=interpret`, so I'll try `make test-some-fast` next.

LiberalArtist:
I've had `make test-some-fast` fail, too (last line: `-------- o=3 cp0=t cp0 --------`), so I think I just need to skip tests for now.

LiberalArtist:
I have not encountered this issue myself or in the Guix CI process since backporting https://github.com/cisco/ChezScheme/pull/813, so hopefully this is fixed!

PR #813, opened 2024-02-29
constrain signal delivery to Scheme to the main thread

mflatt:
This change is an attempt to address #809.

My guess here is that the test crashes when signal 14 is delivered to a thread that is not a Scheme thread. In particular, if a GC has happened previously while multiple threads were running, then GC worker threads may be waiting around. They haven't masked signals, and the thread-local variable for a thread context will be NULL in those worker threads.

In that scenario, I don't think the signal would get delivered to a non-main thread on macOS, because `sigprocmask` and the automatic masking of signals during delivery is process-wide there. On Linux, however, the signal mask is thread-specific. That difference would explain why I see the crash rarely and (in retrospect, as far as I can remember) only when trying out different Linux systems and not when working in my main macOS development environment.

Before this change, I can force a crash by ensuring that GC threads have been created, disabling signals in the main thread in Linux, and then having a handled signal delivered to the Scheme process. So, maybe the test was sometimes crashing when the main thread happens to have signals blocked in the main thread when signal 14 is delivered to the process. That particular problem could not have happened before parallelism was added to the GC, and so maybe it would never have happened in the test before. In practice, there are often extra threads running in a process, and that's why this patch adjusts the signal handler to redirect to the main thread instead of setting the signal mask in GC worker threads.

### DIFF
--- a/c/globals.h
+++ b/c/globals.h
@@ -49,6 +49,7 @@ EXTERN int S_num_preserve_ownership_threads;
 # ifdef IMPLICIT_ATOMIC_AS_EXPLICIT
 EXTERN s_thread_mutex_t S_implicit_mutex;
 # endif
+EXTERN s_thread_t S_main_thread_id;
 #endif
 
 /* segment.c */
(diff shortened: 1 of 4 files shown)


burgerrg (approved):
This was a good find!

PR #815, opened 2024-03-09
add configuration for s390x (and related) to `c/version.h`

mflatt:
This configuration enables pb mode to run on s390x.

See also to racket/racket#4951.


akeep (approved):
Looks good! Cool to hear someone is looking to use Racket on a zSeries mainframe!

PR #816, opened 2024-03-10
adjust `configure ZUO=<command>` support

mflatt:
Continuing from b8838c3280, adjust the generated makefile so the supplied <command> is not a makefile dependency. That way, `ZUO=zuo` works if `zuo` is installed and the current build directory is not the source directory. (The `zuo` executable is a dependency in a real and relevant sense, but not in the sense of dependencies that we normally track in makefiles.)

Also adapt the makefile for the case that `ZUO=...` is not supplied and the build directory is not the source directory, in which case `ZUO_LIB_PATH` needs to be relative to the source directory.

Using `make ZUO=zuo` can also work, but in that case, `bin/zuo` is still built as a dependency. It's possible that some portable makefile magic could overcome that limitation, but it doesn't seem important.

CC: @LiberalArtist

### DIFF
--- a/configure
+++ b/configure
@@ -904,10 +904,12 @@ if [ "${zuoExternal}" = "" ] ; then
         submod_instructions 'Source in "zuo" is missing'
     fi
     ZUO="bin/zuo"
+    ZUO_DEP="${ZUO}"
     RM_ZUO="rm -f bin/zuo"
     ZUO_TARGET="bin/zuo"
 else
     ZUO="${zuoExternal}"
+    ZUO_DEP=""
     RM_ZUO="@echo 'Not cleaning external ${zuoExternal}'"
     ZUO_TARGET="DoNotBuildZuo"
 fi
@@ -1153,6 +1155,7 @@ alwaysUseBootFile=$alwaysUseBootFile
 relativeBootFiles=$relativeBootFiles
 
 ZUO=$ZUO
+ZUO_DEP=$ZUO_DEP
 RM_ZUO=$RM_ZUO
 ZUO_TARGET=$ZUO_TARGET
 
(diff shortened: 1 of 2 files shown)


LiberalArtist:
> Continuing from [b8838c3](https://github.com/cisco/ChezScheme/commit/b8838c3280ef10e115236d2f7ac9ae857f83e268), adjust the generated makefile so the supplied `<command>` is not a makefile dependency. That way, `ZUO=zuo` works if `zuo` is installed and the current build directory is not the source directory.

Just to make sure I'm following correctly, without this patch, is the problem that `./configure ZUO=zuo` still had `$(ZUO)` (expanding to `zuo`) as a prerequisite in the makefile rules, but had no recipe to build `zuo`, so it only worked if a file or directory named `zuo` happened to already exist in the build directory (as is the case when building in the source directory)? I know I tested both `./configure ZUO=zuo` and out-of-source builds, but maybe I didn't test the right combination of configurations to see the problem.

> Using `make ZUO=zuo` can also work, but in that case, `bin/zuo` is still built as a dependency. It's possible that some portable makefile magic could overcome that limitation, but it doesn't seem important.

Before https://github.com/cisco/ChezScheme/pull/807, I had been using `make ZUO=zuo`, but, even so, I'm inclined to agree that this minor limitation isn't worth the complexity I expect would be needed to avoid it portably.

LiberalArtist:
LGTM, then!

LiberalArtist:
Oh, trivially, in the commit message, you may want to put backticks around `<command>`: otherwise, where passed as Markdown, it's treated as an unsupported raw HTML tag, so it doesn't render, e.g. "so the supplied <command> is not a makefile dependency".

PR #818, opened 2024-03-23
replace call to iconv procedure with ICONV macro

Bogdanp:
The availability of the iconv procedure varies with different build settings, so I believe the ICONV macro should be used here instead. With this change in place, I'm able to get PB Racket CS running on iOS.


mflatt:
Thanks!

PR #820, opened 2024-03-24
reduce memory use of pbchunk conversion

mflatt:
C code generated by the pbchunk conversion is especially verbose, with long macro names for simple operations and lots of comments to reflect source instructions. Generation of that code was formerly rendering to text in memory, and then splitting into the output files, which turns out to use a lot of memory. Converting to text at the last minute can have a much lower peak memory use.

If memory is at a premimum, then rendering to a temporary file would work even better, but this change already brings memory use in line with other build actions without having to deal with a temporary file. Memory use could also be reduced by dropping comments that connect to source instructions; again, though, doesn't seem necessary for now.

Related to racket/racket#4955.

### DIFF
--- a/s/pbchunk.ss
+++ b/s/pbchunk.ss
@@ -17,7 +17,7 @@
 (define-record-type chunk-info
   (fields (mutable counter)
           seen
-          code-op)
+          code-accum)
   (nongenerative))
 
 ;; A chunklet represents a potential entry point into a code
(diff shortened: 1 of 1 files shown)


s/pbchunk.ss
```diff
@@ -837,65 +846,73 @@
...
+         (define (emit-pre)
+           (emit-c! o "/* 0x~x */" i)
+           (unless (>= i start-i)
+             (emit-c! o " /*")))
+         (define (emit-post)
+           (unless (>= i start-i)
+             (emit-c! o " */")))
+
+         (define (emit-desc fmt . args)
+           (apply emit-c! o fmt args))
```
jltaylor-us, line 858: I think these names should be suffixed with `!` for consistency with `emit-c!`

jltaylor-us (approved):

s/pbchunk.ss
mflatt: Good point about the inconsistency. Do you think it would be ok to instead drop the `!` on `emit-c!`? The `emit-c!` function is replacing `fprintf` here, and there are many more `emit` functions in the pbchunk implementation that were originally about writing to a port. That is, I put a `!` on `emit-c!` because it ends up at `set-box!`, but I'm now thinking I should have left off the `!` because it's implementing something like a port.

jltaylor-us: That would work for me, too.

mflatt: Ok, done!

Issue #821, opened 2024-03-25
bytevector-reference-ref & set! fail to correctly check size of bytevectors smaller than ptr-bytes - 1


burgerrg:
```
> (define rbv (make-reference-bytevector 1))
> (bytevector-reference-set! rbv 100 '(oh oh))
> (bytevector-reference-ref rbv 100)
(oh oh)
> rbv
#vu8(0)
```

The unsigned check `($fxu< (fx- (bytevector-length bv) (fx- (constant ptr-bytes) 1)) i)` doesn't work as designed when the first argument goes negative.

mflatt:
Thanks for tracking that down! In the borken checks, I was trying to use the `$fxu<` trick used in other functions to guard against both a too-large value and a negative value in one step — but that doesn't work for a larger range (word-sized, in this case) that needs to be in bounds on both ends.

PR #823, opened 2024-03-26
repair GC for reference bytevector whose length isn't a multiple of a word

mflatt:
A reference bytevector is intended to be used as a multiple of a word size, but there's no explicit constraint, and so it should work with any length. When the length is not a multiple of the word size, then in a copying step, extra bytes at the end of the bytevector were missed.

The bug was exposed on a 32-bit platform by tests added as part of #821.

### DIFF
--- a/boot/pb/gc-ocd.inc
+++ b/boot/pb/gc-ocd.inc
@@ -131,9 +131,26 @@ static IGEN copy(thread_gc *tgc, ptr p, seginfo *si, ptr *dest)
                   {
                     uptr len = Sbytevector_reference_length(p);
                     memcpy_aligned(&BVIT(new_p, 0), &BVIT(p, 0), ptr_bytes * len);
-                    if ((len & 1) == 0)
                     {
-                      INITBVREFIT(new_p, len) = FIX(0);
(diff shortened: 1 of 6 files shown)


burgerrg:
We should add an entry in Bugs Fixed of the release notes for the reference bytevector work.

burgerrg:
Could the C code simply memcpy the reference bytevector in its entirety the way the other bytevector code does, something like the following?
```
        else if (TYPEP(tf, mask_bytevector, type_bytevector))
        {
          ISPC p_spc = (si->space == space_reference_array) ? space_reference_array : space_data;
          {
            uptr sz = size_bytevector((Sbytevector_length(p)));
            {
              uptr p_sz = sz;
              find_gc_room(tgc, p_spc, tg, type_typed_object, p_sz, new_p);
              memcpy_aligned(&BYTEVECTOR_TYPE(new_p), &BYTEVECTOR_TYPE(p), sz);
            }
          }
        }

```

mflatt:
Good point!. The release notes are now updated in the patch.

burgerrg (approved):

PR #824, opened 2024-04-04
repairs for pb32 foreign interface

mflatt:
Support via libffi for pb32 had not been well tested before. Just running existing tests turned up several problems in the handling of callable addresses and 64-bit arguments and results.

To run tests, I used a 32-bit x86 machine, but it turns out that pb32l is not completely compatible with non-Windows 32-bit x86, because pb32 assumes 8-byte alignment in structures for `double`s and `int64`s. If we wanted to better support pb32 builds on x86, then probably there should be a new machine variant alongside pb32l. I don't see any real use for that, though, since 32-bit x86 machine code is supported directly. (For testing, I just hacked the alignment in a pb32l build by changing `pbcommon.def`.)

### DIFF
--- a/IMPLEMENTATION.md
+++ b/IMPLEMENTATION.md
@@ -1353,6 +1353,15 @@ code for a basic build; on a big-endian machine, the kernel rewrites
 instruction bytes to big-endian form while loading a fasl file, so the
 interpreter can decode instructions in native order.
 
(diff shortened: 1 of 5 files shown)


mflatt:
I've now also run tests for pb32b, so I'm fairly confident that the repair is an improvement.

PR #825, opened 2024-04-07
add missing memory fences for tppc32le and pb variants

mflatt:
For tppc32le, memory-fence generation needed to be turned on in cpprim. For pb variants, use `__sync_synchronize()` as provided by GCC and Clang to support architectures other than ones explicitly covered in "atomic.h".

These changes are intended to address racket/racket#4952 and https://racket.discourse.group/t/let-us-know-when-platforms-specific-packages-have-been-updated-to-8-12/2712/12. In the case of tppc32le, I'd never tried running on a multicore Power-architecture machine; it only today occurred to me that the ppc64l machine that I can access (via GNU cfarm) would be able to compile and run ppc32-bit binaries, too. Using that machine, I was able to confirm that the tppc32le port needs this fence enabled for `thread.ms` to pass.


burgerrg:
Good find! Please update the release notes, too.

burgerrg (approved):

PR #829, opened 2024-04-23
configure: improve inference of compiler flags for pb builds

mflatt:
A follow-up to #828 

When building an pb variant like `tpb64l`, and even when building for a platform without native-code support, the "configure" script may still be able to infer suitable C compilation and linking flags for the kernel. For example, configuring with just `-m=tpb64l` on ppc64l Linux should pick up good compiler and linker flags.

An `--os=...` flag can override inference in "configure" in case it's not correct, which might be needed for cross compilation of pb variants.

With these changes, `-m=pb` by itself is now equivalent to `--pb`, instead of triggering an error to say that `--pb` must be used. Using `-m=pb` is not equivalent to `--pb` in all cases, since `--pb --threads` is like `-m=tpb`, but `-m=pb --threads` would be an error due to the mismatch between an explicit `pb` machine type and `--threads` flag. The interactions are subtle, but I'm hopeful that "configure" now does the right thing in more cases.

### DIFF
--- a/BUILDING
+++ b/BUILDING
@@ -146,6 +146,14 @@ type of build and the installation location. For example,
 specifies the installation root. Run `./configure --help` for
 information on the supported options.
 
+For platforms without support for native-code compilation in Chez
+Scheme, use a machine specification like `-m=tpb64l`, which is a
+threaded, 64-bit, little-endian build. The "configure" script will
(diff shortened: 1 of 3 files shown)


nmeum:
This look great, thanks for following up on this! Should make it easier for us to support portable bytecode targets downstream :)

PR #837, opened 2024-05-29
Allow building against cosmopolitan libc

jjsullivan5196:
~~Accompanies the same changes in the racket repo https://github.com/racket/racket/pull/5004~~

This was originally a patch just to get racket to build on cosmopolitan libc, which unfortunately would require more heavy lifting in the main distribution to work. I took what I had there and made it so that at least Chez will work more or less as advertised, a single binary that can hit many common operating systems on amd64 and aarch64, with no installation required.

See the cosmopolitan project page for more information https://github.com/jart/cosmopolitan

### DIFF
--- a/c/expeditor.c
+++ b/c/expeditor.c
@@ -677,7 +677,7 @@ static void s_ee_set_color(int color_id, IBOOL background) {
 # include <unistd.h>
 # include <time.h>
 #endif
-#if !defined(__GLIBC__) && !defined(__OpenBSD__) && !defined(__NetBSD__) && !defined(__linux__) && !defined(__EMSCRIPTEN__) && !defined(NO_USELOCALE)
+#if !defined(__GLIBC__) && !defined(__COSMOPOLITAN__) && !defined(__OpenBSD__) && !defined(__NetBSD__) && !defined(__linux__) && !defined(__EMSCRIPTEN__) && !defined(NO_USELOCALE)
 # include <xlocale.h>
 #endif
 
--- a/c/prim5.c
+++ b/c/prim5.c
@@ -749,8 +749,8 @@ static ptr s_system(const char *s) {
 #ifdef WIN32
   return Sinteger(status);
 #else
-  if WIFEXITED(status) return Sinteger(WEXITSTATUS(status));
-  if WIFSIGNALED(status) return Sinteger(-WTERMSIG(status));
+  if (WIFEXITED(status)) return Sinteger(WEXITSTATUS(status));
+  if (WIFSIGNALED(status)) return Sinteger(-WTERMSIG(status));
   S_error("system", "cannot determine subprocess exit status");
   return 0 /* not reached */;
 #endif /* WIN32 */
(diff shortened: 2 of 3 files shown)


mflatt:
These changes seem ok to me. The change in `c/prim5.c` is a general improvement and correction, and the `defined(__COSMOPOLITAN__)` tests seem like a low maintenance burden.

With these changes and a small adjustment to `lz4/lib/Makefile` (expand the command that uses `-c $^` to four separate commands), I was able to build with

```
./configure --pb CC=/path/to/cosmocc CFLAGS=-O2 AR=/path/to/cosmoar --disable-curses
```
or specifically for x86_64 with

```
./configure CC=/path/to/x86_64-unknown-cosmo-cc CFLAGS=-O2 AR=/path/to/x86_64-unknown-cosmo-ar --disable-curses --disable-x11
```

Leaving out `--pb` and using plain `cosmocc`/`cosmoar` leads to an error about x64_64 inline assembly when trying to compile for Arm64. I imagine that's expected and that the cosmo tools need to be used in a different way to support both x86_64 and Arm64 at the same time.

jjsullivan5196:
Hi @mflatt ! Actually it that change to the makefile wouldn't be necessary, its probably better to use cosmocc's platform specific compilers with Chez's existing cross compilation pipeline. 

I have a different branch that includes building for CI, I just haven't had a chance to clean it up yet. I can finish up and make a PR for it rather than this one https://github.com/jjsullivan5196/ChezScheme/tree/cosmolibc-ci

jjsullivan5196:
@mflatt I pulled in the CI scripts into this PR. If it would be too much hassle I understand, but I thought it would be illustrative at least.

jjsullivan5196:
One last change, since dynamic linking is only in experimental support in cosmopolitan, I disabled it [explicitly for the platform](https://github.com/cisco/ChezScheme/pull/837/files#diff-1197f456ca6b58eec7e78f6925ac40ce5330d7a8630bdad356464227ff40b415R91)

mflatt:
@jjsullivan5196  — Thank you for the updates, and sorry for the long delay!

As you anticipated, the CI scripts seem like more than we'd want to take on right now. But they do seem usefully illustrative, and I wonder how we can keep them in the record for future reference. I'd like to merge the rest, and we normally merge changes using "Squash and merge" and keep the PR number in the commit message. So, whatever would keep them in this PR (and not in the result) would be good. Attaching the scripts as files to a comment?

jjsullivan5196:
Hi @mflatt 

No worries on the wait, this was more a hobby interest than anything, but always happy to get a really great scheme more places! I force pushed out the CI scripts, but kept them in a separate branch here https://github.com/jjsullivan5196/ChezScheme/commit/4b87fe29b7c918cbf05e8612cc1f995b97422bbf

Also as for properly preserving these, there's a software distribution that was made by one of the cosmo maintainers, I think that might be the most appropriate place to keep that information permanently. I'll work on a patch and send it to them, and update this thread if it gets accepted https://github.com/ahgamut/superconfigure

mflatt:
Thanks!

PR #841, opened 2024-06-16
prefix parameter v should be case insensitive

olopierpa:


jltaylor-us:
Nice find; simple fix... I like it!  Probably worth adding a test - just duplicating one of the few tests of `~v` in mats/format.ms should be sufficient, I'd think.  And an entry in the release notes...

Unfortunately, I think release_notes.stex has not yet been updated for capturing changes since the 10.0.0 release.  That's on us, since I assume that at least some of the bug fixes that @mflatt has committed since the 10.0.0 release are Release Note worthy (and I watched them all go in without thinking about it, either).  We'll get back to you on that.

olopierpa:
Thanks. Added the tests in a different merge request. Sorry I don't know how to merge the 2 requests.

olopierpa:
Trying again shortly

PR #842, opened 2024-06-16
Tests for format's V prefix parameters

olopierpa:
Added tests for format's V prefix parameters


olopierpa:
Trying again shortly

Issue #840, opened 2024-06-16
Bug in V format prefix parameter


olopierpa:
Hello,

the CSUG says: "Chez Scheme's implementation of format supports all of the Common Lisp format directives except for those specific to the Common Lisp pretty printer."

The CLHS says: "In place of a prefix parameter to a directive, V (or v) can be used."
https://www.lispworks.com/documentation/lw60/CLHS/Body/22_c.htm

However,
```
> (format #f "~vA" 10 10)
"10        "
> (format #f "~VA" 10 10)
Exception in format: unrecognized directive ~V
Type (debug) to enter the debugger.
```

burgerrg:
Fixed in https://github.com/cisco/ChezScheme/pull/843. Thank you for finding and fixing this!

PR #843, opened 2024-06-17
format's V prefix parameter

olopierpa:
format's V prefix parameter should be case insensitive

### DIFF
--- a/mats/format.ms
+++ b/mats/format.ms
@@ -1461,8 +1461,11 @@
 
 (mat format-slib17 ; variable parameter
   (equal? (format "~va" 10 "abc") "abc       ")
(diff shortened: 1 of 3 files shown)


burgerrg:
Please add a subsection under Bug Fixes in release_notes/release_notes.stex briefly describing the fix. Thanks!

olopierpa:
I'm not proficient in Latex, so please someone else make the actual editing.

My proposed text is "The prefix parameter V in format strings is case insensitive in CL. Chez previously accepted only lowercase v, now both cases are valid."

Thanks!

burgerrg:
See https://github.com/cisco/ChezScheme/issues/840.

Issue #845, opened 2024-06-23
error: use of unknown builtin '__builtin_add_overflow'


ryandesign:
ChezScheme 10.0.0 fails to build with Apple Clang versions earlier than 800:

```
c/pb.c:122:10: error: use of unknown builtin '__builtin_add_overflow' [-Wimplicit-function-declaration]
    case COMMON_INSTR(pb_bin_op_pb_signal_pb_add_pb_register)
         ^
c/pb.c:51:28: note: expanded from macro 'COMMON_INSTR'
#define COMMON_INSTR(x) x: doi_ ## x(instr); break;
                           ^
<scratch space>:73:1: note: expanded from here
doi_pb_bin_op_pb_signal_pb_add_pb_register
^
c/pb.h:384:4: note: expanded from macro 'doi_pb_bin_op_pb_signal_pb_add_pb_register'
   do_pb_bin_op_pb_signal_pb_add_pb_register(INSTR_drr_dest(instr), INSTR_drr_reg1(instr), INSTR_drr_reg2(instr))
   ^
c/pb.h:390:12: note: expanded from macro 'do_pb_bin_op_pb_signal_pb_add_pb_register'
    flag = __builtin_add_overflow(a, b, &r);                         \
           ^
```

This was reported to MacPorts here: https://trac.macports.org/ticket/70278

You use `__builtin_add_overflow` if `USE_OVERFLOW_INTRINSICS` is defined:

https://github.com/cisco/ChezScheme/blob/3d1579e6c67e145895e6a0e7d0f9bf2b8853fbb3/c/pb.h#L382-L404

And you defined it for gcc 5 or greater or any version of clang:

https://github.com/cisco/ChezScheme/blob/3d1579e6c67e145895e6a0e7d0f9bf2b8853fbb3/c/pb.h#L74-L78

But older versions of clang don't support it. Instead of relying on compiler version numbers (which is complicated to get right for clang, since there are two different clangs, one released by llvm.org, the other released by Apple, with different version numbering schemes), please use a method like `__has_builtin` to discover whether it is available.

mflatt:
Closed by #846

PR #846, opened 2024-06-23
use `__has_builtin` to detect available intrinsics

mflatt:
As recommended in #845, use `__has_builtin` when available.

### DIFF
--- a/c/version.h
+++ b/c/version.h
@@ -33,6 +33,14 @@
 #define FORCEINLINE static inline
 #endif
 
+/* GCC 10 and later and all versions of Clang provide `__has_builtin` for
+   checking for builtins. */
+#ifdef __has_builtin
+# define C_COMPILER_HAS_BUILTIN(x) __has_builtin(x)
+#else
+# define C_COMPILER_HAS_BUILTIN(x) 0
+#endif
+
 /*****************************************/
 /* Architectures                         */
 
--- a/c/atomic.h
+++ b/c/atomic.h
@@ -52,7 +52,7 @@
 # define STORE_FENCE()   __asm__ __volatile__ ("dbar 0" : : : "memory")
 # define ACQUIRE_FENCE() STORE_FENCE()
 # define RELEASE_FENCE() STORE_FENCE()
-#elif (__GNUC__ >= 5) || defined(__clang__)
+#elif (__GNUC__ >= 5) || C_COMPILER_HAS_BUILTIN(__sync_synchronize)
 # define STORE_FENCE() __sync_synchronize()
 # define ACQUIRE_FENCE() STORE_FENCE()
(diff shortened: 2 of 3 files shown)


jltaylor-us (approved):

c/version.h
```diff
@@ -33,6 +33,13 @@
 #define FORCEINLINE static inline
 #endif
 
+/* Clang and recent versions of GC provide `__has_builtin` for
+   checking for builtins. This definition completes a typical pattern
+   to make it usable everywhere: */
+#ifndef __has_builtin
+# define __has_builtin(x) 0
+#endif
```
ryandesign, line 41: I have often seen the recommendation *not* to define your own `__has_builtin` or similar macros since they are part of the implementation, and instead to define and use your own macro name, e.g.:

```suggestion
#ifdef __has_builtin
# define CHEZ_HAS_BUILTIN(x) __has_builtin(x)
#else
# define CHEZ_HAS_BUILTIN(x) 0
#endif
```

This recommendation is important for public headers so that you are not potentially affecting other projects' uses of `__has_builtin`. I don't think version.h gets installed so I'm not sure how important it is in that case. The clang documentation does recommend defining `__has_builtin` if not available, just as you've done.

mflatt: Although these are private headers, I agree that defining a different name in better. Someone reading/modifying "pb.h" or "atomic.h" might spend less time wondering whether the use of `__has_builtin` needs to be guarded with a check to see whether `__has_builtin` exists.

As a minor point, I'd like to go with `C_COMPILER_HAS_BUILTIN`, though, since I can't help reading `CHEZ_HAS_BUILTIN` as a "Chez Scheme has builtin". Other private definitions generally don't use a prefix like `CHEZ`.

c/version.h
```diff
@@ -33,6 +33,13 @@
 #define FORCEINLINE static inline
 #endif
 
+/* Clang and recent versions of GC provide `__has_builtin` for
```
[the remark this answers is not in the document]
mflatt: Thanks!

c/version.h
ryandesign: > I'd like to go with `C_COMPILER_HAS_BUILTIN`, though

As you wish. For definitions in public headers, it would be important to use a macro name that will not conflict with other macros, and the usual way is to include your project name in the macro name. Looking through some of the headers installed on my system, Ruby for example defines `RBIMPL_HAS_BUILTIN`, Eigen defines `EIGEN_HAS_BUILTIN`, and xxHash defines `XXH_HAS_BUILTIN`. If this is a private header it's only a problem if you're including a header from another project and they have inadvisedly defined a macro `C_COMPILER_HAS_BUILTIN`.

PR #852, opened 2024-07-16
small error-message consistency improvement for bit indices

mflatt:
Functions like `vector-ref`, `string-ref`, `substring`, and `bitwise-bit-field` report a non-exact-integer index as a "bad index" error. But `bitwise-bit-set?`, `logbit?`, `logbit0`, and `logbit1` said "not an exact integer", instead, while reporting "bad index" for a negative exact integer; `bitwise-copy-bit` said "not a nonnegative exact integer". This commit changes the error message for those last five functions to always say "bad index".

Fixnum variants of the operations can still say "not a fixnum" or "invalid bit index", depending on whether the index is a negative fixnum. That seems justified on the grounds that fixnum operations add an extra constraint on their arguments. Another possible direction is to change the error messages to consistently "not a nonnegative exact integer" and "not a nonnegative fixnum", since (unlike other index cases), the index validity is independent of the indexed value. After trying out both, sticking with "invalid bit index" seems clearest and the most in the spirit of the current error messages.

### DIFF
--- a/mats/patch-compile-0-t-f-f
+++ b/mats/patch-compile-0-t-f-f
@@ -1320,7 +1320,7 @@
 ! 5_3.mo:Expected error in mat logbit?: "incorrect argument count in call (logbit?)".
 ! 5_3.mo:Expected error in mat logbit?: "incorrect argument count in call (logbit? 1)".
 ! 5_3.mo:Expected error in mat logbit?: "incorrect argument count in call (logbit? 1 2 3)".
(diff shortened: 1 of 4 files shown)


mflatt:
Just to elaborate on my motivation here: the small consistency improvement enables consistency in Racket error                                       messages where that wasn't possible before — even though error messages are rewritten by Racket. Using "not a nonnegative exact integer" and "not a nonnegative fixnum" would be an ever better fit for Racket, so that direction would also work for my purposes.

burgerrg (approved):

PR #853, opened 2024-07-21
fix missing `wcwidth` declaration on GNU Hurd

weinholt:
Building on Debian GNU/Hurd i386 fails because of an implicit declaration of `wcwidth`. This commit makes GNU/Hurd use the same declaration that is used on GNU/Linux.

### DIFF
--- a/c/expeditor.c
+++ b/c/expeditor.c
@@ -681,7 +681,7 @@ static void s_ee_set_color(int color_id, IBOOL background) {
 # include <xlocale.h>
 #endif
 
-#if defined(__linux__) && !defined(_XOPEN_SOURCE)
+#if (defined(__gnu_hurd__) || defined(__linux__)) && !defined(_XOPEN_SOURCE)
 extern int wcwidth(wchar_t);
 #endif
 


mflatt:
I didn't manage to try this myself, but the change makes sense. Thanks!

LiberalArtist:
Thanks for this repair! @weinholt, I noticed that you are checking `defined(__gnu_hurd__)` here. When I added Hurd support, I used `defined(__GNU__)` based on <https://www.gnu.org/software/hurd/hurd/porting/guidelines.html>&mdash;but I don't actually have particular experience with the Hurd; I just maintain the distro packaging of Chez for Guix.

I actually think I prefer `defined(__gnu_hurd__)` as more self-documenting: there are lots of GNU things one might reasonably want to check for, and in fact I tended to write things more like `defined(__GNU__) /* Hurd */`. Do you have a reference for `__gnu_hurd__` being defined?

weinholt:
> Do you have a reference for __gnu_hurd__ being defined?

A GCC changelog entry from 2002 and observed behavior: `gcc -dM -E - < /dev/null`. In the GCC source code tree, `__gnu_hurd__` and `__GNU__` are both defined in `gcc/config/gnu.h`.

PR #854, opened 2024-07-21
fix bytecode endianness handling on s390x, m68k and hppa

weinholt:
Building on the s390x, m68k and hppa architectures using bytecode does not work. The build fails with `illegal pb instruction` or `invalid memory reference` while running bytecode. This commit fixes that by amending an architecture list that lists big endian architectures.


mflatt:
Thanks!

PR #855, opened 2024-07-21
configure: automatically detect endianness and word size on Linux

weinholt:
Most Linux systems have the `getconf` and `lscpu` programs that allow us to automatically detect the system's endianness. This commit uses those commands to set the default bytecode endianness and machine word size.

### DIFF
--- a/configure
+++ b/configure
@@ -126,6 +126,14 @@ CONFIG_UNAME=`uname`
 # using `uname`, infer OS-based defaults
 case "${CONFIG_UNAME}" in
   Linux)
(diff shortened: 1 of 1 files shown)


mflatt:
Thanks for these PRs. I expect to be able to take a look at the end of this week.

mflatt:
I agree that this doesn't help for cross compilation, but I think it also doesn't get in the way. Currently, we require that a target machine type like is specified explicitly for cross compilation, as in `-m=tpb32b`.

PR #856, opened 2024-07-21
fix the whatis lines in scheme.1.in

weinholt:
Manpages have a standard format for the line that is used by the `whatis` command and lintian was complaining that Chez's pages did not use the right format. This commit fixes those lines.


mflatt:
Thanks!

PR #858, opened 2024-08-03
improve support for lists in cptypes

mflatt:
Move `list` form the normalptr slot to the multiplet slot. In particular, this allows the reduction of

```
  (lambda (x f)
    (unless (list-assuming-immutable? x)
      (f)
      (list-assuming-immutable? x)))
```

`pair`s are split in `list-pairs` and `nonlist-pairs`. A `pair` may go from one classification to the other, so the internal representations of the `list?` predicate use both of them or neither.

### DIFF
--- a/s/cptypes-lattice.ss
+++ b/s/cptypes-lattice.ss
@@ -68,6 +68,8 @@
    flzero-pred
    $fixmediate-pred
    $list-pred ; immutable lists
+   list-pair-pred
+   pair-pred
    box-pred
    vector*-pred
    vector-pred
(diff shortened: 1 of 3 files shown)


mflatt:
This is an old commit by @gus-massa that I missed when merging Racket changes toward version 10. I discovered the missing commit when double-checking for differences between Racket's current copy of Chez Scheme and this repo.

I think maybe I was concerned at the time with issue discussed in racket/racket#4610, and meant to set it aside temporarily when merging. But as the discussion there says, this commit doesn't introduce the issue, and I believe it's an improvement on its own.

s/cptypes-lattice.ss
```diff
@@ -285,9 +288,13 @@
     (define flvector*-mask             #b0100000000000000)
     (define box-mask                   #b1000000000000000)
 
+    ;These two are more tricky, becuase they are not a constant property.
```
burgerrg, line 291: These two are trickier, because they are not constant properties.

mflatt: Updated - thanks

burgerrg (approved):

Issue #810, opened 2024-02-28
`makefiles/boot.zuo`'s `main` calls `cross-build-boot` incorrectly


LiberalArtist:
In https://github.com/cisco/ChezScheme/commit/10d59cc04a12469f26caa22af72f63e84a8d97b3#diff-8adebbb27a9540dc754a57375e77d019b00c55580b6574f5ec1b89d58a3e17af, `cross-build-boot` gained an `xpatch?` argument, but `boot.zuo`'s `main` submodule is still calling it with 10 arguments instead of 11:

https://github.com/cisco/ChezScheme/blob/57f92bb76aed694437a2e201780654e9c03a576f/makefiles/boot.zuo#L14-L20

In the scenarios I've tried, it's worked to just add `#t` after `'("all")`, but I'm not sure if that's correct in general.

mflatt:
Should this `main` submodule be here at all? I can't remember the reason for `boot.zuo` to be used directly. I wonder whether it was experiment that turned out not to be useful — or maybe it's useful, and I forgot how it's used.

LiberalArtist:
According to https://racket.discourse.group/t/chez-for-architectures-without-native-backends/950/15, I'd read about it in:

https://github.com/cisco/ChezScheme/blob/2b39576bd751fc60edb486f94fe91742a61751b7/BUILDING#L85-L93

The link in my post was to Racket's `release` branch, and thus is broken, but it looks like you added it in https://github.com/cisco/ChezScheme/commit/5cf80b7a03f033c31b0166f154e7ce7d0c3e4eb0 the day after you wrote the comment above the one I linked to, and you removed that text from the document (but did not change `boot.zuo`) in https://github.com/cisco/ChezScheme/commit/1152c60ef9fd484f02b01ee74d385cb5810208b0.

It fits especially well with the use in Guix, when, having already built our native `chez-scheme` package (and having discarded its build directory), we subsequently want to build bootfiles for some other machine type. The way the Guix packaging is set organized (at least, the best way I've found to organize it so far), we first build a Guix package containing bootfiles, then supply that as an input to the Guix package that does the rest of the Chez Scheme build. That lets us share the package definition with relatively few conditionals and overrides among native and cross-compiled (with `make kernel`) packages that might have gotten their bootfiles through various different strategies: `rktboot`, `build.zuo`, formerly pre-built bootfiles that we couldn't bootstrap, and, as of earlier today in my fork, `re.boot` for Chez 10.

[...]

PR #862, opened 2024-08-10
makefiles/boot.zuo: remove broken main submodule

LiberalArtist:
The main submodule hasn't worked since at least commit 10d59cc, and documented build modes have been added in the mean time that support its former uses.

Closes https://github.com/cisco/ChezScheme/issues/810

----

(The change to Guix to stop using `boot.zuo` directly has been live since April.)


mflatt:
Thanks!

PR #864, opened 2024-08-19
add {,t}{a6,arm64}ios machine types

Bogdanp:
In support of https://github.com/racket/racket/pull/5057


jltaylor-us:
Why does Chez Scheme need to distinguish between OSX and iOS?  I don't see any code in this PR that treats them any differently.

Bogdanp:
@jltaylor-us the intent of the change is to let user code distinguish between a Chez Scheme build for macOS and one for iOS, to make it possible for the same code to support both platforms in the places that they differ. For example, from the Racket PR I linked:

* https://github.com/racket/racket/blob/df580ac929c70438fd894cb94db493bb31309b40/racket/src/cs/rumble/system.ss#L39-L42 uses the `machine-type` to determine the current OS and 
* https://github.com/racket/racket/blob/df580ac929c70438fd894cb94db493bb31309b40/racket/collects/openssl/libcrypto.rkt#L89-L106 then uses that information to load libcrypto in an iOS-specific way.

Bogdanp:
As far as I understand, yes. But, I think so would the objects produced by the `tarm64fb` and `tarm64nb` builds of Chez, yet we still have those machine types so that we can distinguish between FreeBSD and NetBSD builds (but maybe I'm missing something).

jltaylor-us:
hmm... I'm not sure why the different BSDs have different machine types either. (but maybe because they are distinct kernels, whereas iOS and MacOS still share the same XNU kernel, I think?)
In any case, if iOS is going to have its own machine types then they need to be documented in the release notes, and `configure` needs to be updated to detect when it should produce those machine types.  Unless you can't actually build them with the normal build process after `configure`, in which case there should be some instructions somewhere (presumably `BUILDING`) on how to build those machine types.

Bogdanp:
@jltaylor-us I've added some docs. I wasn't sure how much detail to go into in the release notes re. iOS support, so let me know if you think more is needed. 

Building for iOS requires cross-compilation so I've opted to just update `BUILDING` instead of changing `configure`.

jltaylor-us:
Seems ok to me.  I'm not sure I understand the practical applications of iOS support if it only works when attached to a debugger, but I assume _someone_ has a use for it or you wouldn't be bothering with this PR.

Let's give it a few days to see if any of the other maintainers have anything to add.

Bogdanp:
Looks like there haven't been any new comments in the last couple of weeks so maybe it's time to merge this?

jltaylor-us:
Thanks for this contribution!

Issue #861, opened 2024-08-07
Mysterious invalid syntax when building transformer output with `quote-syntax`


dpk:
```scheme
(library (quote-syntax-bug)
  (export my-if yes-or-no)
  (import (chezscheme))

  (define-syntax my-if
    (lambda (stx)
      (syntax-case stx ()
        ((_ test then else)
         (list (quote-syntax if) #'test #'then #'else)))))

  (define-syntax yes-or-no
    (lambda (stx)
      (syntax-case stx ()
        ((_ what)
         (my-if (syntax->datum #'what) #''yes #''no))))))
```

```
> (import (quote-syntax-bug))
Exception: invalid syntax if at line 9, char 30 of quote-syntax-bug.sls
Type (debug) to enter the debugger.
```

Changing `quote-syntax` to `syntax` in the definition of `my-if` fixes it. Commenting out `yes-or-no` and using `my-if` at the REPL fails. But pasting the definition of `my-if` into the REPL and invoking it directly from the REPL works.

mnieper:
I am investigating this issue.  The root cause of the problem seems to be the behaviour of the tests in this program:

```scheme
(import (chezscheme))
(define qif (quote-syntax if))
(assert (symbol? (syntax->datum qif)))
(assert (identifier? qif))
```

The first test succeeds, the second test fails.

mnieper:
Okay, this was an easy-to-find mistake. The issue is fixed with the following patch:

```
diff --git a/s/syntax.ss b/s/syntax.ss
index 197d1b10..9e845119 100644
--- a/s/syntax.ss
+++ b/s/syntax.ss
@@ -6027,10 +6027,11 @@
          (_ (syntax-error (source-wrap e w ae))))))

 (global-extend 'core 'quote-syntax
-   (lambda (e r w ae)
+  (lambda (e r w ae)
+    (let ([e (source-wrap e w ae)])
       (syntax-case e ()
-         ((_ e) (build-data no-source (source-wrap (syntax e) w ae)))
-         (_ (syntax-error (source-wrap e w ae))))))
+         ((_ e) (build-data no-source (syntax e)))
+         (_ (syntax-error e))))))

 (global-extend 'core 'syntax
   (let ()
```

I am going to submit a formal pull request.

LiberalArtist:
This has been completed in https://github.com/cisco/ChezScheme/commit/613e127ba4ad5f6a0d03043c6f6688b5bf0d6145 and can be closed.

PR #867, opened 2024-08-26
fix incorrect wrapping in quote-syntax

mnieper:
This fixes issue #861.

The mistake was an incorrect application of a wrap.

### DIFF
--- a/mats/8.ms
+++ b/mats/8.ms
@@ -12031,6 +12031,12 @@
    (equal? (syntax->datum (quote-syntax (a b c))) '(a b c))
    (equal? (syntax->datum (quote-syntax ...)) '...)
    (identifier? (quote-syntax if))
+   (let* ([sfd (make-source-file-descriptor "src" (open-bytevector-input-port '#vu8()))]
+          [src (make-source-object sfd 1 2)])
+     (identifier? (eval
+                   (make-annotation `(quote-syntax ,(make-annotation 'if src 'if))
+                                    src
+                                    '(quote-syntax if)))))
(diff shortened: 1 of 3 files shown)


burgerrg:
Thank you for finding and fixing this bug! Please add a subsection to the release notes under section:bugfixes for this.

burgerrg:
Did you intend to update the zlib & zuo submodules? I'd prefer to see those kinds of updates in separate pull requests.

mnieper:
> Did you intend to update the zlib & zuo submodules?

I didn't.  Thanks for catching this.

mnieper:
> Thank you for finding and fixing this bug! Please add a subsection to the release notes under section:bugfixes for this.

I missed that.  There is now a subsection about the bug in my patch.

burgerrg:
The fix to the submodule update didn't resolve it, so I took the liberty of rewriting history on your branch to eliminate the submodule updates in the first place.

mflatt:
When I add just the test here, unpatched Chez Scheme still passes. Maybe I'm confused, but I think it's because the bug is related to annotations, and `.ms` files are loaded in that way that doesn't get them. If that's right, here's a test that explicitly constructs and annotation fails for me before and passes after.

```

   (let ([src (make-source-object
               (make-source-file-descriptor "src" (open-bytevector-input-port '#vu8()))
               1
               2)])
     (identifier?
      (eval
       (make-annotation `(quote-syntax
                          ,(make-annotation
                            'if
                            src
                            'if))
                        src
                        '(quote-syntax if)))))
```

Issue #869, opened 2024-09-10
FFI: Add a stdbool type for C99 _Bool/C23 bool


williewillus:
I misread the FFI manual, and mistakenly used the FFI `boolean` type to represent a C99 `_Bool`.

Since Chez Scheme predates C99 being common, `boolean` in the FFI actually just means `int`, which was what was commonly used before C99 `_Bool`. This means I was reading 3 bytes of garbage every time a C function returned a bool, causing things to evaluate to #t on the scheme side when they shouldn't.

Now that it's 2024 and C99 _Bool is pretty widely used, and C23 is even finally reserving the keyword `bool` for the type, so I think the FFI should add `stdbool` to explicitly represent this type on the Chez side. Currently, I have to assume that `sizeof(bool)==1` on my platform, use `unsigned-8`, and do the zero-check myself. Although it's rare that someone would make their platform bools larger than 1 byte, hardcoding the size and having to do the check myself is still a minor annoyance.

williewillus:
Closed by #873, thanks!

PR #873, opened 2024-09-20
add `stdbool` foreign type

mflatt:
As suggested in #869.

The `stdbool` type corresponds to `bool` as defined by the host machine's "stdbool.h", which is 1 byte for all currently supported platforms. That's different than `boolean`, which instead reflects a traditional C encoding of booleans as `int`, which is 4 bytes for all currently supported platforms.

The "stdbool.h" header is standard as of C99, while Chez Scheme's implementation generally sticks to C89. But the header needed only for a test file, and it tends to be available in C89 enviornments, too.


release_notes/release_notes.stex
```diff
@@ -124,6 +124,11 @@ iOS compilation target. Native-code support for iOS is currently limited
 to applications attached to a debugger as the platform does not allow
 executable code to be loaded at runtime.
 
+\subsection{Standard boolean foreign type (10.1.0)}
+
+The \scheme{stdbool} foregin type corresponds to \scheme{bool} as
```
jltaylor-us, line 129:
```suggestion
The \scheme{stdbool} foreign type corresponds to \scheme{bool} as
```

s/cmacros.ss
```diff
@@ -419,7 +419,7 @@
   rv64nb    trv64nb
   la64le    tla64le
 )
-
+<
```
jltaylor-us, line 422: stray character?

jltaylor-us (commented):
Seems ok to me, but I only understand about half of it.

release_notes/release_notes.stex
mflatt: Thanks!

s/cmacros.ss
mflatt: Yes, thanks. (Because I use M-< and M-> to jump to the start/end of the buffer in Emacs, I'm constantly mistyping and immediately jumping away from the mistake, leaving stray `<`s and `>`s in files. Probably I should learn a new keybinding.)

PR #879, opened 2024-10-13
workaround Clang v15 AArch64 miscompile that affects parallel collection

mflatt:
This patch avoids a miscompile using Clang v15 on macOS. The default compiler on macOS was recently upgraded to Clang v16, which appears to fix the problem, and I have not been able to replicate the problem with Clang v15 variants that are available in Linux distributions. So, it might be ok to just ignore the problem. But since v15 installations are likely to hang around for a while in other macOS installations, since the workaround is simple, since Racket users who build themselves are affected, and since I spent *a lot* of time tracking down the problem, I'm inclined to include a workaround.

For details on the miscompile at it affects Chez Scheme, see [clang15-miscompile.zip](https://github.com/user-attachments/files/17355942/clang15-miscompile.zip).

### DIFF
--- a/c/gc.c
+++ b/c/gc.c
@@ -383,6 +383,11 @@ static ptr sweep_from;
 
 static int in_parallel_sweepers = 0;
 
+#ifdef USE_PAR_SWEEPERS_WORKAROUND
+FORCEINLINE int get_in_parallel_sweepers() { return in_parallel_sweepers; }
+# define PAR_SWEEPERS_WORKAROUND() int in_parallel_sweepers = get_in_parallel_sweepers()
+#endif
+
 #define HAS_SWEEPER_WRT(t_tc, tc) 1
 
 # define GC_MUTEX_ACQUIRE() alloc_mutex_acquire()
@@ -460,6 +465,10 @@ static void sweep_dirty(thread_gc *tgc);
 # define PARALLEL_UNUSED    /* empty */
 # define NO_PARALLEL_UNUSED UNUSED
 
+#endif
+
+#ifndef PAR_SWEEPERS_WORKAROUND
+# define PAR_SWEEPERS_WORKAROUND() do { } while (0)
 #endif
 
 #define SWEEP_NO_CHANGE        0
--- a/c/version.h
+++ b/c/version.h
@@ -61,6 +61,10 @@
 # define BIG_ENDIAN_IEEE_DOUBLE
 #endif
 
+#if defined(__arm64__) && defined(__clang__) && (__clang_major__ == 15)
+# define USE_PAR_SWEEPERS_WORKAROUND
+#endif
+
 #ifdef PORTABLE_BYTECODE
 # undef FLUSHCACHE
 # ifdef PORTABLE_BYTECODE_BIGENDIAN
--- a/mats/thread.ms
+++ b/mats/thread.ms
@@ -1637,6 +1637,59 @@
       ok?))
 )
 
+(mat parallel-gc-regression
+  (equal?
+   ;; Create threads that refer to data owned by other threads, and try to
+   ;; mix things up to see if we can provoke a problem in parallel collection.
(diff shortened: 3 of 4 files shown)


mflatt:
I spent so long tracking this down that I'd like to tell you the long story, even though it doesn't really matter. The miscompile seems like a run-of-the-mill compiler error, but the way it affected Chez Scheme and Racket made it especially difficult to find.

During 2022-2024, I've tried off and on to track down an occasional failure in Racket builds on my macOS M1/M2 laptops. Memory would get mangled late in the build — specifically during documentation rendering for he "math" library, which uses libgmp and libmpfr in multi-threaded mode. Since the problem never happened on x86_64, and since it only happened during parallel documentation rendering, I was pretty sure that I was looking for some sort of race condition exposed by AArch64's weak memory coherence.

Although I discovered that I could provoke a crash by just rebuilding documentation, even that step takes 10 minutes, and the crash would only happen rarely, so getting a crash would take hours. Any little change I made to try to gather information would make the crash go away or become much more difficult to provoke, so hours turned to days.

Meanwhile, users of the Racket main distribution were not running into problems, which I chalked up to the fact that documentation is pre-rendered. Also, maybe more generally libgmp or libmpfr needed to be involved, so maybe it wasn't my problem. In any case, the lack of reports made the problem feel less of an emergency than I would normally consider crashing bugs, especially since I had so much trouble replicating the crash or pinpointing an issue. So, I'd burn a day or three on the issue every few months.

In September 2024, I finally gathered evidence to suspect that the problem was in the GC's parallel mode. And with that suspicion, I was finally able to make a small Chez Scheme program with the right ingredients to crash, showing that the problem was independent of Racket and math libraries. The big difference was being able to provoke a crash within seconds instead of hours, and I found the problem over the next day.

In retrospect, it's clear why the problem was so difficult to find. I was pretty sure I was looking for a memory race, but that turned out to be because only multi-threaded programs could reach the miscompiled code. And only during parallel collections. And only when the collector is looking at specific words within a thread representing virtual registers, which are not something that programs normally use directly. The effect of the miscompile was that a "does this object belong to me?" check would succeed when it shouldn't. That matters only when a thread has an object in its virtual register that was allocated by a different thread, which is an even more rare use of a virtual register. And even when it goes wrong, there's only a small chance that different collector threads will end up looking at the same object at the same time, and even concurrent traversal of the same object will turn out ok a lot of the time! Finally, and most perniciously, the miscompile creates a race that isn't in the source code, and in a code template that is put in place by a macro that is used dozens of times in the output (and compiled ok in all other other instances).

Meanwhile, Racket distributions are compiled with Clang v12, which is why it hasn't been a problem for Racket users, even when they run programs with parallelism.

jltaylor-us (approved):
Great detective work, Matthew!

maoif:
Thanks for fix and sharing your experience of tracking down this tricky bug.

glandium:
```
#if defined(__arm64__) && defined(__clang__) && (__clang_major__ == 15)
```
FYI, `__clang_major__` from the clang provided by Apple on macos/Xcode does *not* match the upstream clang version's `__clang_major__`. For some reason, Apple decided clang's version was Xcode's. But it doesn't match the LLVM version it's derived from. Xcode 15's clang is based on LLVM 16.
So, `__clang_major__ == 15` matches entirely different versions of the compiler on clang versions that don't come from Xcode/Apple. You may want to add `defined(__apple_build_version__)`

jryans:
Xcode 15 was the first version that included a [new linker](https://developer.apple.com/documentation/xcode-release-notes/xcode-15-release-notes#Linking), perhaps these issues were related to bugs in that new codebase... A few linking issues were fixed in [15.1](https://developer.apple.com/documentation/xcode-release-notes/xcode-15_1-release-notes#Linking). There may also be bug fixes that didn't make it into the release notes.

PR #880, opened 2024-10-25
repair thread test for `make test-more`

mflatt:
Repair the test so that it works when the maximum collect generation is set to 1, and add a `test-more` case to the CI configuration.

Also, refine a workaround preprocessor check to enable it only when `defined(__apple_build_version__)`.


mflatt:
Includes the suggestion from @glandium in #879.

jltaylor-us (approved):

PR #881, opened 2024-10-25
Add missing Smake_flvector implementation

payneca:
The header file scheme.h defines `Smake_flvector`, but Chez Scheme does not provide an implementation. C code can use the header, but calling `Smake_flvector` results in an error trying to reference an undefined symbol.

### DIFF
--- a/csug/foreign.stex
+++ b/csug/foreign.stex
@@ -3579,14 +3579,16 @@ that it takes a C string (character pointer) as input.
 \end{flushleft}
 
 \noindent
-\scheme{Smake_string}, \scheme{Smake_vector}, \scheme{Smake_bytevector},
-and \scheme{Smake_fxvector} are similar to their Scheme counterparts.
+\scheme{Smake_string}, \scheme{Smake_vector},
+\scheme{Smake_bytevector}, \scheme{Smake_fxvector}, and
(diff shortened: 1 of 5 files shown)


csug/foreign.stex
```diff
@@ -3579,14 +3579,16 @@ that it takes a C string (character pointer) as input.
...
+\scheme{Smake_string}, \scheme{Smake_vector},
+\scheme{Smake_bytevector}, \scheme{Smake_fxvector}, and
+\scheme{Smake_flvector} are similar to their Scheme counterparts.
 
 \begin{flushleft}
 \cfunction{ptr}{Smake_string}{iptr \var{n}, int \var{c}}
 \cfunction{ptr}{Smake_vector}{iptr \var{n}, ptr \var{obj}}
 \cfunction{ptr}{Smake_bytevector}{iptr \var{n}, int \var{fill}}
 \cfunction{ptr}{Smake_fxvector}{iptr \var{n}, ptr \var{fixnum}}
+\cfunction{ptr}{Smake_flvector}{iptr \var{n}, ptr \var{flonum}}
```
mflatt, line 3591: `ptr` here should be `double`, right?

payneca: Yes. Good catch. Pushed the update.

mflatt:
Thanks!

Issue #885, opened 2024-11-09
Keys and values of nested `with-continuation-mark` calls are evaluated before the continuation marks are updated


AlgebraicWolf:
It seems that currently when several nested `with-continuation-mark` calls are being applied to the same continuation, keys and values are evaluated before any continuation marks are set. This causes trouble if, for any reason, the evaluation of some key or value depends on a mark being set in the same continuation.

# Steps to reproduce

Try evaluating the following expression:

```chez
(with-continuation-mark 'a 42
    (with-continuation-mark 'b (continuation-marks-first (current-continuation-marks) 'a)
      (continuation-marks-first (current-continuation-marks) 'b)))
```

# Expected behaviour

The outermost `with-continuation-mark` set the the mark `'a -> 42`. After that, the evaluation of arguments for the innermost `with-continuation-mark` occurs. There is nothing to evaluate in `'b`, so the `(continuation-marks-first (current-continuation-marks) 'a)` gets evaluated to `42`, since this is the value currently associated with the key `'a`, and the mark `'b -> 42' is set. Finally, the expression ` (continuation-marks-first (current-continuation-marks) 'b)` is evaluated to `42`.

# Observed behaviour

The expression evaluates to `#f`.

Moreover, if the nested `with-continuation-mark` call is evaluated in the separated continuation, the behaviour matches the expected one. For example, the following expression:

```chez
(with-continuation-mark 'a 42
  (values
    (with-continuation-mark 'b (continuation-marks-first (current-continuation-marks) 'a)
      (continuation-marks-first (current-continuation-marks) 'b))))
``` 

is evaluated to `42`. If one tries to set the value in other continuation, prior to making nested `with-continuation-mark` calls, the older value is obtained. For example,

```chez
(with-continuation-mark 'a 0
  (values
    (with-continuation-mark 'a 42
      (with-continuation-mark 'b (continuation-marks-first (current-continuation-marks) 'a)
        (continuation-marks-first (current-continuation-marks) 'b)))))
```

gets evaluated to `0` instead of `42`. This highlights that the problem is not with the continuation marks in general, but with the evaluation of arguments to `with-continuation-mark`.

mflatt:
Thanks for the report!

I think the problem is in the way the `with-continuation-mark` macro expands to `$call-consuming-continuation-attachment` and `$call-setting-continuation-attachment`:

```
(define-syntax with-continuation-mark
  (lambda (x)
    (syntax-case x ()
      [(_ key val body)
[...]

PR #886, opened 2024-11-09
repair `with-continuation-mark` when accessing marks in key or value

mflatt:
The expansion of `with-continuation-mark` positioned the key and value expressions outside of the current continuation marks, which meant that attempting to look up a continuation mark in the key or value expression would fail.

This is a repair for #885.

### DIFF
--- a/release_notes/release_notes.stex
+++ b/release_notes/release_notes.stex
@@ -2755,6 +2755,13 @@ in fasl files does not generally make sense.
 %-----------------------------------------------------------------------------
 \section{Bug Fixes}\label{section:bugfixes}
 
+\subsection{\scheme{with-continuation-mark} key and value expressions (10.1.0)}
+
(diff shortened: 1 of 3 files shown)


release_notes/release_notes.stex
```diff
@@ -2755,6 +2755,13 @@ in fasl files does not generally make sense.
 %-----------------------------------------------------------------------------
 \section{Bug Fixes}\label{section:bugfixes}
 
+\subsection{\scheme{with-continuation-mark} key and value expressions (10.1.0)}
+
+A bug in the \scheme{with-continuation-mark} positioned the key and
```
jltaylor-us, line 2760:
```suggestion
A bug in \scheme{with-continuation-mark} positioned the key and
```

jltaylor-us (approved):

PR #888, opened 2024-11-19
Add 1+, 1- and -1+, and improve `abs` in cptypes 

gus-massa:
Two slightly relate patchs.

**Add 1+, 1- and -1+ to cptypes**

I should have added them in the (old) previous commit that added `add1` and `sub1`, but I forget these primitives existed. I refactored some of the test to avoid too much duplication.

**Improve fixnum case of abs in cptypes**

In the (old) previous commit I added a reduction for `(abs <flnumber>)` to `(flabs <flnumber>)`. But it has no reduction for `(abs <fxnumber>)`. 

Now I'm reducing `(abs <fx>)` to

    (let ([t <fx>])
      (if (fx= t (most-negative-fixnum))
          (pariah (- (most-negative-fixnum)))
          (fxabs t)))

that is the straightforward expansion. It's expanded later in `cpprim` to  

    (let ([t <fx>])
      (if (fx= t (most-negative-fixnum))
          (- (most-negative-fixnum))
          (if (fx>= t 0) t (fx- t))))

But perhaps it's better to expand it in `cptypes` directly to

    (let ([t <fx>])
      (if (fx>= t 0)
          t
          (if (fx= t (most-negative-fixnum))
              (- (most-negative-fixnum))
              (fx- t))))

becuse it is faster for positive numbers and slower for `(most-negative-fixnum)` that is very unusual.

I also wanted to add a similar special case for `add1` and `sub1`, but in their expansion in `cpprim` they use the `overflow` flag, so expanding them to a comparison to `(most-negative-fixnum)` or `(most-positive-fixnum)` is probably slower. (I hope to try something like this next month.)


burgerrg:
Thank you for the updates! Please add a subsection to the release notes describing them.

gus-massa:
I fixed one misplace parenthesis that was causing an error in one of the tests.

I added a note for each commit. It is it fine or it's better to accumulate all the changes in a single note? 

But I did't add this text in the notes to avoid repetition. I can add it if you think it's necessary.

> Type recovery is enabled with the \scheme{enable-type-recovery} parameter.

release_notes/release_notes.stex
```diff
@@ -116,6 +116,14 @@ Online versions of both books can be found at
...
 \section{Functionality Changes}\label{section:functionality}
 
+\subsection{Improved support abs in type recovery (10.2.0)}
+
+The type recovery pass has improved support for abs when the argument is a fixnum.
+
+\subsection{Add support for 1+, 1- and -1+ in type recovery (10.2.0)}
+
+The type recovery pass has support for 1+, 1- and -1+.
+
```
burgerrg, line 126: I'd combine these into a single entry, something like:
```
\subsection{Type recovery improvements (10.2.0)}

The type recovery pass has improved support for \scheme{abs} with a fixnum argument
and added support for \scheme{1+}, \scheme{1-}, and \scheme{-1+}.
```

burgerrg: It's fine not to add the sentence about enabling type recovery.

gus-massa:
Done

s/cptypes.ss
```diff
@@ -298,10 +298,38 @@ Notes:
...
+                 [e* e*] [r* r*])
+        (cond
+          [(and (null? e*) (null? r*))
+           (values (reverse rev-rbefore*) (reverse rev-rvar*) (reverse rev-re*) (reverse rev-rref*))]
+          [(check-constant-is? (car r*))
+           (loop (cons (car e*) rev-rbefore*) rev-rvar* rev-re* (cons (car r*) rev-rref*)
+                 (cdr e*) (cdr r*))]
+          [(try-ref->prelex/not-assigned (car e*))
+           => (lambda (v)
+                (set-prelex-multiply-referenced! v #t) ; just in case it was sinlge referenced
```
burgerrg, line 321: sinlge => singly

s/cptypes.ss
```diff
@@ -298,10 +298,38 @@ Notes:
...
-          (if (null? var*)
-              body
-              `(call ,(make-preinfo-call) ,(build-lambda var* body) ,e* ...)))
+
+    (define (prepare-let e* r*) ; ==> (before* var* e* ref*)
+      ; The arguments e* and r* must have the same length.
+      ; In the results:
+      ;   before*, var* and e* may be shorter than the arguments.
+      ;   var* and e* have the same length.
+      ;   ref* has the same lenght than the arguments.
```
burgerrg, line 307: lenght than => length as

mats/cptypes.ms
```diff
@@ -880,97 +893,66 @@
...
   (not (cptypes-equivalent-expansion?
          '(lambda (x) (when (fixnum? x)
                         (fixnum? (sub1 x))))
          '(lambda (x) (when (fixnum? x)
                         #t))))
-  (not (cptypes-equivalent-expansion?
-         '(lambda (x) (when (bignum? x)
-                        (bignum? (sub1 x))))
-         '(lambda (x) (when (bignum? x)
-                        #t))))
```
mflatt, line 931: Should this test have been kept? I don't think it's covered by the `test-closed1` generalization, but I might be missing something else.

burgerrg: No, because when x = (+ (most-positive-fixnum) 1), it's incorrect.

gus-massa: I undeleted the test. It's a to ensure I don't make an easy mistake in the future. I don't know why I deleted it.

It's not included in `test-closed1`. Moreover, I was thinking about writing `test-not-closed1` to reduce duplication, but I decided that it was too weird. But Perhaps I'll add it in the future if there are too many cases of almost-closed functions.

s/cptypes.ss
gus-massa: I fixed both typos and also a "lenght" somewhere else.

PR #899, opened 2024-12-30
repair `fx-/wraparound` with `0` first argument, allow one argument

mflatt:
The `fx-/wraparound` function was treated like R6RS `fx-` by cp0, which meant that it could reduce a 2-argument call with `0` as the first argument to a 1-argument call, even though `fx-/wraparound` was defined with only a 2-argument form. Making `fx-/wraparound` more like R6RS `fx-` seems like the path of least resistance and least likelihood of future bugs.


burgerrg (approved):
Thanks for this fix!

PR #904, opened 2025-01-21
Simplify guard form using call-in-continuation

mnieper:
Chez Scheme 10's new call-in-continuation procedure allows implementing guard in a simpler and likely more efficient way.

### DIFF
--- a/s/7.ss
+++ b/s/7.ss
@@ -1,12 +1,12 @@
 ;;; 7.ss
 ;;; Copyright 1984-2017 Cisco Systems, Inc.
-;;; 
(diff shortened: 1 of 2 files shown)


burgerrg:
Good find! Would `$reset-protect` in 7.ss also benefit from this simplification?

mnieper:
> Good find! Would `$reset-protect` in 7.ss also benefit from using call-in-continuation?

The same simplification should be applicable here. I am going to update the pull request after running a local test.

burgerrg (approved):

PR #906, opened 2025-01-24
Treats # as a delimited in R6RS mode when reading numbers (fixes #687)

### LINKED ISSUE #687: The R6RS reader does not treat # as a delimiter for numbers
In the lexical syntax of R6RS, the `#` character is a delimiter (the rationale for this is, I think, that this excludes comments within identifiers and numbers).  In particular, the input `3#f` should be tokenized into the lexemes `3` and `#f`.  A transcript of the Chez Scheme REPL, however, shows the following:

```
> (let ([p (open-string-input-port "#!r6rs 3#f")])
     (get-datum p))
Exception in get-datum: 3#f symbol syntax is not allowed in #!r6rs mode at char 7 of #<input port string>
```

mnieper:


jltaylor-us:
Looks ok to me.  Probably needs an entry in release notes.

mnieper:
> Looks ok to me. Probably needs an entry in release notes.

Thank you for taking a look at it; I made an entry in the release notes.

mflatt (approved):

Issue #585, opened 2021-10-07
reproducible builds


attila-lendvai:
chez compilation doesn't seem to produce the same output when invoked multiple times on the same file.

even the size of the result is changing:

```
-rwxr-xr-x 1 alendvai users 5699005 Oct  8 01:44 idris2.so
-rwxr-xr-x 1 alendvai users 5699020 Oct  8 01:21 idris2.so.1
-rwxr-xr-x 1 alendvai users 5699012 Oct  8 01:35 idris2.so.2
-rwxr-xr-x 1 alendvai users 5699004 Oct  8 01:36 idris2.so.3
```

i'm willing to look into it, but first i'd like to ask whether there was any effort made to achieve reproducible builds?

dybvig:
If developers are willing to use $fasl-file-equal? (which should probably be surfaced in any case as a user-level primitive) rather than 'cmp' to compare binary output, that should work.  In effect, $fasl-file-equal? attempts to unify the two files with respect to generated names, so it is the same as straight binary identity with each gensym in one file replaced by the same gensym in the other file.

LiberalArtist:
I have some interest in this because I've contributed to the Chez Scheme package for [Guix](https://arxiv.org/pdf/1305.4584.pdf) (a user-level package manager and full system distribution), which benefits greatly from byte-for-byte reproducible build to enable its "purely functional" features.

If I understand correctly, the issue (or at least part of it) is that Chez Scheme uses UUIDs for gensyms, including in generative record types. In particular, at least since <https://github.com/cisco/ChezScheme/commit/2fd3db68230d094a0d396348a8140a4d3693b120>/<https://github.com/cisco/ChezScheme/issues/338>, these are fully random "type 4" UUIDs.

I wonder if "type 5" UUIDs (RFC 4122 §4.3), which are deterministically generated based on a name in some conceptual "namespace", could be useful for making byte-for-byte equivalent object files. I'm not sure I fully understand all the nuances of Chez Scheme's use of gensyms, but I could imagine, for example, a tool that would rewrite object files to give the gensym's predictable names (perhaps incorporating a hash of the source file or something to help prevent collisions).

LiberalArtist:
I've looked at this a bit more, and I think I can see a strategy for gensyms work with reproducible builds.

As I understand this code:

https://github.com/cisco/ChezScheme/blob/8c5f5cc27cf8f2b432aafc4f2fc640435294b74a/s/5_7.ss#L87-L129

[...]

PR #909, opened 2025-01-31
Symbol generation with lazily generated names that behave like standard symbols

mnieper:
This patch implements the `(generate-symbol [pretty-name])` procedure. As `(gensym)`, it returns a symbol with a lazily generated unique name; however, the resulting symbol is not a gensym but an ordinary symbol.

Such a procedure is useful in situations where it is expected that symbols are identical if and only if their names (as returned by `symbol->string`) are spelled the same.

In particular, this is the case for R6RS code. Strictly speaking, by section 11.10 of the R6RS, gensyms are non-conformant. This patch improves R6RS compatibility by providing an R6RS version of `generate-temporaries` where the symbols are generated by `generate-symbol` and not by `gensym`. Moreover, ordinary symbols have write/read invariance even when only standard lexical syntax is allowed.

For the context of this patch, see also [SRFI 260](https://srfi.schemers.org/srfi-260/srfi-260.html).

### DIFF
--- a/c/intern.c
+++ b/c/intern.c
@@ -1,12 +1,12 @@
 /* intern.c
  * Copyright 1984-2017 Cisco Systems, Inc.
- * 
+ *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
  * You may obtain a copy of the License at
- * 
+ *
  * http://www.apache.org/licenses/LICENSE-2.0
- * 
+ *
  * Unless required by applicable law or agreed to in writing, software
  * distributed under the License is distributed on an "AS IS" BASIS,
  * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -151,6 +151,15 @@ ptr S_mkstring(const string_char *s, iptr n) {
   return mkstring(s, n);
 }
 
+static ptr real_symname(ptr sym) {
+  ptr name = SYMNAME(sym);
+  if (!GENSYMP(sym))
+    return name;
+  if (Scdr(name) != Strue)
+    return Sfalse;
+  return Scar(name);
+}
+
 /* handles single-byte characters, implicit length */
 ptr S_intern(const unsigned char *s) {
   iptr n = strlen((const char *)s);
@@ -164,8 +173,8 @@ ptr S_intern(const unsigned char *s) {
   b = S_G.oblist[idx];
   while (b != NULL) {
     sym = b->sym;
-    if (!GENSYMP(sym)) {
-       ptr str = SYMNAME(sym);
+    ptr str = real_symname(sym);
+    if (str != Sfalse) {
        if (Sstring_length(str) == n) {
           iptr i;
           for (i = 0; ; i += 1) {
@@ -203,8 +212,8 @@ ptr S_intern_sc(const string_char *name, iptr n, ptr name_str) {
   b = S_G.oblist[idx];
   while (b != NULL) {
     sym = b->sym;
-    if (!GENSYMP(sym)) {
-       ptr str = SYMNAME(sym);
+    ptr str = real_symname(sym);
+    if (str != Sfalse) {
        if (Sstring_length(str) == n) {
           iptr i;
           for (i = 0; ; i += 1) {
--- a/csug/objects.stex
+++ b/csug/objects.stex
@@ -1,11 +1,11 @@
 % Copyright 2005-2017 Cisco Systems, Inc.
-% 
+%
 % Licensed under the Apache License, Version 2.0 (the "License");
 % you may not use this file except in compliance with the License.
 % You may obtain a copy of the License at
-% 
+%
 % http://www.apache.org/licenses/LICENSE-2.0
-% 
+%
 % Unless required by applicable law or agreed to in writing, software
 % distributed under the License is distributed on an "AS IS" BASIS,
 % WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
@@ -71,7 +71,7 @@ This predicate is not defined by the Revised$^6$ Report, but should be.
 \var{n} must be an exact nonnegative integer less than or equal to
 the length of \var{list}.
 
-\scheme{list-head} and the standard Scheme procedure \scheme{list-tail} 
+\scheme{list-head} and the standard Scheme procedure \scheme{list-tail}
 may be used together to split a list into two separate lists.
 While \scheme{list-tail} performs no allocation but instead returns a
 sublist of the original list, \scheme{list-head} always returns a copy
@@ -581,10 +581,10 @@ source before the operation began.
 
 (string-copy! s1 3 s2 1 3)
 s2 ;=> "-bol------"
- 
+
 (string-copy! s1 7 s2 4 2)
 s2 ;=> "-bolly----"
- 
+
 (string-copy! s2 2 s2 5 4)
 s2 ;=> "-bollolly-"
 \endschemedisplay
@@ -943,7 +943,7 @@ In the case of \scheme{immutable-vector-copy}, \scheme{immutable-vector-append},
 \formdef{self-evaluating-vectors}{\categorythreadparameter}{self-evaluating-vectors}
 \listlibraries
 \endentryheader
-    
+
 \noindent
 The default value of this parameter is \scheme{#f}, meaning that vector literals must be quoted, as
 required by the Revised$^6$ Report.
@@ -1154,7 +1154,7 @@ as \var{fxvector}.
 \endschemedisplay
 
 
-    
+
 \section{Flonum-Only Vectors\label{SECTFLVECTORS}}
 
 \index{flvectors}%
@@ -1352,7 +1352,7 @@ as \var{flvector}.
 \endschemedisplay
 
 
-    
+
 \section{Bytevectors\label{SECTBYTEVECTORS}}
 
 As with vectors, {\ChezScheme} extends the syntax of bytevectors to allow
@@ -1403,7 +1403,7 @@ A negative fill value is treated as its two's complement equivalent.
 
 The values in the returned list are exact eight-bit signed integers,
 i.e., values in the range -128 to 127 inclusive.
-\scheme{bytevector->s8-list} is similar to the Revised$^6$ Report 
+\scheme{bytevector->s8-list} is similar to the Revised$^6$ Report
 \scheme{bytevector->u8-list} except the values in the returned list
 are signed rather than unsigned.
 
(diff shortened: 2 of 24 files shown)


mflatt:
This idea seems ok to me. Some thoughts on the implementation here:

 * It looks like there's a problem with `symbol-hash` on a generated symbol before interning is forced, as in `(symbol-hash (generate-symbol))`.
 *  It seems a little awkward to have two representations of a "plain" symbol: a string as the name or `(cons <string> #t)` after a generated symbol is forced. That could be addressed with a small change to `S_uninterned` and how it is used (illustrated by the alternative linked below).
 * The inlined form of `symbol->string` ends up being quite a lot of code. Maybe `symbol->string` isn't used often enough for it to matter, but I tried an alternate encoding to make it simpler.
 * The code for `generate-temporaries` and for `buildmake-symbol` seems unnecessarily duplicated.

For your consideration, here's what I'd do: https://github.com/mflatt/ChezScheme/tree/srfi-260. I was aiming to simplify the encoding so that the `hash` field is effectively used to indicate whether a gensym or generated symbol is reified, and the `name` field has the `symbol->string` result in a more consistent place. The encoding as a combination of `name` and `hash` didn't end up as nice as I hoped it would, but I like it a little better. Even if you don't like the alternate encoding, some tests and other changes there may be useful.

FWIW (but you may already know this): Another possible approach to `generate-temporaries` is to generate the scope part of the name, instead of the symbol part. Racket uses that strategy so that the symbol part can be deterministic, which is part of a general effort toward deterministic builds.

mnieper:
Thank you for this thorough review and the improvements, @mflatt. I went through them and incorporated them in my pull request. I simply forgot about checking for `symbol-hash`, and fixing the representation for uninterned symbols at the same time is great.

Due to the improvements, I could apply two further simplifications.

You added the "discard" flag back on `string->symbol`. It has always been on `gensym->unique-string`. Is there a good reason why these two procedures should behave differently?

It seems unfortunate that `symbol-hash` forces interning a "generated symbol", which can make them more costly than uninterned symbols. As a way out, `symbol-hash` could generate a random hash for yet uninterned generated symbols, which would then be incorporated into the generated name. The procedure extracting hash values from symbol names would need a special case that detects names constructed in such a way and calculates, in this case, the hash just by extracting it. What do you think?

> FWIW (but you may already know this): Another possible approach to generate-temporaries is to generate the scope part of the name, instead of the symbol part. Racket uses that strategy so that the symbol part can be deterministic, which is part of a general effort toward deterministic builds.

This is how I would implement them if there were no historical baggage. As R6RS does not have `identifier-hash` or `identifier<?`, the symbol part needs to be sufficiently diverse (definitely not constant), as otherwise algorithms using hashtables to map identifiers (e.g. to model environments) behave badly. So we would at least need a symbol counter (so that the symbols can be named, say, `tmp0`, `tmp1`, ...). Moreover, the symbols would then likely be interned by `generate-temporaries` , which needs locking (as would the symbol counter). All in all, I think the current implementation is the simplest one for CS.

mflatt:
> You added the "discard" flag back on `string->symbol`. It has always been on `gensym->unique-string`. Is there a good reason why these two procedures should behave differently?

I think it would be better for `gensym->unique-string` to have a `discard` flag. If a program calls `gensym->unique-string` but doesn't use the result, then it would be nice for the call to go away. But there must be some function that has an effect; `symbol-hash` relies on `gensym->unique-string`  as producing the effect, and there may be other places that rely on it.

Meanwhile, `symbol->string` is used much more widely and seems more likely to end up in a place where its result is unused, so it seems worth keeping the current potential optimization. The side effect can be deferred to `$gensym->pretty-name` or some other internal function.

> As a way out, symbol-hash could generate a random hash for yet uninterned generated symbols, which would then be incorporated into the generated name.

I don't immediately see how this would work. My understanding so far is that the interaction below is meant to reliably produce `#t`, since the same name is written after quote as is printed:

```
> (define s (generate-symbol))
> s
g-nx26tlsbk3mm7fw2h7u1btx9x
> (eq? s 'g-nx26tlsbk3mm7fw2h7u1btx9x)
#t
```

To make sure that `'g-nx26tlsbk3mm7fw2h7u1btx9x` finds the same symbol, the quoted symbol needs the same hash code. How wold it find that? Would interning need to check a table of generated names just before settling on a hash code for a symbol? What would happen if I fasl `s` to a file, restart Chez Scheme, `(define s2 'g-nx26tlsbk3mm7fw2h7u1btx9x)`, and then load `s` out of the fasl file? Or do I misunderstand the intended behavior of generated symbols?

mnieper:
> > You added the "discard" flag back on `string->symbol`. It has always been on `gensym->unique-string`. Is there a good reason why these two procedures should behave differently?

(I meant "_missing_ on `gensym->unique-string", of course.)

> I think it would be better for `gensym->unique-string` to have a `discard` flag. If a program calls `gensym->unique-string` but doesn't use the result, then it would be nice for the call to go away. But there must be some function that has an effect; `symbol-hash` relies on `gensym->unique-string` as producing the effect, and there may be other places that rely on it.

Okay, but then it would make sense to have a dedicated primitive for that purpose like `$force-intern!`, which you introduced in `5_7.ms`, wouldn't it?

> > As a way out, symbol-hash could generate a random hash for yet uninterned generated symbols, which would then be incorporated into the generated name.
> 
> I don't immediately see how this would work. My understanding so far is that the interaction below is meant to reliably produce `#t`, since the same name is written after quote as is printed:
> 
> ```
> > (define s (generate-symbol))
> > s
> g-nx26tlsbk3mm7fw2h7u1btx9x
> > (eq? s 'g-nx26tlsbk3mm7fw2h7u1btx9x)
> #t
> ```
> 
> To make sure that `'g-nx26tlsbk3mm7fw2h7u1btx9x` finds the same symbol, the quoted symbol needs the same hash code. How wold it find that? Would interning need to check a table of generated names just before settling on a hash code for a symbol? What would happen if I fasl `s` to a file, restart Chez Scheme, `(define s2 'g-nx26tlsbk3mm7fw2h7u1btx9x)`, and then load `s` out of the fasl file? Or do I misunderstand the intended behavior of generated symbols?

I believe you understand the intended behaviour of generated symbols correctly. My idea is the following: If `symbol-hash` is called before the symbol is interned, a cheap hash is randomly generated without locking and recorded as part of the hash field. Let us assume for demonstration purposes that the code is the number described by `#xdeadbeef`.

When the symbol is later interned, the symbol's unique name will then be something like `g-nx26tlsbk3mm7fw2h7u1btx9x-generated-deadbeef`.  The hash function in `intern.c` detects symbol names of this form (in this example, `g-<key>-generated-hash` and will return `#xdeadbeef` instead of taking all characters into account.

When a symbol with a name like `g-nx26tlsbk3mm7fw2h7u1btx9x-generated-deadbeef` is input directly, the hash function will make no difference and likewise return `#deadbeef`.

The disadvantage of this approach is that it will be easy to forge given hash values. On the other hand, the symbol hash is not cryptographically secure anyway.

At the moment, `unique_id` is a costly operation (in `stats.c`) because it involves opening `/dev/urandom` each time it is called. Generating a random hash can be done much faster.

Maybe, if `unique_id` is replaced by a fast version (e.g. the device should be kept open and buffered so that reads only seldom have to call the OS), the advantage of computing a cheap hash may be reduced enough. I don't know whether `unique_id` can be made fast on each supported platform. It is okay if it needs locking because interning the symbol (which has to be done once a unique name is calculated) needs locking anyway.

mflatt:
> Okay, but then it would make sense to have a dedicated primitive for that purpose like `$force-intern!`, which you introduced in `5_7.ms`, wouldn't it?

Agreed. It just didn't seem worth the effort of tracking down uses of `gensym->unique-string` to replace them or to figure out whether there's some way that code outside of the implementation could somehow rely on the side effect.

> When the symbol is later interned, the symbol's unique name will then be something like `g-nx26tlsbk3mm7fw2h7u1btx9x-generated-deadbeef`. The hash function in intern.c detects symbol names of this form (in this example, `g-<key>-generated-hash` and will return `#xdeadbeef` instead of taking all characters into account.

Ah, I see now. Yes, that does seem workable.

mflatt:
Not sure I understand the state here. @mnieper, are you planning to look into the precomputed-hash idea? Or faster `unique_id`? Or were you waiting on me / something else?

mnieper:
> Not sure I understand the state here. @mnieper, are you planning to look into the precomputed-hash idea? Or faster `unique_id`? Or were you waiting on me / something else?

I am planning to look into the precomputed-hash idea because it does not involve thinking about different environments (POSIX, Windows, etc.). I'll try to get back to this soon.

LiberalArtist:
> At the moment, `unique_id` is a costly operation (in `stats.c`) because it involves opening `/dev/urandom` each time it is called. Generating a random hash can be done much faster.
> 
> Maybe, if `unique_id` is replaced by a fast version (e.g. the device should be kept open and buffered so that reads only seldom have to call the OS), the advantage of computing a cheap hash may be reduced enough.

For `gensym`, it is fine that `unique-id` is costly because it is called (at most) once per process, to initialize `$session-key`. For gensyms created during the same process, uniqueness is achieved by incrementing the process-wide [`suffix`](https://github.com/cisco/ChezScheme/blob/2f4c2ccd8ff8e4adb274cdb974884783cf7eae91/s/5_7.ss#L89), guarded by the tc mutex.

Obviously, using cryptographic-quality randomness only once increases the vulnerability of gensyms to adversarial attempts at forgery. However, that property does not seem useful enough to burden general-purpose gensyms. In exchange, we limit the gensym-related nondeterminism to `$session-key`. Adding more nondeterminism would vastly complicate the prospect of achieving reproducible builds. (See https://github.com/cisco/ChezScheme/issues/585#issuecomment-1615366186 for my thoughts on a strategy.)

I strongly suggest that `generate-temporaries` should decide that &ldquo;all practical purposes&rdquo; are consistent with using a process-wide `$session-key`. In fact, as I look at SRFI 260, the most rigorous statement of uniqueness:

> This means that for every Scheme expression `⟨expression⟩` whose (possibly non-deterministic) value does not depend on the result of `symbol->string` applied to the result *`gensym`* of a previous invocation of `generate-symbol`, its value is not equal (in the sense of `equal?`) to the value of `(symbol->string `&#x2060;*`gensym`*&#x2060;`)` for all practical purposes.

explicitly excludes inspection of &ldquo;a previous invocation of `generate-symbol`&rdquo;.

LiberalArtist:
> Adding more nondeterminism would vastly complicate the prospect of achieving reproducible builds. (See [#585 (comment)](https://github.com/cisco/ChezScheme/issues/585#issuecomment-1615366186) for my thoughts on a strategy.)

Even if reproducible builds are out of scope for this PR, it is probably worth thinking through how `generate-symbol` should interact with `$fasl-file-equal?`.

PR #912, opened 2025-02-06
Add reduction for add1/sub1 for fixnums in cptypes

gus-massa:
Reduce `(add1 <fixnum>)` to `(%3%$fxx+ x 1)`, and add equivalent reductions for similar functions. This extends  the reductions used when the argument is a flonum.

The new primitive `$fxx+` is similar to `fx+`, but in case of an overflow it calls `+` to get the correct result. This is slightly faster than using `+` directly.

(I never added a primitive, so I'm not sure that I dotted all the i and crossed all the t.)

It's split in 3 commits, but my idea is to squash them. 

* The first adds `$fxx+`, and is useful to test the speed change.

* The seconds uses it to reduce `add1`/`sub1`/... and is the interesting one

* The third adds signatures to `$+`/`$-`/... that are somewhat related and may improve slightly other parts of the optimizer.

I tested the time with the code bellow. With 100000000 repetitions, the time of `(add1 x)` changes from .84 to .75 seconds. It's faster than `(+ x 1)` because it assumes the arguments are fixnum and skip the tests. 

Initially, I was thinking about a special case for the overflow of `($fxx+ x 1)` and `($fxx+ x -1)`because they have only one possible result, but 50% I got confused about how to add bignums in `cpprim` and 50% I go lazy because it's a very rare corner case. 

Note that `(#3%fxx+ x 1)` is not as fast as `(#3%fx+ x 1)` that takes only .70 seconds, but `#3%fx+` gives the wrong result when there is an overflow.

I tried with other versions to avoid adding a primitive. It's interesting that `($+ 'add1 x 1)` is extremely slow, but I guess it's not inlining a fast cases for fixnum like all the others, so I guess it's not a problem.

```
(let repeat ([R 0])
  (unless (fx= R 3)
	(time (let loop ([N 0])
	  (unless (fx= N 100000000)
	   (loop (add1 N)))))                ;   .84 -> .75
	(time (let loop ([N 0])
	  (unless (fx= N 100000000)
	    (loop (+ N 1)))))                ;   .84
	(time (let loop ([N 0])
	  (unless (fx= N 100000000)
	    (loop (#3%fx+ N 1)))))           ;   .70 incorrect
	(time (let loop ([N 0])
	  (unless (fx= N 100000000)
	    (loop (#3%$fxx+ N 1)))))         ;   .75
	(time (let loop ([N 0])
	  (unless (fx= N 100000000)
	    (loop (let ([t (#3%$fx+? N 1)])
	            (if t t (+ N 1)))))))    ;  1.97 slow
	(time (let loop ([N 0])
	  (unless (fx= N 100000000)
	    (loop (#3%$+ 'add1 N 1)))))      ; 13.25 very slow
	(repeat (add1 R))))

```


release_notes/release_notes.stex
```diff
@@ -116,6 +116,11 @@ Online versions of both books can be found at
 %-----------------------------------------------------------------------------
 \section{Functionality Changes}\label{section:functionality}
 
+\subsection{Type recovery improvements (10.3.0)}
```
mflatt, line 119: This should be 10.2.0. (The order on version numbers puts the current "10.2.0-pre-release.2" before "10.2.0".)

mflatt (commented):
This looks generally right to me, except for a small issue with the release note.

Would it be worth adding specialization of `+` here when both arguments are known to be fixnums? I imagine that would be similarly common, or maybe more so. (And then `-` and `*`, but those would need `$fxx*` and `$fxx-`.) Or is there something about `add1` and `sub1` that makes this corner seem more worthwhile?

gus-massa:
> This looks generally right to me, except for a small issue with the release note.

I collapsed the release note with the next one that was very similar about `abs` -> `fxabs`. Is that ok? Otherwise I can revert that part.

> Would it be worth adding specialization of + here when both arguments are known to be fixnums?

I agree, but I need to think about all the cases for `+`

    (+ <flonum> <flonum>) --> (fl+ <flonum> <flonum>)  
    (+ <fixnum> <ixnum>) --> ($fxx+ <fixnum> <fixnum>)  
    (+ <flonum> <real>) --> (fl+ <flonum> ($real->flonum <real>))  
    (+ <flonum> <number>) --> cfl+ in case it's a complex?  
    (+ <flonum> <any>) --> no reduction, in case of an error 
    ...

the third may help unboxing the fist argument

mflatt (approved):

Issue #914, opened 2025-02-11
cp0 misses error case of make-record-constructor-descriptor


burgerrg:
When a record-constructor descriptor is formed, a protocol must be specified if the parent record-constructor descriptor has a protocol, as seen in the transcript below.

```
Chez Scheme Version 10.1.0
Copyright 1984-2024 Cisco Systems, Inc.

> (define A.rtd (make-record-type-descriptor 'A #f #f #f #f '#((immutable x))))
> (define A.rcd (make-record-constructor-descriptor A.rtd #f (lambda (new) (lambda () (new 1)))))
> (define B.rtd (make-record-type-descriptor 'B A.rtd #f #f #f '#((immutable y))))
> (define B.rcd (make-record-constructor-descriptor B.rtd A.rcd #f))
Exception in make-record-constructor-descriptor: no protocol specified, but parent #<record constructor descriptor> has protocol
```

The cp0 pass does not honor this restriction, but instead ignores the parent record-constructor descriptor's protocol, as seen in the transcript below:
```
> (expand/optimize '(let ()
  (define A.rtd (make-record-type-descriptor 'A #f #f #f #f '#((immutable x))))
  (define A.rcd (make-record-constructor-descriptor A.rtd #f (lambda (new) (lambda () (new 1)))))
  (define B.rtd (make-record-type-descriptor 'B A.rtd #f #f #f '#((immutable y))))
  (define B.rcd (make-record-constructor-descriptor B.rtd A.rcd #f))
  (record-constructor B.rcd)))
(let ([#{B.rtd kvydt3ia1y4xarwl4jx8txvm3-0} (#2%make-record-type-descriptor 'B
                                              (#2%make-record-type-descriptor 'A #f #f #f #f
                                                '#((immutable x)))
                                              #f #f #f '#((immutable y)))])
  (lambda (#{g0 kvydt3ia1y4xarwl4jx8txvm3-1}
           #{g1 kvydt3ia1y4xarwl4jx8txvm3-2})
    (#3%$record
      #{B.rtd kvydt3ia1y4xarwl4jx8txvm3-0}
      #{g0 kvydt3ia1y4xarwl4jx8txvm3-1}
      #{g1 kvydt3ia1y4xarwl4jx8txvm3-2})))
```

PR #915, opened 2025-02-12
fix cp0 make-record-constructor-descriptor issue (#914)

owaddell-ib:
This aims to address #914.


burgerrg (approved):
Thanks for fixing this!

burgerrg (approved):

PR #922, opened 2025-03-09
Add fxvector-copy! and flvector-copy!

williewillus:
This PR adds fxvector-copy! to provide bulk blitting operations on fxvectors.
Useful for all the same reason bytevector-copy! is useful, to avoid the overhead
of a manual loop and assignments for every single element. I have an array I need to shift and the lack of copying operations that don't copy the entire vector means I have to do it manually.

Implementation and tests are mostly stolen from bytevector-copy! and adjusted to match.
One thing I was unsure of is noted inline.

Once I have indication that I'm not going completely in the wrong direction, I'll update the docs and changelog, and will send a flvector version as well (which is probably even more useful because it guarantees no float boxing/unboxing for each element).


s/primdata.ss
```diff
@@ -1429,6 +1429,7 @@
   (fxvector [sig [(fixnum ...) -> (fxvector)]] [flags alloc cp02 safeongoodargs])
   (fxvector->list [sig [(fxvector) -> (list)]] [flags alloc safeongoodargs])
   (fxvector-copy [sig [(fxvector) -> (fxvector)]] [flags alloc safeongoodargs])
+  (fxvector-copy! [sig [(fxvector sub-length fxvector sub-length sub-length) -> (void)]] [flags true])
```
williewillus, line 1432: is there documentation on what all these flags mean?

s/5_6.ss
```diff
@@ -210,6 +210,30 @@
...
+     (unless (and (fixnum? k) (fx>= k 0))
+       ($oops who "invalid count ~s" k))
+     (unless (fx<= k (fx- n1 i1)) ; avoid overflow
+       ($oops who "index ~s + count ~s is beyond the end of ~s" i1 k fxv1))
+     (unless (fx<= k (fx- n2 i2)) ; avoid overflow
+       ($oops who "index ~s + count ~s is beyond the end of ~s" i2 k fxv2))
+     ($ptr-copy! fxv1 (fx+ (constant fxvector-data-disp)
+                           (fx* (constant ptr-bytes) i1))
+                 fxv2 (fx+ (constant fxvector-data-disp)
+                           (fx* (constant ptr-bytes) i2))
```
williewillus, line 234: the problem here is this can overflow. How should I handle that, add explicit checks? But it seems like an arbitrary limit on the user-facing side.

s/primdata.ss
gus-massa: The flags are defined in https://github.com/cisco/ChezScheme/blob/main/s/cmacros.ss#L1871 but not all of them are explained. 

`true` means that the result is always true, so 

`(if (fxvector-copy! ...) x y)`

is reduced in `cp0` to 

`(begin (fxvector-copy! ...) x)`

Now it's an automatic flag, so the `(void)` result in the signature implies this flag, so `true` is not necessary but not harmful.

s/5_6.ss
mflatt: Just use `+` and `*`? The calculations can overflow the fixnum range, but not the range allowed by `$ptr-copy!` due to the bounds checks. The overflow checks in `+` and `*` (which will not fail in practice) seem like a negligible cost, especially relative to the bounds guards.

mflatt:
This seems ok to me, so I encourage you to go ahead with `flvector-copy!`, docs, and release notes.

williewillus:
Add docs, changelog entry, and flvector versions.

csug/objects.stex
```diff
@@ -1351,7 +1381,37 @@ as \var{flvector}.
...
+
+\noindent
+\var{src} and \var{dst} must be flvectors.
+\var{src-start}, \var{dst-start}, and \var{n} must be exact nonnegative
+integers.
+The sum of \var{src-start} and \var{n} must not exceed the length of \var{src},
+and the sum of \var{dst-start} and \var{n} must not exceed the length of \var{dst}.
 
+\scheme{flvector-copy!} overwrites the \var{n} elements of \var{dst}
+starting at \var{dst-start} with the \var{n} elements of \var{dst}
```
schemer666, line 1399:
```suggestion
starting at \var{dst-start} with the \var{n} elements of \var{src}
```
I think it should be `src`. The same goes for `fxvector-copy!` on line **1171**

williewillus: Good catch. That looks like a mistake in the original string-copy! doc I copied this from, will send another pr to fix that

PR #924, opened 2025-03-23
Fix comment on ARM register r18

soegaard:
The comment said r18 is `caller-save` but the Apple docs says register r18 is a callee-save one. The exact words are [1]:

> The platforms reserve register x18. Don’t use this register.

The Arm docs [2] call  r18 (x18) a platform register. I interpret this as the each platform ABI can decide how to use it on their own. That is, Apple's restriction follows the Arm spec.

Digging a little more: Apple uses r18 for Rosetta. 

Since register r18 is missing from the `define-registers` form, it isn't in use. Potentially, `r18` could be used on Linux.

[1]
https://developer.apple.com/documentation/xcode/writing-arm64-code-for-apple-platforms [2]
https://developer.arm.com/documentation/102374/0102/Procedure-Call-Standard

### DIFF
--- a/s/arm64.ss
+++ b/s/arm64.ss
@@ -5,8 +5,8 @@
 ;;;  Register usage:
 ;;;   r0-r7: C argument/result registers, caller-save
 ;;;   r8: indirect-result register, caller-save
-;;;   r9-18: caller-save
-;;;   r19-28: callee-save
+;;;   r9-17: caller-save
+;;;   r18-28: callee-save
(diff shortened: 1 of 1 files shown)


mflatt:
I expected this edit to say something about "r18" as "reserved on some platforms" instead of "callee-saved". Would it make sense to add an "r18"-specific line between the "r9-17" and "r19-28" lines?

soegaard:
Closed due to #925.

I meant to change this PR, but the github interface resulted in a new PR.

PR #925, opened 2025-03-23
Elaboration on r18

soegaard:
Added a bit more information in the comment.


soegaard:
The Windows docs [1] has the following to say:

>  x18 |  Reserved platform register: in kernel mode, points to KPCR for the current processor; In user mode, points to TEB

Note: TEB stands for "Thread Environment Block (TEB) is a structure that stores per-thread information."

[1] 
https://learn.microsoft.com/en-us/cpp/build/arm64-windows-abi-conventions?view=msvc-170

mflatt (approved):

mflatt:
Thanks!

PR #926, opened 2025-03-24
Add fxvector-copy!

williewillus:


williewillus:
woops, sapling somehow made a new pr instead of updating the old one

PR #930, opened 2025-04-02
Windows console mouse events translation:

dzoep:
- enlarge the static buffer of s_ee_read_char to 20 as the generated sequences are often longer than the current size (10)
- extract dwMouseEvent from the buffer
- generate basic mouse motion event with coordinates wrapped at 1000
- keep old button state in a static variable
- decode left, right and middle button press and release
- support MOUSE_WHEELED and convert to virtual button presses

This PR adds support for translating MOUSE_EVENT_RECORD[1] structures in windows console input event buffer to VT console sequences as generated by most modern terminal emulators in the SGR mouse reporting mode[2] (see the SET_SGR_EXT_MODE_MOUSE). The mechanism is the same as employed by the code already present on windows terminals which do not support VT sequences for cursor keys (see the code just above the main section added in expeditor.c. The static buffer had to be extended to allow storing strings which may be as long as "\e[<99;999;999m".

By default this does not inject any character sequences into the input because the windows console must be programmatically configured to send them. For an example of how this can be achieved after setting the terminal into "raw mode", the GetConsoleMode/SetConsoleMode has to be used.

See also relevant Racket PR [3].

[1] https://learn.microsoft.com/en-us/windows/console/mouse-event-record-str
[2] https://www.invisible-island.net/xterm/ctlseqs/ctlseqs.html#h2-Mouse-Tracking
[3] https://github.com/racket/racket/pull/5243

### DIFF
--- a/c/expeditor.c
+++ b/c/expeditor.c
@@ -92,7 +92,7 @@ static ptr s_ee_read_char(IBOOL blockp) {
   ptr tc;
 #endif /* PTHREADS */
   BOOL succ;
-  static wchar_t buf[10];
+  static wchar_t buf[20];
   static int bufidx = 0;
   static int buflen = 0;
   static int rptcnt = 0;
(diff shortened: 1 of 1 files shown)


mflatt:
Looks good to me, at least in that I detect no problems, and I have tried building with the Racket variant of this PR at racket/racket#5243. I'm not in a position to confirm that mouse support works as intended, but I know that @dzoep has been working on this for a while and is putting it to real use.

mflatt (approved):

Issue #928, opened 2025-03-27
`system` calls on aarch64 darwin fail opaquely with llvm-{18,19}


tkerber:
This may be related to #896, although I cannot tell.

Empirically, calls to the `system` procedure (for instance `(system "echo foo")`) fail on chez scheme 10.1.0 and 10.0.0 when compiled with llvm 18 / 19, but not with llvm 17. I have not observed this on aarch64 linux. Here's a sample repl transcript:

```
$ scheme
Chez Scheme Version 10.1.0
Copyright 1984-2024 Cisco Systems, Inc.

> (system "echo foo bar")
Exception in system: Resource temporarily unavailable
Type (debug) to enter the debugger.
> 
```

This was tested through nix, and currently affects the latest scheme obtainable on macos on nixos, though I'll make a PR to patch this downstream.

PR #931, opened 2025-04-02
change `TARGET_OS_IPHONE` preprocessor handling

### LINKED ISSUE #896: build fails with `invalid memory reference` on x64 darwin and llvm-{18,19}
when compiling chez scheme `10.1.0` on x64 darwin the build fails with an invalid memory reference exception  with llvm-{18,19} -- (llvm-{16,17} builds fine and aarch64 builds fine will all compiler versions)

```
compiling s/interpret.ss with output to ta6osx/s/interpret.ta6osx
failed
 in build-one
 in loop
 in module->hash
make: *** [Makefile:8: build] Error 1
```
https://cache.nixos.org/log/a35n1jjyzg77bi0x6v3z2sbf29yxga1r-chez-scheme-10.1.0.drv

mflatt:
Recent versions of the build tools on macOS define `TARGET_OS_IPHONE` as 0 when not compiling for iOS.

Closes #896 and #928

### DIFF
--- a/c/version.h
+++ b/c/version.h
@@ -311,9 +311,10 @@ typedef int tputsputcchar;
 #define MMAP_HEAP
 #define IEEE_DOUBLE
 /* for both iPhone and iPhoneSimulator */
-#if defined(TARGET_OS_IPHONE)
+#if defined(TARGET_OS_IPHONE) && TARGET_OS_IPHONE
 # define SYSTEM(s) ((void)s, -1)
 # define WRITE_XOR_EXECUTE_CODE
+# define S_TARGET_OS_IPHONE
 #endif
 #if defined(__arm64__)
 # if !defined(WRITE_XOR_EXECUTE_CODE)
--- a/BUILDING
+++ b/BUILDING
@@ -568,7 +568,8 @@ To cross compile for iOS, provide an iOS SDK path and define the
  ./configure \
    --cross \
    -m=tarm64ios \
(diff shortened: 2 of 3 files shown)


c/version.h
```diff
@@ -311,7 +311,7 @@ typedef int tputsputcchar;
 #define MMAP_HEAP
 #define IEEE_DOUBLE
 /* for both iPhone and iPhoneSimulator */
-#if defined(TARGET_OS_IPHONE)
+#if TARGET_OS_IPHONE
```
Bogdanp, line 314: When I build this patch on my machine with clang 16, I get warnings like:

```
gcc -DPORTABLE_BYTECODE -Ipb/boot/pb -Ipb/c -Ic/ -Ipb/zlib -Ipb/lz4/lib -arch arm64 -O2 -Wpointer-arith -Wall -Wextra -Wno-implicit-fallthrough -o pb/c/pb.o -c c/pb.c
In file included from c/pb.c:1:
In file included from c/system.h:29:
c/version.h:314:5: warning: 'TARGET_OS_IPHONE' is not defined, evaluates to 0 [-Wundef-prefix=TARGET_OS_]
  314 | #if TARGET_OS_IPHONE
      |     ^
1 warning generated.
```

Maybe it would be better to go with `#if defined(TARGET_OS_IPHONE) && TARGET_OS_IPHONE` instead?

mflatt: Ugh, ok. I actually tried looking at documentation to make sure that `-Wall` and `-Wextra` do not imply `-Wundef`, and I ran some experiments to check my conclusion. But I had never heard of `-Wundef-prefix`, and it didn't occur to me that I would need to experiment with a name that starts `TARGET_OS_`. Thanks again, Apple.

PR #936, opened 2025-04-23
Flush stdout when printing the version.

### LINKED ISSUE #935: scheme --version prints to stderr
This makes it clumsy to use from shell scripts and in %() in RPM spec files:
```
sh-5.2$ v=$(scheme --version)
10.1.0
sh-5.2$ echo $v

sh-5.2$
```

Needs to remember the syntax for how to redirect from stderr to stdout:
```
sh-5.2$ v2=$(scheme --version 2>&1)
sh-5.2$ echo $v2
10.1.0
sh-5.2$
```

samth:
Fixes #935.

### DIFF
--- a/c/main.c
+++ b/c/main.c
@@ -281,7 +281,8 @@ int main(int argc, const char *argv[]) {
       } else if (strcmp(arg,"--verbose") == 0) {
         Sset_verbose(1);     
       } else if (strcmp(arg,"--version") == 0) {
-        fprintf(stderr,"%s\n", VERSION);
+        fprintf(stdout,"%s\n", VERSION);
+        fflush(stdout);
         exit(0);
       } else {
         argv[new_argc++] = arg;


burgerrg:
Thanks! Please also update the "--help" case to use stdout.

burgerrg:
Please also update the release notes, because current users may be depending on the current stderr behavior.

hegjon:
For my own learning; are there platforms that does not flush the streams on exit? 

`$ man 3 exit #On Fedora Linux 41`
 > All open stdio(3) streams are flushed and closed.  Files created by tmpfile(3) are removed.

jltaylor-us:
> For my own learning; are there platforms that does not flush the streams on exit?

It's part of the ISO and IEEE C standards

> The exit() function shall then flush all open streams with unwritten buffered data, close all open streams...

and I found that behavior documented all the way back in the original 1978 K&R _The C Programming Language_.

To be pedantic, "when calling `exit`" is not exactly the same thing as "on exit", but I assume you intended to ask about the former.

PR #942, opened 2025-05-01
Avoid invalid memory reference when 1st arg of `path-build` is ""

Jarhmander:
Avoid invalid memory reference when 1st arg of `path-build` is ""

Before:
```
> (path-build "" "a")
```
After:
```
> (path-build "" "a")
"a"
```

Code was simplified a bit at some places to exploit the new changes, but without changing existing behavior.


burgerrg:
Thanks for finding this! There's another definition of path-build in s/reboot.ss that may need to be updated as well.

Jarhmander:
> Thanks for finding this! There's another definition of path-build in s/reboot.ss that may need to be updated as well.

Thanks!

For the occurrence in s/reboot.ss, I didn't fully understand why this (re?)definition is here, and I just assumed that this script worked correctly in normal circumstances, and since this `path-build` function is not exported, it could be left alone. I nevertheless modified it to act like the one in 6.ss.

burgerrg:
I think reboot.ss is loaded by older versions of Chez Scheme, so it's good to have it updated.

Please also add a subsection in the release notes under bugs fixed about this. Thanks!

release_notes/release_notes.stex
```diff
@@ -2873,6 +2873,12 @@ An incorrect approach to finding the current executable's path on
 NetBSD meant that boot files could not be found relative to the
 executable.
 
+\subsection{Prevent \scheme{path-build} from doing invalid memory accesses (10.2.0)}
+
+\scheme{path-build} mades invalid memory refences when first argument
+(\scheme{dir-path}) is the empty string.
+It now instead returns the second argument unmodified (\scheme{path}).
+
```
burgerrg, line 2881:
```
\subsection{Invalid memory access in \scheme{path-build} (10.2.0)}

When the \scheme{dir-path} argument is the empty string, the \scheme{path-build}
procedure now returns the \scheme{path} argument instead of performing an invalid memory access.

File: mats/6.ms [hunk 2]
```diff
@@ -3332,6 +3332,7 @@
   (error? (path-last 'a/b/c))
   (error? (path-root 'a/b/c))
   (error? (path-extension 'a/b/c))
+  (error? (path-build 'a 'b))
```
burgerrg, line 3335: Add a check for both strings:
```
(error? (path-build 'a "fn"))
(error? (path-build "dir" 'b))
```

mats/6.ms
```diff
@@ -3409,6 +3410,20 @@
...
+  (if (windows?)
+      (equal? (path-build ".\\" "b.ss") ".\\b.ss")
+      (equal? (path-build "./" "b.ss") "./b.ss"))
+  (if (windows?)
+      (equal? (path-build "foo" "b.ss") "foo\\b.ss")
+      (equal? (path-build "foo" "b.ss") "foo/b.ss"))
+  (if (windows?)
+      (equal? (path-build "quux\\" "b.ss") "quux\\b.ss")
+      (equal? (path-build "quux/" "b.ss") "quux/b.ss"))
+
```
burgerrg, line 3426: Simplify:
```
(equal? (path-build "" "fn") "fn")
(equal? (path-build "dir" "fn") (format "dir~cfn" (directory-separator)))
(equal? (path-build (format "dir~c" (directory-separator)) "fn") (format "dir~cfn" (directory-separator)))
```

burgerrg (changes requested):

mats/6.ms
Jarhmander: I thought of doing that, but in a sense, I find it a bit less readable. Anyway, just did what you suggested.

mats/6.ms
burgerrg: Thanks, and please remove the (path-build 'a 'b) case.

release_notes/release_notes.stex
Jarhmander: This wording is much better indeed. Took it.

mats/6.ms
Jarhmander: Checks added.

burgerrg (approved):
Thank you!

PR #948, opened 2025-06-22
Allow setting history file location from environment

CosmicToast:
A CLI flag isn't very convenient for modifying history file locations, since it needs to be specified in many contexts, some of which are difficult to actually modify the argument list for.
It's common practice (less, node, psql, python, sqlite… the list goes on) to allow setting the history file location from an environment variable named something like `PROJECT_HISTORY`.
When set, it overrides the default when nothing else is specified, but is still overriden by a CLI argument.
This implements this feature in Chez.

I haven't tested this yet (I might get to it somewhat later), and I'm relatively unfamiliar with sTeX (changes to which I also haven't tested so far). I figured I would open this "early" to allow for any relevant comments to be made in the interim, though I don't expect this change to be controversial.

The expected UX is that a user that wants all of their history files in a centralized location (e.g. I use ~/.local/state/history/*) shall set a bunch of environment variables in their login shell, and from that point on won't have to think about it anymore.

The final questions remain with respect to the release notes and tests.
How exactly should I test this? Where in the release notes should this be mentioned?
This part of the contributing document wasn't entirely clear, besides that it should be done.
I'm perfectly willing to, but would rather have a little bit of direction as to what's expected (while I do other things and later make sure the documentation changes actually look correct) :)

### DIFF
--- a/c/main.c
+++ b/c/main.c
@@ -99,7 +99,9 @@ int main(int argc, const char *argv[]) {
   int exists_timestamp_mode = 0;
 #ifdef FEATURE_EXPEDITOR
   int expeditor_enable = 1;
-  const char *expeditor_history_file = "";  /* use "" for default location */
+  /* use "" for default location */
+  const char *expeditor_history_file = getenv("CHEZSCHEME_HISTORY");
+  if (!expeditor_history_file) expeditor_history_file = "";
 #endif /* FEATURE_EXPEDITOR */
 
   if (strcmp(Skernel_version(), VERSION) != 0) {
--- a/csug/use.stex
+++ b/csug/use.stex
@@ -347,6 +347,8 @@ been redirected, or if the \scheme{--eedisable} command-line option
 (Section~\ref{SECTUSECOMMANDLINE}) has been used.
 The history is saved across sessions, by default, in the file
 ``.chezscheme\_history'' in the user's home directory.
+If the \scheme{CHEZSCHEME_HISTORY} environment variable is set during startup,
+its value becomes the default history file location.
 The \scheme{--eehistory} command-line option
 (Section~\ref{SECTUSECOMMANDLINE}) can be used to specify a different
 location for the history file or to disable the saving and restoring of
--- a/release_notes/release_notes.stex
+++ b/release_notes/release_notes.stex
@@ -116,6 +116,13 @@ Online versions of both books can be found at
 %-----------------------------------------------------------------------------
 \section{Functionality Changes}\label{section:functionality}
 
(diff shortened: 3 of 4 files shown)


mflatt:
This looks like a good idea to me. I tried out the patch, and it does seem to work as intended.

How about `CHEZSCHEME_HISTORY` instead of `CHEZ_HISTORY`? Existing precedents here are mostly "Scheme" instead of "Chez", such as the bootfile environment variable `SCHEMEHEAPDIRS`. But I think `CHEZSCHEME_HISTORY` is usefully more specific than `SCHEME_HISTORY`, and the distributions that I checked call the executable `chezscheme` instead of `scheme`.

The documentation update in "use.stex" currently says `CHEZ_SCHEME` where `CHEZ_HISTORY` was intended. But, of course, I'm suggesting `CHEZSCHEME_HISTORY`, instead.

burgerrg:
I also think it's a good idea and prefer `CHEZSCHEME_HISTORY`. Would you please update the C code to call getenv once, and then write a section in the release notes?

CosmicToast:
I still haven't had the time to take a look at the things I said I would, but by sheer coincidence I did have a couple of minutes, so I've acted on the feedback here :)

P.S. gave github a chance to resolve the fact that the release notes needed a rebase, it immediately did a merge of main into a feature branch, so I guess I won't be using that feature either.

burgerrg (approved):

LiberalArtist:
I like this approach much better than the `--eehistory` CLI flag: thanks!

> The expected UX is that a user that wants all of their history files in a centralized location (e.g. I use ~/.local/state/history/*) shall set a bunch of environment variables in their login shell, and from that point on won't have to think about it anymore.

I note that the path you mentioned is in the default `$XDG_STATE_DIR`. Some time ago, I had proposed following the XDG specification for the history file: https://github.com/cisco/ChezScheme/pull/703 At the time, maintainers weren't interested, but I wonder if they'd be open to reconsidering? If the next release of Chez Scheme will contain new options for the history file, this seems like a good time to consider it.

While `CHEZSCHEME_HISTORY` is a big improvement over `--eehistory`, cooperating with the interoperable configuration mechanism would be a complementary enhancement. As I wrote in that PR:

> It's a collective action problem: from the perspective of an individual tool, it's easy to argue that one should just use tool-specific configuration options, but much of the benefit as a user comes from being able to manage the scores of tools one has installed in a consistent manner.  

I'd be glad to rework the patch adding XDG support on top of this change, if maintainers are open to the approach. As I wrote there, it can be done without disrupting the use of `~/.chezscheme_history` for those who prefer it.

PR #950, opened 2025-06-26
improve `expt` on a flonum to a large exact power

mflatt:
Advice and implementation provided by @gambiteer via racket/racket#5228.

I'm not sure about my approach to testing here. In principle, `expt` should produce a accurate result, but the result relies on `pow` from the C library, and the quality of that function varies. On GitHub CI, even `a6osx` doesn't produce a accurate result, even though it does on machines that I have access to — at least, I get a precise result with the changes in `prim5.c`. Overall, maybe there is a better approach.

### DIFF
--- a/c/prim5.c
+++ b/c/prim5.c
@@ -1463,13 +1463,16 @@ static double s_pow(double x, double y) {
   } else
     return pow(x, y);
 }
-#elif defined(MACOSX)
-/* intel macosx delivers precise results for integer inputs, e.g.,
- * 10.0^21.0, only with long double version of pow */
+#elif (machine_type == machine_type_i3osx || machine_type == machine_type_ti3osx)
+/* intel macosx delivers accurate results for integer inputs, e.g.,
+ * 10.0^21.0, only with long double version of pow; it's not clear
+ * whether that has been true for x86_64 as opposed to i386, but as
+ * of macOS 10.15 (Catalina), pow seems ok for x86_64, while powl
+ * is less accurate as of macOS 13; so, we're using powl only for i3osx */
 static double s_pow(double x, double y) { return powl(x, y); }
-#else /* i3fb/ti3fb */
-static double s_pow(double x, double y) { return pow(x, y); }
-#endif /* i3fb/ti3fb */
+#else /* i3fb/ti3fb/i3osx/ti3osx */
+static double s_pow(double x, double y) { return powl(x, y); }
+#endif /* i3fb/ti3fb/i3osx/ti3osx */
 
 #ifdef __MINGW32__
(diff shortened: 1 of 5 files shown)


jltaylor-us:
I think you're mixing up terms for accuracy and precision in your comment.  It's been a long time since my numerical analysis class, but I'm pretty sure the error propagation rules say that there are _no_ significant digits left after using 2^56 as an exponent for a value that only has 2^53 bits of precision to start with.

I agree that we'd like to be as accurate as is feasible, but this whole issue is poorly motivated.  Flonums are, by definition, inexact values, and the computation being performed on them magnifies the "inexactness" (i.e., the range of real values that might be represented by the inexact number) so that it's larger than the result.  The operation simply doesn't make sense.

On the other hand, if we want to _pretend_ that we're doing an operation on the value represented exactly by the IEEE double precision floating point number then the test needs to _use a number with an exact floating point representation_.  The difference in the mathematical result of `1.000000000000001^(2^56)` and `(4503599627370501/4503599627370496)^(2^56)` (i.e., the exact value represented by the double-precision float) is two or three orders of magnitude.

As an aside, I suspect that the difference you're seeing in the C library implementations may have something to do with whether the underlying hardware (or VM in the case of GitHub CI) supports 80-bit extended precision (which I think is an Intel-only thing).

jltaylor-us:
I think I confused myself into thinking that the particular number in the test was something other than an arbitrary number close enough to 1.0 that the result isn't +inf.  (Specifically I was thinking that it was supposed to be 1+ε but it isn't.)  But something is still nagging at me about this test.  I think maybe that it's just a bunch of "magic" numbers with no indication of _why_ we expect those to be the correct values.  They definitely aren't the _mathematically_ correct values (i.e., if input and computed with infinite precision) – not least of which is because the input has no exact flonum representation, and the rounding error there will compound until it's larger than the result.  Perhaps that concern can be "solved" with a short comment like "these values are computed by decomposing the calculation like _blah_ and computing in 80-bit extended precision" or similar.

This also illustrates my other concern, which has very little to do with the _accuracy_ of the computation.  The numbers involved in this example are so large that any difference between the exact value stored in the flonum and the mathematical quantity that the user _thought_ they were operating on will be magnified so much that the computation may well be nonsense.  But I suppose it isn't our job to make sure the user understands error propagation and what that means in terms of what conclusions they can draw from the result of one of these calculations.

In summary, I agree that this is an improvement, and we absolutely should accept it, but I am concerned that it may give users an unwarranted idea of how much they can rely on the results of these calculations with huge exponents unless they are _very_ sure about the inputs.

jltaylor-us (approved):

PR #953, opened 2025-06-30
correct `powl` that was intended to be `pow`

mflatt:
Despite helpful reviews, I managed to mangle part of #950.

An intended change was to use `pow` instead of `powl` for x86_64 and AArch64 macOS. The mangled change caused `powl` to be used even more broadly, instead. Also, simplify the comment, since extra caveats seem related to experiments with a similarly incorrect change, and adjust the test, since results on x86_64 macOS should be reliably accurate.

Noticed by @gambiteer by the time it was merged to Racket: https://github.com/racket/racket/commit/de67e3372038316b956b958d1f9864a406d20d01#r161150380

### DIFF
--- a/mats/5_3.ms
+++ b/mats/5_3.ms
@@ -2723,11 +2723,11 @@
         ;; The accuracy of `expt` ultimately depends on the C
         ;; library's `pow`, and platforms vary. The `fl~=` predicate
         ;; reflects that variation for most purposes, but `pow` can
-        ;; var more. Enumerate a few platforms where we trust `pow` to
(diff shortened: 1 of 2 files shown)


mats/5_3.ms
```diff
@@ -2727,7 +2727,7 @@
         ;; be good, and increase the fuzz for other platforms.
```
burgerrg, line 2727: Line 2726: `var more` => `vary more`.

burgerrg (approved):
Good find!

mats/5_3.ms
mflatt: Fixed - thanks!

PR #954, opened 2025-07-03
Add vector-copy!.

maoif:
This commit adds `vector-copy!`, in the same spirit of `fxvector-copy!` and `flvector-copy!`.

### DIFF
--- a/csug/objects.stex
+++ b/csug/objects.stex
@@ -792,6 +792,38 @@ themselves are not copied.
 
 %----------------------------------------------------------------------------
 \entryheader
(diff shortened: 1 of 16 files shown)


mflatt:
This looks ok to me.

My only hesitation is that there's not a clear performance benefit to having this function built in (since it cannot be turned into a memove like the related operations). But `vector-copy!` seems useful in general, and it does fit with the others.

I also considered whether `$vector-copy!` should be used in internally, but since `$vector-copy!` would need another offset argument to be useful here, I concluded that having separate loops is a fine choice.

PR #959, opened 2025-07-23
Add reductions for + in cptypes

gus-massa:
This version is working and reduces a lot of cases of `+`, in particular to make `(+ x 1)` as fast as `(add1 x)` in some useful common cases. It can be merged after fixing some typos, but I wish to add a similar reduction for `-` to keep the balance of the universe.

The main idea is that if `x` is a flonum, then `(#3%fl+ x 1.0)` is 4 or 6 times faster than `(+ x 1)` even if we add some corner cases for `0`, `0.0` and `-0.0`.  I found too many corner cases, so I wish to share this as an advanced draft to get feedback. 

I'll update this PR with a version for `-` next week, probably, because I'm not sure how many unexpected corner case I'll find. (The version for `*` is quite different, I'll wait a few months until attempting it. And `/` has even more cases!)    

----

Reduce `+` with some combinations of values that are known at compile time to be real

    (+ <fx> <fx> ...) => ($fxx+ <fx> <fx> ...)
    (+ <fl> <fx> ...) => (fl+ <fl> (fixnum->flonum <fx>) ...)
    (+ <fl> <real> ...) => (fl+ <fl> (real->flonum <real>) ...)
    (+ <fl> <fl> ...) => (fl+ <fl> <fl>)

with some special cases for `0`, in particular

    (+ <fl> 1 ...) => (fl+ <fl> 1.0 ...)
    (+ <fl> 0 ...) => (fl+ <fl> -0.0 ...)
    (+ 1.0 0 ...) => (fl+ 1.0 0.0 ...)

----

In particular, I actually had to reduce 

    (+ <fl> <fx>) => (fl+ <fl> (if (eq? <fx> 0) -0,0 (fixnum->flonum <fx>)))

in case the first argument is `-0.0`. (This case was pointed out by @mflatt.)

    (+ -0.0 0) => -0,0 
    (+ -0.0 0.0) => 0,0 
    (+ -0.0 -0.0) => -0,0 

I fount no version of `(fixnum->flonum <fx>)` that return `-0.0`, and it looks like it's implemented using an assembler function with the same behavior, so the only solution was to use an `if`. The code tries to avoid adding this check when possible, in particular when the second argument is never `0` or the first argument is never `-0.0`.

Also, some reductions apply only when the left associativity is enabled.


gus-massa:
A few minor changes, probably not worth reviewing until the final version. In particular a rebase because in a previous PR the internal representation of fixnums changed from `'fixnum` to something like `(union 'non-zero-fixnum 0)`. Also, after reading the docs and looking here and there, I think that the only way to get a `-0.0` in a `+` is something like `(+ 0 -0.0 -0.0 0 -0.0)` so I'm skipping the conversion to `-0.0` when one of the parameters is neither `0` nor `-0.0`.

---

A few days ago I [asked in HN](https://news.ycombinator.com/item?id=44767692) in case someone there realizes there is another example. Then [sparkie@HN](https://news.ycombinator.com/user?id=sparkie) replied with an example in C that used `FE_DOWNWARD` https://godbolt.org/z/5qvqsdh9P . 

Later, the reply explains how to use `vfixupimm` to change only `0.0` to `-0.0` without branches. I don't know enough about the later pass of Chez Scheme to be sure how to use `vfixupimm` and IIUC it's not available in all microprocessors, so I'll not use it. Anyway, it looks like an interesting instruction because it can make a lot of other specific changes to flonums without branching.

gus-massa:
I'm optimistic, so I removed [WIP] from the title. 

I updated the version. Is that ok?

---

Some minor remarks:

* I'm using `partial-fold-minus`, so some expressions like 

```
    (#3%$fxx- (most-negative-fixnum) 7)
````

are not reduced, in spite 

```
    (- (most-negative-fixnum) 7)
```

is constant folded to a bignum. I may try to fix this, but `partial-fold-minus` has already too many options and I don't want to add even more. `#3%$fxx-` is strange because the arguments are restricted to fixnums, but the result is not restricted.

* Fixing the problem, I noticed a lot of cases of

```
    (- <positive-fixnum>  <positive-fixnum>)
```

that are reduced to 
 
```
    (#3%$fxx- <positive-fixnum>  <positive-fixnum>)
```

but a smart enough compiler  could be reduced to 

```
    (#3%fx- <positive-fixnum>  <positive-fixnum>)
```

that avoids the overflow check, so it's slightly faster.

The problem is that cptypes does not distinguish positive and negative fixnums. I'm adding this to my wish list.

mflatt (approved):

PR #960, opened 2025-07-28
Add a few improvements for predicates in cptype, in particular for zero?

gus-massa:
**Allow #f and (void) as results of primitive handlers**

The new specialization handlers can return `#f` or `(void)` in case they don't want to make changes, and then the default is used.

When I wrote this initially a few years ago, I was afraid of doing too much work in the `ntypes` variable and then calling the default handler and that this could break the `O(Nlog^?(N))` runtime bound, but now I consider that it's not a problem in real cases, so I think it's better to allow it. The use of `#f` is similar to `define-inline` in `cp0`, but `(void)` allows me to use `when`or `cond` to select the useful cases. So instead of

    [(x) (and (some-check? x)
              (do-something x)))

I can write

    [(x) (when (some-check? x)
           (do-something x)))

and similarly, I can use `cond` instead of a lot of mixed `or`s and `and`s.

---

**Improve predicates handling**

When a primitive is marked as a predicate and has a handler, then try to reduce it as a predicate and then call the handler for more specific reductions.

Previously if it has both a `[pred ???]` mark and a `cptypes` flag in `primdata.s` then the handler was ignored. This is confusing (even for me that wrote it), and if someone ever want to use the information in `[pred ???]` in another part of the compiler and add more `[pred ???]` cases, it would surprisingly break the handler in `cptypes`. This is fixed with this commit. Also the idea is that the common handler for predicates reduces the usual cases, and then each handler can care about only about the specific parts.

Also, I made a small refactor of the chain of function that reduce calls to primitives, so if a handler changes the primitive, it can just call to the middle of the chain and run the reductions of the new primitive instead of adding more special cases and repeat the code. In particular, in the previous code there was a very special case for `integer? -> flinteger?`, and now it's just a handler for `integer?` that when the argument is a flonum changes the primitive and tries again.  

---

**Add rational?**

Also `finite?`, `infinite?` and `nan?`. It was mean to show the new capabilities, but after looking more carefully I rewrote it as a trivial refactor of the `integer?` handler. So it was a success, but it can be squashed with the previous commit.

---

**Improve support of zero? in cptypes**

II think the handler for `zero?` was one of the first I wrote and the type system was very simple, perhaps only a few (20?) types instead of allowing a lot of combinations. I also split the exact zero from the other exact integers and put it in the singleton list so the system can track it completely. Similarly,  the 4 complex inexact zeros are now a multiplet split form the other inexact complex numbers.

This updates the implementation of the handler of `zero?` and also eliminates the special handlers for `fxzero?` and `flzero?` that can use the new default handler for predicates.

In particular, with the new structure we can reduce

    (if (zero? x) (zero? x) (zero? x))
    => [new reduction] =>
    (if (zero? x) #t #f)
    => [another unrelated old reduction] =>
    (zero? x)

---

**Track 0.0 and -0.0**

The main version, can track one number like

    (if (eqv? x 0.0) (eqv? x 0.0) (eqv? x 0.0)) => (if (eqv? x 0.0) #t (eqv? x 0.0))

but not two numbers

    (if (or (eqv? x 0.0) (eqv? x -0.0)) (or (eqv? x 0.0) (eqv? x -0.0)) (or (eqv? x 0.0) (eqv? x -0.0))) => no reduction

    (if (or (eqv? x 0.0) (eqv? x -0.0)) (zero? x) (zero? x)) => (if (or (eqv? x 0.0) (eqv? x -0.0)) #t (zero? x))

After this commit, there are special cases for `0.0` and `-0.0` that reduces both branches to `#t` and `#f` completely.

(This change is only for ero, for other numbers, let's say `7.0` and `-7.0` cptypes maintains the old behavior.)

The main idea is to isolate `0.0` and `-0.0` as singetons modulo `eqv?`, because `eq?` is unreliable on flonums because, for example, after constant propagation the compiler can replace them with another allocation of the same flonum.  

This can be squashed with the previous commit, but I didn't want to hide the idea of using singletons modulo `eqv?` in a big commit.

It's also useful to reduce

    (cond
      [(eqv? x 0.0) ...]
      [(eqv? x -0.0) ...]
      [(flzero? x) ...] ; this can be eliminated
      [...])


mflatt:
@gus-massa — Normally, we squash when merging PRs in this repo, but your commits seem carefully organized. Is it important to merge without squashing in this case?

PR #961, opened 2025-07-29
Fix build-let in cptypes

gus-massa:
When `cptypes` has to duplicate an expression, it uses `build-let` to create temporary variables
But it tries to propagate the obvious cases, like in

    (bitwise-and (f) 7)
    =>
    (let ([t (f)])
      (if (fixnum? t)
          (fxand t 7)
          (bitwise-and t 7)))

where `7` is copied instead of assigned to a temporary variable.

Don't propagate numbers that are not fixnums to avoid problems with `eq?`.

The main idea is that it's fine to have information moduo `eqv?` inside the `types tree`, that is good to reduce

    (when (eqv? x 7.0) (eqv? x 8.0)
    =>
    (when (eqv? x 7.0) #f)

but it's important to not copy that info like in 

    (when (eqv? x 7.0) (list x 8.0)
    => this does not happen =>
    (when (eqv? x 7.0) (list 7.0 8.0))

I took a look and I think all the other similar sites are careful to avoid copying non-fixnum-numbers, but the auxiliar function `prepare-let` does not check this, so I'm adding the check now.

I used `prepare-let` and `build-let` mostly with special cases for fixnums, so I couldn't find a way to cause a bug, but I think it's better to be more cautions. So, this is a change to make it more safe just in case.

### DIFF
--- a/mats/cptypes.ms
+++ b/mats/cptypes.ms
@@ -1606,6 +1606,18 @@
                        (if (#2%fixnum? y) ;the specialization uses #2%fixnum?
                            (fxand x y)
                            (bitwise-and x y))))))
+  (cptypes-equivalent-expansion?
+    '(lambda (x) (bitwise-and (begin (newline) 7) x))
+    '(lambda (x) (newline) ; ensure side effects are not duplicated
+                 (if (#2%fixnum? x) ;the specialization uses #2%fixnum?
+                     (fxand 7 x)
+                     (bitwise-and 7 x))))
+  (cptypes-equivalent-expansion?
+    '(lambda (x) (bitwise-and 7 (begin (newline) x)))
+    '(lambda (x) (newline) ; ensure side effects are not duplicated
+                 (if (#2%fixnum? x) ;the specialization uses #2%fixnum?
+                     (fxand 7 x)
+                     (bitwise-and 7 x))))
 )
 
 (mat cptypes-unreachable
--- a/s/cptypes.ss
+++ b/s/cptypes.ss
@@ -227,9 +227,14 @@ Notes:
          #t]
         [else #f]))
 
+    (define (result-expr e)
+      (nanopass-case (Lsrc Expr) e
+        [(seq ,e1 ,e2) e2]
+        [else e]))
+
     (define make-seq
       ; ensures that the right subtree of the output seq is not a seq if the
-      ; last argument is similarly constrained, to facilitate result-exp
+      ; last argument is similarly constrained, to facilitate result-expr
       (case-lambda
         [(ctxt e1 e2)
          (make-seq/no-drop ctxt (drop e1) e2)]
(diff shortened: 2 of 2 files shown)


dybvig:
This might not be necessary.  Unless something has changed, the FASL printer and linker maintain eq-ness of values.  So the only danger would be in propagating non-immediate values to separately compiled object files, e.g., through cross-library optimization.  In any case, the behavior of `eq?` is not defined for numbers, and `eqv?` does not depend on pointer equivalence for numbers.  So it should always be safe to copy numbers.

gus-massa:
Just to be clear, I'd be happy if the decision is **not** to merge this. Moreover, I'd send a PR in the oposite direction. 

For example, let's imagine I delete the guard in https://github.com/cisco/ChezScheme/blob/07b679ba4b49cf5507c04efa7856fce1e2da8ce8/s/cptypes.ss#L1737-L1741 that ensure that when a reference has a unique posible value, it's replaced by that value only if that value "super" safe to copy.

Anyway, my reasoning is this:

---

In **`cp0`**, `okay-to-copy?` is fine because `cp0` only does "**vertical**" constant propagation. [I don't know if this has a well know name.] For example in 

    (define (f x y)
      (let ([z (+ 4.0 4.0)])
        (list x y z 8.0)))

then `z` is also `8.0` but probably different than the `8.0` that the reader creates from the source code. So the result of

    (let ([temp (f 8.0 8.0)])
      (eq? (car temp) (caddr temp)))

is mysterious. I guess `#f` if run directly and perhaps `#t` if it's compiled to a .so and loaded in other program later. 

But in 

    (let ([temp (f 8.0 8.0)])
      (eq? (car temp) (cadr temp)))

I expect it to be `#t` always, but I'm not sure if there is a guaranty.

---

In **`cptypes`**, `there is "**diagonal**" constant propagation. [I don't know if this has a well know name.] For example in 

    (define (g x y)
      (let ([z (+ 4.0 4.0)])
        (when (eqv? y z)
          (list x y z 8.0))))

then, *if I remove the guard*, in the final expression both `y` and `z` will be replaced by the new `8.0` that is probably different than the `8.0` that the reader creates from the source code. So the result of

    (let ([temp (g 8.0 8.0)])
      (eq? (car temp) (caddr temp)))

is I guess `#f` if run directly and perhaps `#t` if it's compiled to a .so and loaded in other program later. 

But in this case

    (let ([temp (g 8.0 8.0)])
      (eq? (car temp) (cadr temp)))

is also I guess `#f` if run directly and perhaps `#t` if it's compiled to a .so and loaded in other program later. 

The difference is the new `8.0` created by `(+ 4.0 4.0)` jumps from `z` to `y`. So in spite `x` and `y` are obviously `eq?` the new value jumps and breaks the eq?-ness.

---

Anyway, in all the scenarios I consider that  `eq?` on the same flonum is very unreliable and weird interactions of optimization and saving can cause unexpected results, I'm not sure how many of the unexpected results we want to avoid.

gus-massa:
I'm closing this, but  anyway I squashed the commit in case someone wants to review it in the future. (Fell free to reopen it if you wish.)

PR #962, opened 2025-07-30
Fix undeclared __clear_cache.

rmolives:
When I was compiling ChezScheme for RISC-V and Loongarch, I encountered an issue

```
[  361s] ChezScheme/c/clearcache.c:48:3: error: call to undeclared function '__clear_cache';  ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
[  361s]    48 |   __clear_cache((char *)start, (char *)end);
[  361s]       |   ^
[  361s] 1 error generated.
```

This issue is due to the fact that implicit function declarations are not allowed after ISO C99, so I fixed it.

### DIFF
--- a/c/clearcache.c
+++ b/c/clearcache.c
@@ -45,6 +45,7 @@ void S_doflush(uptr start, uptr end) {
 #ifdef S_TARGET_OS_IPHONE
   sys_icache_invalidate((void *)start, (char *)end-(char *)start);
 #else
+  extern void __clear_cache(void*, void*);
   __clear_cache((char *)start, (char *)end);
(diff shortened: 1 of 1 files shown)


mflatt:
Thanks for checking on `#include<unistd.h>`.

Can you tell me more about your Linux variant and compiler? I think adding an `extern` is probably ok, but I'd prefer to limit this change to certain environments, because I'm unclear on the implications of an `extern` for something that is normally a compiler built-in. Toward that end, I'd like to be able to replicate the problem and investigate more, but I don't get a compilation error with GCC on Debian Sid for riscv64.

rmolives:
[EweOS](https://os.ewe.moe/), Clang

mflatt (approved):

Issue #901, opened 2025-01-15
Process substitution in using chez scheme at command prompt


taimoon:
On my x86 Fedora distro WSL laptop, the scheme version is 10.0.0 and when I try to run this below in bash

```bash
$ scheme --script <(echo "(+ 2 3)")
```

It throws error showing

```
Exception in port-position: failed on #<binary input port /dev/fd/63>: illegal seek
```

It does not work.

As comparison, it works for python

```bash
$ python3 <(echo "print(2 + 3)")
5
```

Will this is supported?

The chez scheme manual includes example of using pipe operator together scheme in command prompt.

```bash
(echo '(+ 2 3)')  | scheme -q
```

taimoon:
Besides, if `test.scm` file is as follows,

```scheme
(write (read (open-input-file (cadr (command-line)))))
(newline)
```

```bash
scheme --script test.scm <(echo '(+ 2 3)')
```

It works and prints out

```
(+ 2 3)
```

It is sufficient for my use case.

PR #966, opened 2025-08-05
Load without seek support.

rmolives:
It will appear when using stream transmission:

```bash
rmolives@debian:~/works/bin$ scheme --script <(echo "(display (+ 2 3))")
Exception in port-position: failed on #<binary input port /dev/fd/63>: illegal seek
```

Related issues are #901 

So we changed some codes to make it support the same operation as Python.


mflatt:
Well, I see that `$source-file-descriptor` is also an obstacle. Maybe there's not going to be an approach that avoids reading all of the input.

Issue #969, opened 2025-08-12
FEATURE REQ: Add CAS! for PAIR, maybe others


JBThiel:
It seems to me a CAS! capability for Pair would be useful,
similar to box/vector-cas!, along the lines:

(pair-cas! pair ['car | 'cdr] oldval newval)

or distinct names like
car-cas!, cdr-cas!  or set-car-cas!,  set-cdr-cas!,  etc.

In particular, CAS for CDR would be useful to support threadsafe adjusting list structures in-place. It's surprising this was not provided with the other two functions, box-cas! and vector-cas!.
Is there a reason, perhaps it's already achievable in some other way?

Currently lock or mutex, both heavier weight, are the main options for managing list structure contention, so CAS would be simpler and more elegant in certain situations.

A pair is very similar to a 2-vector, and box, though the internal slot offsets are different.
For completeness, one might also consider adding CAS for other value holding
types like bytevector, fxvector, string, ...
But it gets into diminishing returns, rare usecases.

Indeed, one can generally just use Box or Vector when needing CAS.
But Pairs/Lists are ubiquitous, and often involved in larger structures, so it would be worthwhile to additionally support at least this case.

mflatt:
I agree that `car-cas!` and `cdr-cas!` should be added, and I'll create a PR.

JBThiel:
Very nice, with the test cases, documentation and everything.
I look forward to trying it, will probably install the upcoming release
soon. I don't rev my Chez version very often, but
have immediate uses for this.

I have not specially needed CAS for the other mentioned types yet;
but will add some notes here or a new issue if it comes up.

Btw, I had been following that business about spurious failures,
as per #720.  That is some eerie unsettling behavior, is it understood
exactly how/why that happens? ... almost sounds like microprocessor bug,
extreme timing edge cases, even quantum effects.

At any rate, the program algorithm around CAS is supposed to cleanly
handle the fail case, so the occasional rare event like that is "acceptable",
given the clear warning and documentation to hopefully save the program/mer
from faulty assumptions.  I agree with your view/implementation that the Chez CAS functions propagate the direct machine result, versus undertake their own loops, which would be a very minor convenience at best and could hide some important info.

PR #970, opened 2025-08-13
add `car-cas!` and `cdr-cas!`

mflatt:
Addresses #969.


jltaylor-us (approved):

LiberalArtist:
While many procedures follow the naming convention *`type`*&#x2060;`-`&#x2060;*`verb`*, the precedents `set-car!` and `set-cdr!` led me to expect these names to be `cas-car!` and `cas-cdr!` (since they abbreviate e.g. `compare-and-set-car!`). Arguably, this correspondence may already be broken, since we have `box-cas!` despite `set-box!`. On the other hand, a difference is that a box is a type of value, whereas `car` and `cdr` are fields in a pair type. Chez Scheme doesn't have very many operations on boxes, but `box-immutable` and `box-immobile` both use `box-` as a prefix, whereas the only other entry in the index that uses `car` or `cdr` is `lisp-cdr` (from TSPL).

I don't have a strong opinion beyond &ldquo;naming things is hard&rdquo;, but I thought it was worth at least mentioning before this makes it into a release.

JBThiel:
I suppose the -cas! naming originates from vector-set! and would best be consistent,
else confusing. The vector and bytevector have many -set!
functions, thus vector-cas!.
And box-cas! followed that, being essentially a 1-vector so you don't need an index.

I had proposed pair-cas! with an "index" designator of maybe the symbols 'car / 'cdr
to parallel the (vector-cas! vector index ...), but it's more verbose,
and not many functions take symbol args like that.

car-cas! and cdr-cas! are short and sufficiently mnemonic to me, and follow the
convention so far.  Also cas as the suffix connects to the !, looks visually
distinctive.

cas- prefix is sensible too, for the reasons you mention.

I tend to think of cas! as a highly distinct operation, not really a set,
much more complicated than that. It might very well *not set*, and generally
requires additional logic around it.  It's akin to -lock! to my sense,
which all the lock functions have as suffix.

Whether CAS prefix or suffix I'm relatively impartial, except I think it should be consistent, because a distinct conceptual domain.  
So since it's -cas! suffix for the vector/box, that should probably continue,
unless wanting to re-think / rename all of them.

PR #976, opened 2025-09-04
add `__atomic` foreign convention and improve related flonum unboxing

### LINKED ISSUE #921: Optimizing allocations under successive flonum operations
I have this code to draw a ring using raylib: https://git.sr.ht/~williewillus/thdawn/tree/479f8b10eb3a520dcb37104924170aee9c47f1be/item/main.ss#L1861

It's numeric all around, all the inputs are flonums besides `texture` (an ftype) and `rgba` (currently unused).
The raylib: calls are all foreign procedures that take floats, except again for `set-texture` which takes the `texture` ftype and reads a fixnum out of it to forward to a foreign procedure.

When I added calls to this code to my render loop, I was surprised to see allocation pressure increase pretty sharply, I was expecting pretty minimal allocation. I'm not actually sure if the allocation pressure is due to flonum ops not being boxed, or overhead with all of the foreign calls.

I'm not too proficient at reading the Chez assembly output, and I wasn't sure if flonum optimizations have improved since #248, so I thought I'd log this in case anyone spots something easy.

mflatt:
The `__atomic` convention promises that a foreign function will not call back into Scheme, such as through a foreign callable, before it returns. That promise enables a more direct call to the foreign function. For example, on my machine, calling

```
(foreign-procedure __atomic "(cs)s_errno" () int))
```

takes about 80% the time of calling

```
(foreign-procedure "(cs)s_errno" () int))
```

When a foreign function has arguments or results that are represented as flonums, using `__atomic` allows the call to be exposed enough in intermediate compiler passes so that it can partitcapte in flonum-unboxing optimizations. In principle, unboxing optimizations could also apply to non-`__atomic` foreign functions, but the current calling convention for Scheme functions and the protocol between Scheme code and the kernel (especially the garbage collector) make that difficult. Further improvements there are possible, but functions that fit the `__atomic` constraint can be handled relatively easily. Probably `__atomic` is wanted for the best performance, anyway, when unboxing is relevant.

Meanwhile, various problems were interfering with flonum unboxing around `ftype-ref` and `ftype-set!` on `double` and `float` fields, and those problems are now fixed.

Intended to close #921 and uses parts of the example there by @williewillus in tests.


release_notes/release_notes.stex
```diff
@@ -152,6 +152,16 @@ The \scheme{expt} function produces a more accurate result when its
 first argument is a flonum and its second argument is an exact integer
 that has no equivalent flonum representation.
 
+\subsection{Foreign-procedure thread activation (10.3.0)}
+
+A new \scheme{__atomic} foreign-procedure convention indicates that
```
jltaylor-us, line 157: I think this section needs a different title, since `atomic` doesn't have anything to do with threads (or at least isn't explained in terms of threads).

release_notes/release_notes.stex
```diff
@@ -152,6 +152,16 @@ The \scheme{expt} function produces a more accurate result when its
...
+the foreign function will not call back to Scheme, such as through a
+foreign callable. Calls to atomic foreign functions may be compiled to
+more efficient form, particularly if the function accepts or returns
+flonums that may be unboxed (in the sense of Section~\ref{sec:unbox-floats}).
+Unboxing is also improved for \scheme{fptr-ref} and \scheme{fptr-set!}
+on \scheme{double} and \scheme{float} fields.
```
jltaylor-us, line 163: Does this improvement belong in the same section as the `atomic` calling convention, or are they two separate features that happen to target improvements in the same use case?

release_notes/release_notes.stex
mflatt: Ah, yes — copy and paste bug, where I forgot to update the text!

release_notes/release_notes.stex
mflatt: Split

PR #980, opened 2025-09-22
add `black-box`

mflatt:
The `black-box` function takes one argument and returns it, but optimization knows only that `black-box` returns a single value. This makes `black-box` useful for some testing and benchmarking tasks.


burgerrg (approved):
Since the tests take less than 1 ms, there's no need to risk doing less allocation.

jltaylor-us (approved):

PR #987, opened 2025-10-23
Only check the label, not the label/pl, for duplicate imports

### LINKED ISSUE #929: Combination of re-exports with identifier properties and renaming imports causes spurious ‘multiple definitions’ errors
```scheme
(library (beep-lib)
  (export cons beep)
  (import (chezscheme))

  (define beep 'beep)
  (define-property cons beep #t))

(library (boop-lib)
  (export button)
  (import (chezscheme)
          (rename (beep-lib)
                  (beep boop)))

  (define (button) (display boop)))
```

Attempting to load and import the `(boop-lib)` causes an error:

```
Exception: multiple definitions for cons in body (library (boop-lib) (export button) (import (chezscheme) (rename (...) (...))) (define (button) (display boop))) at line 1, char 1 of boop-lib.sls
```

This is incorrect because `cons` has the same binding in both libraries from which `(boop-lib)` imports it, and there is no identifier property conflict.

This only happens when one of the bindings imported from the re-exporting library is renamed. This library works fine:

```scheme
(library (button-lib)
  (export button)
  (import (chezscheme)
          (beep-lib))

  (define (button) (display beep)))
```

(Chez Scheme Version 10.1.0)

dpk:
Fixes #929.

I am not yet certain that this patch is complete. At the very least, this demonstrates that the behaviour of importing identifier properties appears to work as expected:

<details>

```scheme
(library (foo)
  (export floop floop-val)
  (import (chezscheme))
  (define floop)
  (define-syntax floop-val
    (lambda (stx)
      (lambda (lookup)
        (syntax-case stx ()
          ((_ id) #`'#,(lookup #'id #'floop))))))
  (define-property floop floop #t))

(library (bar)
  (export floop bart)
  (import (foo) (chezscheme))
  (define-syntax bart
    (lambda (stx)
      (lambda (lookup)
        (syntax-case stx ()
          ((_ id) #`'#,(lookup #'id #'bart))))))
  (define-property floop bart 12))

(library (baz)
  (export barted flooped)
  (import (foo) (bar) (chezscheme))
  (define barted (bart floop))
  (define flooped (floop-val floop)))

(library (baz2)
  (export barted flooped)
  (import (foo) (only (bar) bart) (chezscheme))
  (define barted (bart floop))
  (define flooped (floop-val floop)))

(eval 'barted (environment '(baz)))   ;=> 12
(eval 'flooped (environment '(baz)))  ;=> #t

(eval 'barted (environment '(baz2)))  ;=> #f
(eval 'flooped (environment '(baz2))) ;=> #t
```

</details>

However, I am not sure if `label` can contain a `label/pl` as well as the `entry-label`. Moreover, the Right Thing is probably to still signal a duplicate if an imported identifier has a *conflicting* property definition, but this patch won’t do that. (Then again, this matches the behaviour that it also won’t stop you defining the same property multiple times in a body.)

It may also be that there is a better place to fix this.

There should likely also be a test for this; the above example could serve as a basis.

I would appreciate feedback from those familiar with the innards of `syntax.ss`.

### DIFF
--- a/s/syntax.ss
+++ b/s/syntax.ss
@@ -3039,7 +3039,8 @@
                 (symbol-hashtable-set! ht sym (cons (make-entry marks label) orig-entry*))
                 (let ([entry (car entry*)])
                   (if (same-marks? (entry-marks entry) marks)
-                      (unless (eq? (entry-label entry) label)
+                      (unless (eq? (label/pl->label (entry-label entry))
+                                   label)
                         (defn-table-dup*-set! tbl (cons sym (defn-table-dup* tbl))))
                       (f (cdr entry*))))))))))
   (define record-only!


dpk:
Apologies for the bump, but I would appreciate an initial review of this.

The problem is that when you use any kind of qualified import (`only`, `except`, `rename`, etc.) of a library with identifier properties, a label/pl ends up in the `entry-label` slot when checking for duplicate imports. Since the duplicate check was with `eq?`, it would always falsely report a duplicate. This patch modifies the check for a duplicate by converting a label/pl to a label before doing the check.

The question is whether it might be better to change it to avoid putting a label/pl in an `entry-label` slot in the first place. That might be a somewhat more invasive patch, although ultimately cleaner: I’m just not sure enough of the data flow through `syntax.ss` to make that change with certainty at the moment.

If someone could confirm I’m on the right track or suggest a better way to fix this issue, I’ll add some tests and mark this ready to merge.

PR #991, opened 2025-11-10
Add exact and inexact to cptypes

gus-massa:
And a few related changes ...

---

**Add exact and inexact to cptypes**

This is the main part, it has a few subparts inside the same commit:

* Add `inexact` to cptypes: The old implementation of `real->flonum` calls `inexact` and this goes in the "wrong" direction of detail level, and it cause a problem when I add `inexact` to cptypes (or perhaps when I add `inexact` to cpprim). So I added a new function `$real->flonum` to 5_3.ss

* Add `exact` to cptypes: The problem is that `exact` is not total, and the domain is unusual, so it can't be converted to the unsafe version automatically, so I added a special case. An alternative is to add a new type `complex-rational` to primvars.ms and then change the signature to `(exact? [sig [(complex-rational) -> (boolean)]] ...)`. (Side note: Also, the handle has a special case to collaborate with `zero?` that is where I notice the problem of moving stuff from the tail position to non-tail.)

* Replace `$value` with `fl+`: This is an experimental idea. I'd like to add in the future a few `fl+` here and there so later `np-unbox-fp-vars!` can unbox more flonums. I'm not sure when it's a good idea to add new `fl+`, but replacing some `$value` with `fl+` looks like a win.

---

**Reverse arguments of $real->flonum**

Rename in library `real-flonum` to `$real-flonum` because it has two arguments like the primitive with the same name. 

Also, in all (most?) of the other cases, `who` is the first argument. I'm not sure if it's necessary to keep backward compatibility at with the system functions. 

Both changes are just cosmetic, but  they make the code more consistent.

---

**Add exact and inexact to cpprim**

I'm not very happy with this third part. There is too little work to do and it may increase the executable file size without increasing speed. Perhaps `inexact` is interesting but `exact` has too few possible improvements unless I add a lot of code to cpprim. 

(The boot files have probably only a stub for the increase of items in the library, without the implementation.)

### DIFF
--- a/boot/pb/equates.h
+++ b/boot/pb/equates.h
@@ -304,6 +304,7 @@ typedef uint64_t U64;
 #define forward_marker (ptr)0x3E
 #define forward_marker_disp 0x0
 #define ftype_guardian_rep (ptr)0x56
+#define fuel_word_count_shift 0x2
 #define fxvector_data_disp 0x9
 #define fxvector_length_factor 0x10
 #define fxvector_length_offset 0x4


s/cpprim.ss
```diff
@@ -5408,9 +5408,46 @@
...
+             e-x
+             (bind #t (e-x)
+               `(if ,(%type-check mask-fixnum type-fixnum ,e-x)
+                    ,(build-fixnum->flonum e-x values)
+                    (if ,(build-simple-or
+                          (%type-check mask-flonum type-flonum ,e-x)
+                          (%typed-object-check mask-inexactnum type-inexactnum ,e-x))
+                        ,e-x
+                        ,(build-libcall #t src sexpr inexact e-x)))))])
+      (define-inline 2 exact->inexact
```
mflatt: This seems like too much code duplication. Add a `build-exact->inexact` parameterized over the library entry to call?

PR #993, opened 2025-11-11
adjust argument checking for `>` and `>=`

mflatt:
If both arguments to `>` or `>=` are not real numbers, complain about the earlier argument. There's generally no guarantee about the order that arguments will be checked, but most checks are left-to-right, so it becomes expected by users. The checks for `>` and `>=` were backward only because they're implemented using `$<` and `$<=`. This small adjustment to error reporting should have minimal overall impact, since the adjustment is only only in the error case.


mflatt:
Release notes updated — with the same reservations as for #992 , and with the intent of merging the notes, assuming both PR get merged.

Issue #550, opened 2021-02-12
errno, threads, sockets et al.


amirouche:
Is it generally possible to query `errno` using C FFI?

amirouche:
It is possible, but does Chez reset errno because for instance threads management or GC ?

amirouche:
While greping the code I found out about the following:

```scheme
> (#%$errno)
0
```

It is not documented.

cjfrisz:
Anything that's called using the `#%$` prefix is a system procedure (see `s/primdata.ss`) and not guaranteed to have consistent behavior across versions or platforms. None of them are documented.

That said, it's possible this one could have a non-system counterpart that calls the system procedure `$errno`, but there could be a limitation I don't know off the top of my head.

jltaylor-us:
I presume the reason for wanting to do this is so that you can write some code like
```scheme
(some-foreign-function ...)
($errno)
```
The problem with this is that the garbage collector could run between those two calls (unless you've taken steps to disable it).  Whether the default garbage collector does or does not ever reset `errno` is irrelevant, since the garbage collector is pluggable - so arbitrary behavior could be inserted there.

So, while `#%$errno` _could_ be exposed and documented, it's almost certain to be used in an unsafe way (no matter how much documentation we write to encourage safe use, IMO).

If we were to go about trying to solve the problem of getting both the return value and final value of `errno` from a foreign call then that sounds a whole lot like multi-value return.  We could explore adding something like `foreign-procedure/errno` to create a function that returns the two values when invoked.

qzivli:
> If we were to go about trying to solve the problem of getting both the return value and final value of `errno` from a foreign call then that sounds a whole lot like multi-value return. We could explore adding something like `foreign-procedure/errno` to create a function that returns the two values when invoked.

We can have a `saved-errno` procedure as Racket did.

amirouche:
> We can have a `saved-errno` procedure as Racket did.

Quickly looking at Racket's code, I do not see how this can work without continuation marks / attachements. 

Using liburing, transparent procedures (without async / await, nor continuations), and code that can be preempted with something like [`make-engine`](https://www.scheme.com/csug8/control.html#g44); user code looks like:

```scheme
(call-with-values (lambda () (listen fd backlog)
 (lambda (ok? why)
  (unless ok?
   (error 'listen "Listen on socket failed" fd backlog why)))
```

[...]

PR #999, opened 2025-11-30
add __collect_errno foreign-procedure convention for reliable errno c…

amirouche:
…apture

This adds a new `__collect_errno` convention for foreign procedures that captures errno atomically immediately after the C call, before any potential GC or interrupt handling can corrupt it.

Usage:
  (define my-open
    (foreign-procedure __collect_errno "open" (string int int) int))

  (let-values ([(fd err) (my-open "/path" O_RDONLY 0)])
    (if (= fd -1)
        (error 'my-open ($errno->string err))
        fd))

Implementation:
- Added `saved-errno` field to thread context (cmacros.ss)
- Added `__collect_errno` convention recognition (syntax.ss)
- Added `$saved-errno` primitive to read captured errno (cpprim.ss, primdata.ss)
- Modified wrapper generation to return (values result errno) (syntax.ss)
- Added `errno-location` C entry for platform-specific errno access (prim5.c)
- Implemented errno capture in all architecture backends: x86_64, arm64, arm32, riscv64, loongarch64, ppc32, x86, pb

The convention is mutually exclusive with __collect_safe as errno capture requires atomic-like behavior to prevent GC from corrupting errno.

### DIFF
--- a/c/prim5.c
+++ b/c/prim5.c
@@ -29,6 +29,7 @@
 
 /* locally defined functions */
 static INT s_errno(void);
+static void *s_errno_location(void);
 static IBOOL s_addr_in_heap(uptr x);
 static IBOOL s_ptr_in_heap(ptr x);
 static ptr s_generation(ptr x);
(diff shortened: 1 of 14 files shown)


jltaylor-us:
Addresses #550 

The name `__collect_errno` is not great, because the term "collect" is already used in relation to garbage collection.  I suggest something like `__with_errno` or `__return_errno`

> The convention is mutually exclusive with __collect_safe as errno capture requires atomic-like behavior to prevent GC from corrupting errno.

I do not think this is true, at least on any modern-ish POSIX compliant system where `errno` is thread-local storage.

amirouche:
Sorry! That is a semi-public experiment, that end up as a pull requests here, because the default behavior of claude.ai, and github is to pull request upstream.

That is a mislick, and unintended. Sorry for the noise.

Closing.

PR #1006, opened 2026-01-14
Unicode 17.0.0

burgerrg:
Support Unicode 17.0.0.

The function `char-indic-break-property` was added to support correct grapheme cluster identification for Indic scripts.

The grapheme cluster break test was updated to use the test file from the Unicode Consortium.

Follow unicode/Readme to make future Unicode updates.

### DIFF
--- a/unicode/extract-char-cases.ss
+++ b/unicode/extract-char-cases.ss
@@ -407,28 +407,22 @@
     (fxior br
            (fxsll state grapheme-break-state-shift)))
 
-  (let ([states (make-eqv-hashtable)]
+  (let ([seen (make-eqv-hashtable)]
         [transitions (make-eqv-hashtable)])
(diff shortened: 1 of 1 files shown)


mflatt:
Would the name `char-indic-conjunct-break-property` (with `property` at the end) for the new function better parallel `char-grapheme-break-property`?

Historical note: I originally implemented `char-grapheme-cluster-step` for Unicode 14.0. It looks like the Indic conjuct break refinement was added in Unicode 15.0. I like how the improved testing approach here will ensure that we don't miss those kinds of additions in the future.

burgerrg:
Using a table for the step function is a great idea! I incorporated your code and moved the $char-extended-pictographic? function so that it uses the updated grapheme-break table.

I also renamed the new function to `char-indic-break-property`.

Issue #1009, opened 2026-01-24
Feature request: Add support for saving `errno` after foreign calls


spcfox:
After calling foreign procedures that use system calls, it may be useful to retrieve `errno`. According to [this comment](https://github.com/cisco/ChezScheme/issues/550#issuecomment-812032331), retrieving it via a separate call is unsafe, since `errno` may be modified by the Chez runtime. It would be useful to have an analogue of Racket’s [`#:save-errno` option](https://docs.racket-lang.org/foreign/foreign_procedures.html).

amirouche:
@spcfox see https://github.com/cisco/ChezScheme/issues/550#issuecomment-4085287267

burgerrg:
Fixed in https://github.com/cisco/ChezScheme/pull/1013

PR #1013, opened 2026-02-05
add `__errno`, `_alloc`, and `__disable_interrupts` foreign procedure conventions

mflatt:
The `__errno` convention makes a call for a foreign procedure return two values: the foreign procedure's direct result and the value of `errno` right after the foreign procedure returns.

On Windows, `current-errno-source` selects which `errno` is used for `__errno`. Also, `__get_last_error` is available as an alternative to `__errno`.

Since all backends are affected, I ran at least the `foreign.ms` test on all architecture+ABI combinations.


jltaylor-us:
#1009, #550

amirouche:
Thanks.

amirouche:
I just upgraded my epoll based http server, here are microbenchmark results:

```
  Benchmark Results

  ┌─────────────┬───────────────────────┬────────────────────────┬────────┐
  │   Metric    │ Old (call-with-errno) │ New (__atomic __errno) │ Change │
  ├─────────────┼───────────────────────┼────────────────────────┼────────┤
  │ Req/sec     │ 10,905                │ 11,361                 │ +4.2%  │
  ├─────────────┼───────────────────────┼────────────────────────┼────────┤
  │ Avg latency │ 6.80ms                │ 3.78ms                 │ -44%   │
  ├─────────────┼───────────────────────┼────────────────────────┼────────┤
  │ Max latency │ 1.53s                 │ 1.17s                  │ -23%   │
  └─────────────┴───────────────────────┴────────────────────────┴────────┘
```

Previously I was using the following hack:

```scheme
(define-syntax call-with-errno
  (syntax-rules ()
    ((_ thunk proc)
     ;; Chez GC must be disabled or it could stomp on errno.
     ;;
     ;; See:
     ;;
     ;;   https://github.com/cisco/ChezScheme/issues/550
     ;;
     (with-interrupts-disabled
      (let ((out (thunk)))
        (proc out (#%$errno)))))))
```

The most biggest delta is average latency that almost halfed. That is neat!

PR #1016, opened 2026-02-09
Use no-return in cptypes

gus-massa:
There are a few things that I don't understand, so this is more a long question with a proof of concept than a finished PR.

---

When `cptypes` detects an error that is not obvious, use `$app/no-return` so the other passes can notice it. But `cp0` transform `$app/no-return` to special flags in `preinfo`, so let's add the flags directly instead.

After this change, in the second leap `cp0` can notice that `cptypes` detected an error and apply more reductions. See the example at the bottom of cptypes.ms  Also, there is an old example in `cp0.ms` that has type problems on purpose, and is broken with an explicit use of `$app/no-return`.

I'm worry that this is wrong for some primitives that write/read the continuation attachment. Is there an obvious example? Is there a not obvious example?

Is it wrong with `lambdas` instead of primitives? I'm almost sure it is, but I'd like to see an example to be sure what to care about . (Anyway, probably `lambdas` will be left for a PR in  the future.)

---   

If this make sense and is (almost) correct, I'd like to take a look if the same change can be applied in more parts of `cptypes`, and do some refactor and cleanup.

### DIFF
--- a/mats/cptypes.ms
+++ b/mats/cptypes.ms
@@ -2482,3 +2482,19 @@
         '(lambda (x y) (fxdiv-and-mod x y) (zero? x))
         '(lambda (x y) (fxdiv-and-mod x y) #f)))
 )
+
+(mat cptypes-no-return
+  ; Ensure that cptypes and cp0 cooperate using no-return.
+  ; In particular, this is like a `begin0` form
+  ; where cptypes detects one of the branches will fail
+  ; and then cp0 can detect the first lambda is single-valued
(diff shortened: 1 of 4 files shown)


mflatt:
Not in that case, either, I'm pretty sure. A `($app/no-return fail ...)` is supposed to mean that the call to `fail` can be moved into a "more tail" position. It's not about the `error` that `fail` eventually reaches, so the part inside `(lambda (x) ...)`is not relevant.

PR #1025, opened 2026-03-06
fix bytevector-ieee-double-native-set! error

gus-massa:
When the last argument raises an error and the offset is out of the bounds, the compiler was reorganizing the code and the out of bound error was raised instead.

    (define bv (make-bytevector 16 0))
    (define (f x) (bytevector-ieee-double-native-set! bv 100 (+ x 1)))
    (f 'bad)

I think it's isolated enough to be fixed now, but it's more tricky than what I expected. 

It's almost a single line change, but I'm not sure if I'm using correctly the flags of `bind` to ensure that the last argument of `#2%bytevector-ieee-double-native-set!` is correctly handled when it's an `unboxed flonum`.  Also, I think once the temporary variable is marked as `fp` then `#3%bytevector-ieee-double-native-set!` will not expand to `$real->flonum` another time, just use the variable.

### DIFF
--- a/release_notes/release_notes.stex
+++ b/release_notes/release_notes.stex
@@ -3012,6 +3012,9 @@ When \scheme{>} and \scheme{>=} are applied to two non-real arguments, the first
 is now reported as incorrect instead of the second. In general, when two or more arguments
 are invalid, there is no guarantee about which one is reported invalid.
 
+A bug in the error produced by \scheme{bytevector-ieee-double-native-set!}---when the last
+argument raises and error and the offset is out of bounds---has been fixed.
+
 \subsection{Fix \scheme{remainder} and \scheme{modulo} for inexact/exact mixtures (10.4.0)}
 
 When \scheme{remainder} and \scheme{modulo} receive an inexact first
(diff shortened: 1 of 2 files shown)


gus-massa:
For some reason I was assuming that `#3%` version do not raise an error under **normal** conditions, they just crash or give a nonsensical result or give the correct result by chance or they even may raise an unexpected error or even the correct error, but you should not expect them to raise the correct error.

This is wrong, but perhaps it's correct for some specific cases like `#3$exact?`, but not in general.  (I made a fix in the sibling PR just in case. ) Here I was trying to avoid raising an error in `bytevector-ieee-double-native-set!!`, but it's not necessary. 

I agree with your changes. I guess it does not break flonum unboing but I'm not sure if it's possible to "see" it using something similar to `expand/optimize`. 

---

A minor comment, is that I don't like `maybe-fp` because I interpret it as `(or (flonum? x) (not x))`. In this context my interpretation of `maybe-fp` makes no sense, and I don't have a better name for it.

mflatt:
I merged the commit to use `try-fp` (renamed from `maybe-fp`). Does this look ok to you, @gus-massa?

gus-massa:
I think it's fine, and I like `try-fp` better.

PR #1029, opened 2026-03-19
repair for x86_64 struct return case

mflatt:
When a callable returns a struct that is small enough to fit into return registers, and when arguments to the callable include floating-point values, arguments were not delivered correctly to the callable.


burgerrg (approved):
Thank you for fixing this bug!

PR #1034, opened 2026-04-15
Create new a6hk and ta6hk machine types for haiku

tealeg:
This PR defines two new machine types for building Chez on the x86_64 variant of the [Haiku operating system:](https://www.haiku-os.org/)
- a6hk 
- ta6hk

These are essentially the same patches I applied to build the [Haikuports](https://github.com/haikuports/haikuports/wiki/) package that is already available on the platform.  It makes sense to me to push this upstream and and maintain the machine types here, on the upstream project rather than juts re-creating them for each new release.  I'm quite happy to take responsibility for maintaining them. 

Haiku isn't fully POSIX compliant, so there are some variations from the behaviour of UNIX like systems, reflected in the tests.

Looking forward to any feedback / ideas.


mflatt:
Thanks for this port!

Most of these changes make sense to me, but I worry about having to disable all the `"file exists"` expected-error tests (where `"file or directory already exists"` is the message on Haiku). I'll propose a different approach, which is to rewrite the error message as captured in tests to normalize it as `"file exists"` .

I don't see why some other test changes would be needed, such as disabling the `((lambda (x) (and (or (fixnum? x) (bignum? x)) (>= x 0))) (time-second $time-t8))` test. Maybe that was needed for an intermediate point during porting?

I think you probably included for changes than needed in `install.zuo`. It looks like the change was meant to be about hard links being unavailable on BeFS, and it seems like the adjustment to the `ln-f` function is all that is needed in the end.

Here's an additional commit that reverts/adjusts some changes along those lines:

https://github.com/mflatt/ChezScheme/commit/ff3b6f6736a973bcf8f2d780d77774ac87f1c240

With these changes, `make test-one-safe` passes in a build for me (on a Haiku installation in a VM), and `make install` still works. Does it still work on your machine?

mflatt:
FWIW, I also experimented with using this patch to build Racket (enough to start, but more is needed on the Racket side to work properly): https://github.com/mflatt/racket/tree/haiku

tealeg:
@mflatt - funnily enough, I also started work in that direction, but didn't get far yet.

tealeg:
@mflatt thanks also for the suggestions.  I'll bring up my build server on Sunday and give it a go.

tealeg:
@mflatt I've cherry-picked in your patch and tested.  Everything is passing.  Thank you so very much!

Issue #1031, opened 2026-04-06
Bug: `(open-source-file)` leaks file descriptor if path is a non-seekable device


cosmos72:
TL;DR
calling  `(open-source-file sfd)` with `sfd` containing a path to a non-seekable device, fails and returns `#f`.
Internally it opens a file descriptor that is never closed nor associated to a returned port - i.e. it is leaked.

How to reproduce:
```
$ scheme
Chez Scheme Version 10.3.0
Copyright 1984-2025 Cisco Systems, Inc.

> (define sfd (make-source-file-descriptor "/dev/tty" (open-bytevector-input-port #vu8())))
```

On Linux, executing `ls -l /proc/CHEZ_SCHEME_PID/fd` from another terminal shows:
```shell
total 0
lrwx------ 1 max users 64 Apr  6 14:23 0 -> /dev/pts/0
lrwx------ 1 max users 64 Apr  6 14:23 1 -> /dev/pts/0
lrwx------ 1 max users 64 Apr  6 14:23 2 -> /dev/pts/0
```

Now run the following in Chez Scheme:
```scheme
> (open-source-file sfd)
#f
```
Note: `(open-source-file sfd)` fails and returns `#f` because `/dev/tty` is not seekable. One expects that any resource internally allocated by `(open-source-file sfd)` is released in case of failure.

Execute again  `ls -l /proc/CHEZ_SCHEME_PID/fd` from another terminal. The output is:
```shell
total 0
lrwx------ 1 max users 64 Apr  6 14:23 0 -> /dev/pts/0
lrwx------ 1 max users 64 Apr  6 14:23 1 -> /dev/pts/0
lrwx------ 1 max users 64 Apr  6 14:23 2 -> /dev/pts/0
lr-x------ 1 max users 64 Apr  6 14:24 3 -> /dev/tty
```
i.e. file descriptor for `/dev/tty` is still open, but not associated to a port or some similar resource: it has been leaked.

ThePuzzlemaker:
I don't believe the FD is *truly* leaked, as it was released after a `(collect)` after my testing. However, it is still likely desirable to release the FD on failure of `open-source-file`.

burgerrg:
Thanks for finding this! The code in question is here: https://github.com/cisco/ChezScheme/blob/78429379b1d2a10c4f676f91a9d2d32e0a16761e/s/read.ss#L1715-L1729

It needs another `guard` clause after binding `ip` on line 1722 that closes the port when there's an exception.

[...]

PR #1036, opened 2026-04-16
eagerly close file port when open-source-file fails on a non-seekable device

burgerrg:
Addresses https://github.com/cisco/ChezScheme/issues/1031


mflatt (approved):
It's a worrying that `open-source-file` is currently not tested at all, but adding a portable test for this change would be a little painful (maybe more than it's worth), anyway. I can't see or detect any problem.

burgerrg:
That's exactly what I thought, and I'm satisfied with a code review and testing it manually.

PR #1038, opened 2026-04-23
repair cross-compilation of `ftype-scheme-object-pointer`

mflatt:
Checking whether a parent type represents a Scheme-object ftype pointer used an accessor technique that does not work for cross compilation.

Also, repair a documentation typo and a technically too-weak check for consistent record definition.

Unfortunately, there's not an added test here, because the problem is pretty far outside the current test harness. Detecting the bug requires cross-compilation on a 64-bit platform for a 32-bit platform, and then running in 32-bit mode and loading the cross-compiled program.  That could be checked via CI using a Linux host that can run both 32-bit and 64-bit executables, for example, but it seems a lot to add to check this small corner.

There's also not a bug-fix note in "release_notes.stex", because it's a repair to functionality that's new since most recent release.


burgerrg (approved):

Issue #1043, opened 2026-05-09
Regression on powerpc: 10.4.0 fails to build: `Exception: invalid memory reference`


barracuda156:
```
--->  Extracting csv10.4.0.tar.gz
Executing:  cd "/opt/local/var/macports/build/chez-scheme-cd21fd34/work" && /usr/bin/gzip -dc '/opt/local/var/macports/distfiles/chez-scheme/csv10.4.0.tar.gz' | /usr/bin/gnutar --no-same-owner -xf - 
--->  Patching version.h: s|/usr/X11R6/lib/libX11.dylib|/opt/local/lib/libX11.dylib|
--->  Patching expeditor.c: s|/usr/X11/bin/resize|/opt/local/bin/resize|
--->  Configuring chez-scheme
Warning: ccache_dir /opt/local/var/macports/build/.ccache could not be initialized; disabling ccache: couldn't execute "ccache": no such file or directory
Executing:  cd "/opt/local/var/macports/build/chez-scheme-cd21fd34/work/csv10.4.0" && ./configure --installprefix=/opt/local --libkernel --temproot=/opt/local/var/macports/build/chez-scheme-cd21fd34/work/destroot --threads LZ4=-llz4 ZLIB=-lz 
Configuring for tppc32osx, and will create boot files via pb
--->  Building chez-scheme
Executing:  cd "/opt/local/var/macports/build/chez-scheme-cd21fd34/work/csv10.4.0" && /usr/bin/make -j2 -w 
make: Entering directory `/opt/local/var/macports/build/chez-scheme-cd21fd34/work/csv10.4.0'
mkdir -p bin
/usr/bin/gcc-4.2 -pipe -Os -arch ppc  -DZUO_LIB_PATH='"'".././zuo/lib"'"' -o bin/zuo ./zuo/zuo.c
bin/zuo tppc32osx MAKE="/usr/bin/make"
checking boot files for tppc32osx against pb
/usr/bin/gcc-4.2 -I/opt/local/include -DPORTABLE_BYTECODE -Ipb/boot/pb -Ipb/c -Ic/ -pipe -Os -arch ppc -o pb/c/statics.o -c c/statics.c
/usr/bin/gcc-4.2 -I/opt/local/include -DPORTABLE_BYTECODE -Ipb/boot/pb -Ipb/c -Ic/ -pipe -Os -arch ppc -o pb/boot/pb/main.o -c c/main.c
[...]

PR #1045, opened 2026-05-10
ppc32 and logtest repairs

mflatt:
Repairs two problems related to #1043.

One problem is specific to ppc32 and `__atomic __alloc`.

The other one affects `logtest` on all platforms when given a mixture of fixnum and bignum arguments. That kind of mixture is probably more likely on a 32-bit platform, so that may be why testing did not discover the problem before, but it's not clear why ppc32 was affected more than other 32-bit platforms.


burgerrg:
Thanks for finding and fixing this! Please update the release notes.

burgerrg (approved):

PR #1048, opened 2026-06-07
fix domain error message of 2-ary `log`

usaoc:
When `log` is called with exact `1` as the second argument, it results in a division by zero, which was reported in `/`.  Report this error in terms of `log`, instead.

### DIFF
--- a/s/5_3.ss
+++ b/s/5_3.ss
@@ -1604,7 +1604,9 @@
             (/ (log (magnitude-squared x)) 2)
             (angle x))]
          [else (nonnumber-error 'log x)])]
-      [(x y) (/ (log x) (log y))])))
+      [(x y)
+       (when (eqv? y 1) (domain-error2 'log x y))
+       (/ (log x) (log y))])))
 
 (define-trig-op exp $flexp cflexp 1)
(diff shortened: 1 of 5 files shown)


s/5_3.ss
```diff
@@ -1604,7 +1604,9 @@
             (/ (log (magnitude-squared x)) 2)
             (angle x))]
          [else (nonnumber-error 'log x)])]
-      [(x y) (/ (log x) (log y))])))
+      [(x y)
+       (when (and (fixnum? y) (fx= y 1)) (domain-error2 'log x y))
```
mflatt: Minor, but using `(eqv? y 1)` here would be better than `(and (fixnum? y) (fx= y 1))`, since `(eqv? y 1)` will compile the same as `(eq? y 1)` and generate an immediate comparison to the fixnum representation of `1`.

mflatt (changes requested):

s/5_3.ss
usaoc: Changed.

mflatt (approved):

mflatt:
Thanks!

Issue #1041, opened 2026-04-27
installing under Termux fails: permission denied when creating hardlinks


mbakhterev:
I am building Chez under Termux on the phone. It is doable, but has two issues. Second one is that creation of hard links is prohibited.

The log of actions:

```
$ ./configure --prefix=$HOME/.local --disable-x11 CC=clang CFLAGS='-flto -O3' LIBS+='-liconv'
...
$ make
...
$ make install
bin/zuo tarm64le install MAKE="make" DESTDIR=""
makefiles/installsh -o "" -g "" -d -m 755 /data/data/com.termux/files/home/.local/lib/csv10.4.0-pre-release.4
makefiles/installsh -o "" -g "" -d -m 755 /data/data/com.termux/files/home/.local/lib/csv10.4.0-pre-release.4/tarm64le
makefiles/installsh -o "" -g "" -d -m 755 /data/data/com.termux/files/home/.local/lib/csv10.4.0-pre-release.4/examples
makefiles/installsh -o "" -g "" -d -m 755 /data/data/com.termux/files/home/.local/bin
makefiles/installsh -o "" -g "" -d -m 755 /data/data/com.termux/files/home/.local/share/man/man1
rm -f /data/data/com.termux/files/home/.local/bin/scheme
rm -f /data/data/com.termux/files/home/.local/bin/petite
rm -f /data/data/com.termux/files/home/.local/bin/scheme-script
makefiles/installsh -o "" -g "" -m 555 tarm64le/bin/tarm64le/scheme /data/data/com.termux/files/home/.local/lib/csv10.4.0-pre-release.4/tarm64le/scheme
ln -f /data/data/com.termux/files/home/.local/lib/csv10.4.0-pre-release.4/tarm64le/scheme /data/data/com.termux/files/home/.local/lib/csv10.4.0-pre-release.4/tarm64le/petite
ln: failed to create hard link '/data/data/com.termux/files/home/.local/lib/csv10.4.0-pre-release.4/tarm64le/petite' => '/data/data/com.termux/files/home/.local/lib/csv10.4.0-pre-release.4/tarm64le/scheme': Permission denied
shell command failed
 in install {3}
 in build-one
 in loop
 in module->hash
make: *** [Makefile:20: install] Error 1
```
When i change the line: https://github.com/cisco/ChezScheme/blob/4bde3b35316737b1c5d34329d02dbdf609fdd399/makefiles/install.zuo#L184
to `(if #t`, everything works

```
$ make install
[...]

Issue #1040, opened 2026-04-27
building in Termux fails: -liconv is not listed in libs


mbakhterev:
I am building Chez inside Termux environment on the phone. It buildable, but there is two issues. The first one: -liconv is not listed in LIBS.

The log of actions:

```
$ ./configure --prefix=$HOME/.local --disable-x11 CC=clang CFLAGS='-flto -O3'
...
Configuring for tarm64le, and will create boot files via pb
$ make
...
clang -flto -O3 -D_REENTRANT -pthread -o tarm64le/bin/tarm64le/scheme tarm64le/boot/tarm64le/main.o tarm64le/boot/tarm64le/libkernel.a tarm64le/zlib/libz.a tarm64le/lz4/lib/liblz4.a -rdynamic -lm -ldl -lncurses -lrt -lpthread
ld.lld: error: undefined symbol: libiconv_open
>>> referenced by ld-temp.o
>>>               tarm64le/bin/tarm64le/scheme.lto.o:(s_iconv_open)

ld.lld: error: undefined symbol: libiconv_close
>>> referenced by ld-temp.o
>>>               tarm64le/bin/tarm64le/scheme.lto.o:(s_iconv_close)

ld.lld: error: undefined symbol: libiconv
>>> referenced by ld-temp.o
>>>               tarm64le/bin/tarm64le/scheme.lto.o:(s_iconv_from_string)
>>> referenced by ld-temp.o
>>>               tarm64le/bin/tarm64le/scheme.lto.o:(s_iconv_to_string)
clang: error: linker command failed with exit code 1 (use -v to see invocation)
link failed
 in build-one
 in loop
 in module->hash
make: *** [Makefile:8: build] Error 1
```

But when `-liconv` manually specified with configure, make succeeds:

```
$ ./configure --prefix=$HOME/.local --disable-x11 CC=clang CFLAGS='-flto -O3' LIBS+='-liconv'
Configuring for tarm64le
$ make
bin/zuo tarm64le MAKE="make"
clang -flto -O3 -D_REENTRANT -pthread -o tarm64le/bin/tarm64le/scheme tarm64le/boot/tarm64le/main.o tarm64le/boot/tarm64le/libkernel.a tarm64le/zlib/libz.a tarm64le/lz4/lib/liblz4.a -rdynamic -liconv -lm -ldl -lncurses -lrt -lpthread
: tarm64le/bin/tarm64le/scheme
[...]

PR #1050, opened 2026-07-17
add `arm7le` machine type, `configure` supoprt for Termux, `arm32` FFI repairs

### LINKED ISSUE #1049: soft-float armv7 ABI support? (needed for termux)
Chez Scheme supported machine types currently include  `arm32le` and `tarm32le`.
They both assume `armhf` C ABI i.e. hard-float: when calling a C function via `(foreign-procedure)`, floating point values with C type `double` are passed and returned in floating point registers `d0 d1 ...`  by C functions.

On the other hand, Android termux for 32-bit ARM devices uses `armel` ABI i.e. soft-float:
floating point values with C type `double` are passed and returned in general registers `r0 r1 ...` by C functions.

Trying to compile Chez Scheme 10.4.1 on Android termux "almost" works (after working around issues #1040 and #1041).

The compiled executable can be used for most purposes, except floating point computations:

- displaying any flonum or converting it to string raises a condition `expt: #<foreign> is not a number`
- all functions that internally call C functions accepting or returning a C double are broken. Examples:
`(= 3.0 (floor 3.7))` returns `#f`
`(= 3.7 (floor 3.7))` returns `#t`
`(= 4.0 (sqrt 4.0))` returns `#t`

The reason is the ABI mismatch summarized above:
Chez Scheme assumes flonum values are exchanged with C functions via floating point registers `d0 d1 ...` as per `armhf` ABI,
while Android termux libc.so, libm.so, compiler etc. expect and return floating point values in general registers `r0 r1 ...` as per `armel` ABI.

I have not found any mention of an `armhf` version of Android termux for 32-bit ARM devices: termux publishes only **one** 32-bit ARM ABI and it's soft-float.

Is there any chance to add support for `armel` soft-float C abi to Chez Scheme?

Note: there is no need to change the compiler internals or the ABI of compiled scheme functions:
only `(foreign-procedure)` needs to know about the new ABI - possibly as a compile time switch when building Chez Scheme.

P.S. the main alternative to run Chez Scheme on Android is UserLand, a proot-based app that allows installing many Linux distributions.
Chez Scheme works perfectly on it, but the environment is much more limited: no access to files outside proot (unless one buys the "Pro" version), and job control is broken (proot implementation currently has a bug that swallows `SIGSTOP` `SIGTSTP` `SIGTTIN` `SIGTTOU`)

mflatt:
The `arm7le` machine type corresponds to ARMv7 armel (like Android), while `arm32le` is still ARMv6 armhf (like Raspberry Pi).

I was not able to set up a 32-bit ARM Termux environment to make sure everything works. I tired out Arm64 Termux, and I built on armel Debian to test `tarm7le`.

Running "foreign.ms" tests on `arm32le`, meanwhile, exposed old problems via tests that were added relatively recently. Those problems involved unusual situations: `__collect-safe` mode, structs with unions that have only floating-point fields, and callables that have struct results and floating-point arguments.

Closes #1049, #1040, #1041

### DIFF
--- a/configure
+++ b/configure
@@ -90,6 +90,7 @@ disablex11=no
 disablecurses=no
 disableiconv=no
 addflags=yes
+hardlinks=yes
 addwarningflags=no
 default_warning_flags="-Wpointer-arith -Wall -Wextra -Wno-implicit-fallthrough"
 : ${CC:="gcc"}
@@ -113,6 +114,7 @@ enableFrompb=yes
 pbendian=l
 emscripten=no
 empetite=no
+android=no
 crossCompile=no
 moreBootFiles=
 preloadBootFiles=
@@ -137,6 +139,10 @@ case "${CONFIG_UNAME}" in
     unixsuffix=le
     installprefix=/usr
     installmansuffix=share/man
+    if [ "$(uname -o)" = "Android" ]; then
+        android=yes
+        hardlinks=no
+    fi
     ;;
   GNU)
     unixsuffix=gnu # the Hurd
@@ -147,6 +153,7 @@ case "${CONFIG_UNAME}" in
     unixsuffix=hk
     installprefix=/system
     installmansuffix=documentation/man
+    hardlinks=no
     ;;
   QNX)
     if uname -m | egrep 'x86' > /dev/null 2>&1 ; then
@@ -254,6 +261,10 @@ if [ "$unixsuffix" != "" ] ; then
         m64=arm64${unixsuffix}
         tm32=tarm32${unixsuffix}
         tm64=tarm64${unixsuffix}
+        if [ "$android" = "yes" ]; then
+            m32=arm7${unixsuffix}
+            tm32=tarm7${unixsuffix}
+        fi
     elif uname -m | grep 'riscv64' > /dev/null 2>&1 ; then
         m32=""
         m64=rv64${unixsuffix}
@@ -410,6 +421,9 @@ while [ $# != 0 ] ; do
       ;;
     --enable-warning-flags)
       addwarningflags=yes
+      ;;
+    --disable-hard-links)
+      hardlinks=no
       ;;
     --libkernel)
       Kernel=KernelLib
@@ -687,6 +701,7 @@ if [ "$help" = "yes" ]; then
   echo "  --enable-libffi                   enable libffi support for pb"
   echo "  --disable-auto-flags              no auto additions to CFLAGS/LDFLAGS/LIBS"
   echo "  --enable-warning-flags            add GCC warning flags to CFLAGS"
+  echo "  --disable-hard-links              install with soft instead of hard links"
   echo "  --libkernel                       build libkernel.a (the default)"
   echo "  --kernelobj                       build kernel.o instead of libkernel.a"
   echo "  --installprefix=<pathname>        final installation root ($installprefix)"
@@ -894,6 +909,9 @@ if [ "$addflags" = "yes" ] ; then
   case "${flagsm}" in
     *le|*gnu|*hk)
         LDFLAGS="${LDFLAGS} -rdynamic"
+        if [ "$android" = "yes" ]; then
+            LDFLAGS="${LDFLAGS} ${iconvLib}"
+        fi
         ;;
     *fb|*nb)
         LDFLAGS="${LDFLAGS} -rdynamic -L/usr/local/lib"
@@ -1238,4 +1256,5 @@ GzipManPages=$gzipmanpages
 InstallSchemeName=$installschemename
 InstallPetiteName=$installpetitename
 InstallScriptName=$installscriptname
+InstallHardlinks=$hardlinks
 END
--- a/makefiles/install.zuo
+++ b/makefiles/install.zuo
@@ -38,6 +38,8 @@
         (~a s ".exe")
         s))
 
+  (define hardlinks? (not (equal? (lookup 'InstallHardlinks) "no")))
+
   ;; The following variables determine where the executables, boot files,
   ;; example programs, and manual pages are installed.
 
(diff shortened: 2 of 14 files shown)


makefiles/install.zuo
```diff
@@ -179,10 +183,8 @@
     (shell/wait* "rm" "-f" f))
   (define (rm-rf d)
     (shell/wait* "rm" "-rf" d))
-  (define haiku? (glob-match? "*hk" m))
   (define (ln-f from to)
-    (if haiku?
-        ;; no hard links on BeFS
+    (if (not softlinks?)
```
burgerrg: It seems odd to use a symbolic link when `(not softlinks?)`.

mflatt: ~~Ah, right. Among other problems, I've gotten confused there about Haiku/BeFS versus Termux.~~ I have the meaning consistently backwards, and things need to be renamed from "soft" to "hard". I'll sort this out - thanks!

mflatt: Now fixed.

cosmos72:
That was fast!
I've successfully compiled and installed ChezScheme with this patch on a 32-bit arm Android tablet, and it works basically perfectly, including all floating point computations and `(foreign-procedure)` that I tried. mpressive.

cosmos72:
A minor note: checking for environment variable `$TERMUX_VERSION` is not very robust - the termux app from Google Play Store defines it, while the one from F-Droid does not.

A more thorough solution for detecting termux is `[ "$(uname)" = "Linux" -a "$(uname -o)" = "Android" ]`

For completeness: `uname -o` is Linux-specific, and it curiously prints `Linux` on pure Android - as for example from `adb shell` - while it prints `Android` from Termux app running on Android.

So the Linux-specific  `configure` fragment
```shell
    if [ -n "$TERMUX_VERSION" ]; then
        hardlinks=no
    fi
```
could be rewritten as
```shell
   if [ "$(uname -o)" = "Android" ]; then
        hardlinks=no
   fi
```


Similary, the `configure` fragment
```shell
        if [ -n "$TERMUX_VERSION" ]; then
            m32=arm7${unixsuffix}
            tm32=tarm7${unixsuffix}
        fi
```
could be rewritten as follows, taking advantage of `$CONFIG_UNAME` already being set to `$(uname)`
```shell
        if [ "$CONFIG_UNAME" = "Linux" -a "$(uname -o)" = "Android"  ]; then
            m32=arm7${unixsuffix}
            tm32=tarm7${unixsuffix}
        fi
```

And the final `configure` fragment
```shell
        if [ -n "$TERMUX_VERSION" ]; then
            LDFLAGS="${LDFLAGS} ${iconvLib}"
        fi
```
could be rewritten as
```shell
        if [ "$CONFIG_UNAME" = "Linux" -a "$(uname -o)" = "Android"  ]; then
            LDFLAGS="${LDFLAGS} ${iconvLib}"
        fi
```

mflatt:
@cosmos72 Thanks for the pointers! I've opted to set an `android` variable in the initial `CONFIG_UNAME` switch. I think it still does the right thing for 64-bit Arm installed via a GitHub release of Termux, at least. A further check by you to make it it works more generally would be helpful.

cosmos72:
Where can I find the new (or updated) PR ?

mflatt:
Oops - now pushed here!

cosmos72:
Apart from the trivial default initialization `android=yes` that should be `android=no`, looks good to me.

The missing `include <errno.h>` reported by CI is worth fixing, but looks unrelated to this pull request.

Tested on:
Termux version googleplay.2026.06.21 (from google play) on Android 11 **tarm7le**

Termux version 0.118.3 (from f-droid) on Android 16 **tarm64le**

mflatt:
> The missing `include <errno.h>` reported by CI is worth fixing, but looks unrelated to this pull request.

I think that was due to the `android=yes` mistake. Let's see how the fixed version goes.

mflatt:
It wasn't obvious to me how the `android=yes` mistake led to the compilation error with `errno`, so I investigated a little more. The reason is that adding `-liconv` breaks zlib's configure script's conclusions when `libiconv` does not exist in the compilation environment, causing it to arrive at an incorrect configuration.

cosmos72:
Because `-liconv` is added to **all** linker invocations done by configure script?

That's ugly, as zuo, zlib, lz4, and most linker invocatios from configure don't need it - from what I understand, only the main `scheme` executable (and possible few other binaries) need it

mflatt:
Fair point. The practical implications are minimal in this case, since `zlib` built this way will be statically linked. From a wider perspective, I would say that automatically building a vendored zlib is where things go wrong in principle: using `ZLIB=` with either system-available libraries or separate builds (supported by something like `vpkg` if not the OS's package manager) is better. But the vendored approach simplifies the build for many users, and having one `LDLIBS` is part of that simplification.

configure
```diff
@@ -147,6 +153,7 @@ case "${CONFIG_UNAME}" in
     unixsuffix=hk
     installprefix=/system
     installmansuffix=documentation/man
+    hardlinks=no
```
burgerrg, line 156: Can we get rid of the `(glob-match? "*hk" m)` code in makesfiles/install.zuo thanks to this?

mflatt: Good point! I've updated "makefiles/install.zuo" and double-checked that installing still works on Haiku.

burgerrg (approved):

hyln9:
@mflatt One small naming suggestion: would armv7le be preferable to arm7le? It makes the ARMv7 architecture version explicit and avoids potential confusion with the older ARM7 processor family. It also seems more consistent with commonly used names such as armv7l.

mflatt:
@hyln9: That's a good point, and `armv7le` does seem a little better to me. Do you have an opinion on that, @cosmos72?

cosmos72:
armvt7le sounds fine to me too

PR #1051, opened 2026-08-17
rename `arm7le` to `armv7le`

mflatt:
Per continued discussion at #1050.


burgerrg (approved):

cosmos72:
There's a small missing part in this PR:

file `s/arm7.def` needs to be renamed to `s/armv7.def` because that's the new path compilation tries to open with this patch

mflatt:
Yep — I got that wrong. To confirm, `make arm7.boot` fails right now, and it succeeds for me after the renaming in #1053.

PR #1052, opened 2026-08-21
Raise implementation restriction for oversized objects

### LINKED ISSUE #1047: (make-vector k) should raise &implementation-restriction if k exceeds the max vector length
In Chez 10.2.0, the maximal capacity of a vector is limited by a constant `maximum-vector-length`, which was set in the compilation of Chez. If `k` in `(make-vector k)` is larger than `maximum-vector-length`, Chez doesn't actually allocate any objects but raise an exception instead.

However, according to my interpretation on R6RS, the condition object being raised must contain a `&implementation-restriction` component. Quoting from R6RS 5.3 "Exceptional situations"

> Implementations must raise an exception when they are unable to continue correct execution of a correct program due to some implementation restriction. For example, an implementation that does not support infinities must raise an exception with condition type &implementation-restriction when it evaluates an expression whose result would be an infinity

Currently, it just raise a simple error.

```
> (guard (e ((implementation-restriction-violation? e) #t) (else #f)) (make-vector (+ 1 (greatest-fixnum))))
#f
```

Furthermore, the practice of "raising exception with condition type `&implementation-restriction` when specific restriction prevent further evaluation" has precedents in R6RS: `fx+` and `fx-` will raise `&implementation-restriction` if its arithmetic result exceeds the range of fixnum.

paveluv:
Fixes #1047.

This changes the checked public constructors to raise
`&implementation-restriction` when their length is a nonnegative exact integer
that exceeds Chez Scheme's representable object-length limit. Other invalid
lengths, including negative, inexact, and nonnumeric values, continue to raise
`&assertion`.

The change covers the related constructor family instead of only
`make-vector`:

- `make-vector` and `make-immobile-vector`
- `make-string`
- `make-bytevector` and `make-immobile-bytevector`
- `make-reference-bytevector` and
  `make-immobile-reference-bytevector`

The immobile and reference variants are included because the Chez Scheme User's
Guide documents them in terms of the corresponding base constructors. Internal
allocation helpers, `make-fxvector`, `make-flvector`, and
`make-phantom-bytevector` are intentionally unchanged because their contracts
or allocation semantics differ.

The existing fast success checks are unchanged. Classification occurs only on
the exceptional path, using `$impoops` for an exact integer above the relevant
maximum and retaining the existing `$oops` path and message for other invalid
arguments.

There is also relevant R6RS design history. In item 30 of a 2006 exception
classification ballot, all three completed ballots located voted that an
exception must be raised when `make-string` or `make-vector` receives a
nonnegative exact integer above an implementation-specific size limit. R. Kent
Dybvig and Anton van Straaten selected `&implementation-restriction`; Michael
Sperber preferred proposed `&string-size` / `&vector-size` condition types.
This is supporting design history, not an explicit procedure-specific rule in
the final R6RS:

- Template: https://r6rs.org/r6rs-editors/2006-May/001260.html
- Dybvig: https://r6rs.org/r6rs-editors/2006-May/001287.html
- Sperber: https://r6rs.org/r6rs-editors/2006-May/001284.html
- van Straaten: https://r6rs.org/r6rs-editors/2006-May/001290.html

Tests cover every changed arity, verify the condition type, message, and
irritant for oversized inputs, and verify that negative inputs remain assertion
violations. The `primvars` accepted-message logic recognizes the new message
shape while still requiring matching `who` and irritants, avoiding
machine-word-size-dependent expected-error output.

Validation performed on `ta6le`:

- bootstrap completed unchanged after rebuild
- focused string, vector, bytevector, immobile-object, and reference-bytevector
  mats passed in compiled safe mode
- affected focused mats passed in interpreter mode
- `test-some-fast` passed all three configurations
- full `o=0` default and `spi=t` error recordings were generated
- the default root and derived `spi=t` expectations matched their recordings
  byte-for-byte
- the `spi=t` expectation patch applied without failed or fuzzy hunks


jltaylor-us (approved):
Looks ok to me.

Note that this is _technically_ a backwards-incompatible change, although I have difficulty imagining how any real program would rely on catching the specific exception type that's changed by this PR in some scenarios.

burgerrg:
Because this is a breaking change, there needs to be a compelling reason to make it. What trouble is caused by the current exception type?

paveluv:
> Because this is a breaking change, there needs to be a compelling reason to make it. What trouble is caused by the current exception type?

I personally haven't bumped into it (yet). The motivation is semantic consistency, not practical trouble. But, logically, if changing the condition type can break code that relies on &assertion, then the current type can likewise break code that correctly treats &assertion as a contract violation.

mflatt:
This change seems fine to me. The backward incompatibility seems small and unlikely to create trouble.

PR #1056, opened 2026-09-02
fix typos in pb.ss documentation

nasso:


burgerrg (approved):
Thank for finding and fixing these.

PR #1057, opened 2026-09-04
Fix typos in CSUG

JuanitoFatas:
Found some typos in the [User’s Guide](https://cisco.github.io/ChezScheme/csug/csug.html).

### DIFF
--- a/csug/csug.stex
+++ b/csug/csug.stex
@@ -23,7 +23,7 @@
 \endschemeinit
 
 \generated
-;; We assume here that the verison used to build the documentation
+;; We assume here that the version used to build the documentation
 ;; is the same as the documented version. If it's a release version,
 ;; then the date is fixed to the release date. Otherwise, use
 ;; the date when the document is built.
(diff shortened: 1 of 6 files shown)


csug/compat.stex
```diff
@@ -647,7 +647,7 @@ If no fields other than those initialized by the arguments to the
 constructor procedure are needed, the second subexpression,
 \scheme{((\var{id_2} \var{expr}) \dots)}, may be omitted.
 
-\index{pares}\index{\scheme{make-pare}}The following simple example
+\index{pairs}\index{\scheme{make-pare}}The following simple example
```
sorawee: I think it's intentionally spelled `pares`?

jltaylor-us: Yes, as it is the index entry for the `pare` structure and `make-pare` function in the immediately following example.

JuanitoFatas: Ah thanks, reverted this change 🙌🏻

jltaylor-us:
Thanks!

PR #1061, opened 2026-09-15
Fix mprotect addresses in the W^X chunk walk

dspearson:
## Summary

Under `WRITE_XOR_EXECUTE_CODE`, `enable_code_write`'s threaded chunk walk accumulates a run's length in segments but derives that run's start address from `chunk->addr`. A chunk's first segment need not begin at `chunk->addr`. Where the two differ, every `mprotect` the walk issues covers a window of the correct width sitting that difference too low, so the top of a qualifying segment keeps whatever protection an earlier flip left on it.

The mis-protection is silent. The shifted range is still inside the mapping and still page aligned, so `mprotect` succeeds and returns 0.

## Where

`c/segment.c`, `enable_code_write`, the `#if defined(PTHREADS)` arm. Two expressions, at `:698` (read by the run loop) and `:725`:

```c
698:      addr = chunk->addr;
699: # if defined(PTHREADS)
...
705:      } else {
706:        /* Flip bits for whole runs of segs that are either unused or
707:           whose generation is within the range [0, maxg]. */
708:        int j;
709:        for (j = 0; j < chunk->segs; j++) {      /* addr is still chunk->addr here */
...
725:            addr = TO_VOIDP((char *)chunk->addr + (j + 1) * bytes_per_segment);
```

Both lines arrived verbatim with `9eade055e894` ("add support for cross-compiling to iOS", 2021-01-22), the commit that introduced `enable_code_write` — that is, they were written for the one platform on which they cannot misbehave.

There is a near-miss in between. `81b5d322b865` ("repair to W^X mode", 2021-05-04) rewrote the **hint** path at the top of the same function from

```c
-    addr = TO_VOIDP((char*)hint - ((uptr)hint % bytes_per_segment));
+    for (seg = start_seg; seg <= end_seg; seg++) {
+      addr = TO_VOIDP(build_ptr(seg, 0));
```

which is precisely this correction, applied to one of the two places that needed it. The chunk walk was left on the arithmetic form.

## Mechanism

`allocate_segments` over-allocates one segment and rounds the segment base up (`c/segment.c:400-414`):

```c
  bytes = (nact + 1) * bytes_per_segment;
  addr = S_getmem(bytes, 0, for_code);
  base = addr_get_segment((uptr)TO_PTR(addr) + bytes_per_segment - 1);
  if (build_ptr(base, 0) == TO_PTR(addr) && ...) nact += 1;
  chunk->addr = addr;
  chunk->base = base;
```

Write `delta = (uptr)build_ptr(chunk->base, 0) - (uptr)chunk->addr`.

* If `S_getmem` returned segment-aligned memory then `delta == 0`, and (barring the wrap-around guard) the `nact += 1` arm is taken, so the chunk's segments exactly fill the mapping.
* Otherwise `0 < delta < bytes_per_segment`: the mapping is `delta` bytes of head padding, then `nact` segments, then `bytes_per_segment - delta` bytes of tail padding.

The existence of that `nact += 1` arm is itself the proof that `build_ptr(chunk->base, 0)` and `chunk->addr` can differ — it is the code checking whether they do.

`chunk->sis[j]` is the seginfo for segment number `chunk->base + j`, which lives at `build_ptr(chunk->base + j, 0)`, i.e. at `chunk->addr + delta + j * bytes_per_segment`. The walk uses `chunk->addr + j * bytes_per_segment`. Every run is therefore shifted down by `delta`: it begins `delta` bytes inside whatever lies below its first segment, and it ends `delta` bytes short of the end of its last segment. That trailing `delta` bytes belongs to a qualifying segment and is never flipped; the `delta` bytes below the run, which do get flipped in its place, are either head padding or the tail of a segment that was deliberately excluded from the run.

Consequences, by direction:

| call | `flags` | bytes left behind |
|---|---|---|
| `S_thread_start_code_write` (`on = 1`) | `PROT_WRITE\|PROT_READ` | stay `R-X` while Chez believes code writes are enabled, so the next write there faults. The collector's `copy()` installing a forwarding pointer into a from-space code object is one such write. |
| `S_thread_end_code_write` (`on = 0`) | `PROT_EXEC\|PROT_READ` | stay `RW-` while Chez believes code is execute-only. |

A flip sets an absolute permission, so the tail is never both writable and executable at once — only out of step with the segment it belongs to. Which of the two failures you get depends on the order of the flips that reached the segment.

## Why this has never fired upstream

Three conditions are needed at once:

1. **`WRITE_XOR_EXECUTE_CODE`**, or `enable_code_write` is not compiled at all.
2. **`PTHREADS`**, since the run loop is in the threaded arm. The non-threaded arm takes `bytes = chunk->bytes` with `addr = chunk->addr`, i.e. the whole mapping, which is correct.
3. **a page size finer than `bytes_per_segment`**. `S_getmem` is `mmap`, so its result is page aligned; if pages are at least as coarse as a segment then page aligned implies segment aligned and `delta` is always 0. `bytes_per_segment` is 16K on 64-bit targets and 8K on 32-bit ones.

`WRITE_XOR_EXECUTE_CODE` is defined at exactly one place in tree — `c/version.h:346`, inside `#if defined(TARGET_OS_IPHONE) && TARGET_OS_IPHONE`. macOS/arm64 does not take that arm: the next block reads `# if !defined(WRITE_XOR_EXECUTE_CODE)` and installs `pthread_jit_write_protect_np` instead. `PORTABLE_BYTECODE` undefines the macro, and nothing in `configure` sets it. So the only supported configuration that compiles this code is iOS, and `BUILDING` documents that as `-m=tarm64ios` — arm64, 16K pages, `delta == 0` always.

Two caveats. `TARGET_OS_IPHONE` is also true for the iOS Simulator; on an Intel Mac that is x86-64 with 4K pages and would be exposed, though that host is end-of-life and on Apple Silicon the simulator is arm64/16K. And any port that enables `WRITE_XOR_EXECUTE_CODE` on a finer-paged host is exposed, which is how we found it.

## How we hit it

We hit it porting Chez to an operating system of our own, with a kernel, a libc and a userland all written from scratch. Chez runs there natively as a threaded x86-64 build. Our kernel refuses to map any page both writable and executable, so the default x86-64 `S_PROT_CODE` of `PROT_READ|PROT_WRITE|PROT_EXEC` is not available to us; we define `WRITE_XOR_EXECUTE_CODE` for that target so that code is written `RW` and then flipped to `RX`, which is what puts `enable_code_write` in our build at all. That is the third bullet above: 4K pages against 16K segments, so `S_getmem` hands back non-segment-aligned blocks as a matter of course and `delta` is usually not zero.

We instrumented two code chunks; both had `chunk->addr` congruent to `0x2000` modulo `0x4000`, i.e. `delta = 0x2000`. A fault at `0x100038f6678` fell inside `[0x100038f6000, 0x100038f8000)` — exactly the abandoned top half of the qualifying segment based at `0x100038f4000` — and `0x100038f6000` is `chunk->addr + 96 * bytes_per_segment`, a *run* boundary that is not a *segment* boundary. The hardware fault decoded as a write-protection fault on a present page (`P=1 W=1 U=1`) whose PTE had `WRITABLE` clear. Applying the change below took the abort and the fault to zero.

That is offered as corroboration only. The report does not depend on it: the mechanism follows from `allocate_segments` and `enable_code_write` alone, and the reproducer below runs on stock Linux.

## The fix

Derive the run addresses from the segments, the way the hint path in the same function already does:

```diff
         int j;
+        /* chunk->addr can be below the chunk's first segment */
+        addr = TO_VOIDP(build_ptr(chunk->base, 0));
         for (j = 0; j < chunk->segs; j++) {
@@
-            addr = TO_VOIDP((char *)chunk->addr + (j + 1) * bytes_per_segment);
+            addr = TO_VOIDP(build_ptr(chunk->base + j + 1, 0));
```

The `chunk->nused_segs == 0` branch is deliberately **not** touched. It protects the whole allocation, head padding included, and pairs `chunk->addr` with `chunk->bytes`, which is itself measured from `chunk->addr`; starting it at `build_ptr(chunk->base, 0)` with length `chunk->bytes` would run `delta` bytes off the end of the mapping. For the same reason the initialisation at `:698` stays where it is — it is the value that branch, and the non-threaded arm below it, both use.

## Why the change is a no-op where the bug cannot occur

When `delta == 0`:

* `build_ptr(chunk->base, 0) == chunk->addr`, so the added initialisation assigns the value `addr` already held; and
* `build_ptr(chunk->base + j + 1, 0) == chunk->addr + (j + 1) * bytes_per_segment`, since `build_ptr(s, 0)` is exactly `s << segment_offset_bits` and `chunk->base` is then `chunk->addr >> segment_offset_bits`.

The two expression sets are equal by construction, so every configuration that could not reach the bug — any target whose pages are at least as coarse as a segment, any non-threaded build, `pb` — sees an identical sequence of `mprotect` calls.

## Testing

**No mat accompanies this change, and I do not think one is possible.** Reaching the defect requires a chunk whose first segment is not at `chunk->addr`, which is a property of the block `S_getmem` happens to be handed. There is no way to ask for one of those from Scheme, and no way to observe the mis-protection from Scheme when you get one — the `mprotect` succeeds either way.

What I can offer instead is a standalone C harness, which I am happy to attach to this PR or to an issue if it is wanted. It does not build or link Chez. It transcribes `chunkinfo`, `build_ptr`, `addr_get_segment`, the relevant part of `allocate_segments` and the run loop; places the "`S_getmem`" mapping by hand so that a chosen `delta` is exercised rather than waited for; calls the real `mprotect(2)`; and probes each qualifying segment with a `SIGSEGV`-guarded store. On a 4K-page host it reports, for `delta = 0x2000`:

```
--- delta = 0x2000, current expressions, S_thread_start_code_write path ---
  mprotect(0x7f3c79f46000, 0x10000, RW)
  mprotect(0x7f3c79f5a000, 0xc000, RW)
  segment 3 at 0x7f3c79f54000: first byte writable, last byte not writable
  segment 7 at 0x7f3c79f64000: first byte writable, last byte not writable

--- delta = 0x2000, proposed expressions, S_thread_start_code_write path ---
  every qualifying segment is uniformly writable

--- delta = 0: current vs proposed, argument for argument ---
  call 0: current chunk+0x0      len 0x10000  | proposed chunk+0x0      len 0x10000
  call 1: current chunk+0x14000  len 0x10000  | proposed chunk+0x14000  len 0x10000
  identical: the change is a no-op when delta == 0
```

On a host whose pages are at least 16K it reports that there is nothing to reproduce. The transcription is short enough to check against `c/segment.c` by eye, and that check is the point. It is not a Chez crash: it demonstrates that the loop's arithmetic mis-protects real pages, which is the defect, not a particular Chez-level failure.

The change is also verifiable by reading, and that is the argument I would rather rest on. `allocate_segments` sets `chunk->base` from `addr_get_segment(addr + bytes_per_segment - 1)`, a round *up*, and takes the `nact += 1` arm only when `build_ptr(base, 0) == addr`.

## Measured on stock x86-64 Linux

The harness above is a transcription, so I also built the real configuration: `main` at `659c263e`, Chez 10.5.0-pre-release.1, machine type `ta6le`, `./configure --threads --disable-x11 --disable-curses CFLAGS+=-DWRITE_XOR_EXECUTE_CODE`, GCC 16.1.0, 4K pages. Instrumenting the run loop to compute the current expression and the segment-true address side by side, and running `mats/7.ms`:

- In all 39 Chez processes that ran long enough to emit a counter dump, **every** chunk walk found `chunk->addr != build_ptr(chunk->base, 0)`, and **every** run-emitting `mprotect` in the walk was issued below the segments it was meant to cover. The longest-running process counted 159,744 chunk walks and 70,256 runs with no exception to either.
- 222 distinct misplaced run-starts were sampled, at 4096, 8192 and 12288 bytes below the intended address — one, two or three pages. A representative line:

```
used=0x4061e000  true=0x40620000  low_by=8192  len=1572864
```

  a 1.5 MB window placed 8K low.

Nothing failed as a result, which is exactly the report's claim: the shifted range is still mapped and still page aligned, so `mprotect` returns 0 and the mis-protection is silent. What the measurement adds is that the precondition is not exotic — on this 4K-page host it was the normal case rather than the rare one, and the offsets are the ones the mechanism predicts.

## Release notes

`release_notes/release_notes.stex` gains a `Bug Fixes` subsection, "Code-write protection of segment runs (10.5.0)".

## Scope

* Both entry points are affected, since `S_thread_start_code_write` and `S_thread_end_code_write` share `enable_code_write`.
* The `maxg == 0 && hint != NULL` and `maxg == 0 && current` paths are not affected: the first already uses `build_ptr`, the second works from `tgc->base_loc`.
* I found no existing issue or PR covering this. Searches of this repository for `enable_code_write`, `WRITE_XOR_EXECUTE_CODE`, `chunk->addr`, `build_ptr` and `bytes_per_segment` returned nothing on point.
* There is a **companion PR**, #1062, for a second, independent address bug in the same function — the `sweep_next` drain loop at `:685`. The two are unrelated in mechanism and touch different lines; either can be taken without the other, and they merge cleanly together.


mflatt:
Although this looks like a valid repair, it's in a part of the implementation that has deeper issues. I'm preparing a separate PR #1066 to offer a change that subsumes this one.

mflatt:
Addressed by #1066

PR #1064, opened 2026-09-15
Windows x64: repair XMM register preservation by callables

mflatt:
Only the low 64 bits were saved for each preserved XMM register, instead of all 128 bits. Also, `__collect_safe` mode could mangle the preserved bits.

See also racket/racket#5489 and racket/ChezScheme#66 (but I should put the latter repo in archived mode).


mflatt:
Compiling "foreign4.c" for tests with optimization exposed an unrelated problem with foreign calls and callables on 32-bit x86 Windows, where (unlike Linux and Mac OS), an argument added to receive a result struct should not be popped by the callee. I still need to add a release note for that.

s/np-languages.ss
```diff
@@ -770,6 +770,7 @@
   (declare-primitive fpcastto/hi value #t) ; 32-bit only
   (declare-primitive fpcastto/lo value #t) ; 32-bit only
   (declare-primitive fpcastfrom value #t) ; 64-bit: 1 argument; 32-bit: 2 arguments
+  (declare-primitive fpmove128 value #t) ; x86)64 only
```
jltaylor-us: I assume that comment is supposed to say `x86_64`

mflatt: Fixed - thanks!

release_notes/release_notes.stex
```diff
@@ -3023,6 +3023,21 @@ argument raises and error and the offset is out of bounds---has been fixed.
...
+
+A foreign callable on x86\_64 Windows did not properly preserve XMM
+registers, keeping only the low 64 bits of the registers in some cases
+and potentially further corrupting the registers values in
+\scheme{__collect_safe} mode. That bug has been fixed.
+
+\subsection{Repairs for x86 Windows foreign calls and callables (10.5.0)}
+
+For a foreign call or callable on x86 Windows with a struct result
+that does not fit into registers, the callee recieves a result pointer
```
burgerrg: receives

s/x86_64.ss
```diff
@@ -1869,6 +1875,12 @@
       (Trivit (dest src)
         (emit sse.movsd src dest code*))))
 
+  (define asm-fpmove128
+    (lambda (code* dest src)
+      (Trivit (dest src)
+	;; assuming aligned
```
burgerrg: The comment should be aligned to match the indentation.

release_notes/release_notes.stex
mflatt: Fixed

s/x86_64.ss
mflatt: Fixed.

burgerrg (approved):
Thank you for this important fix!

PR #1065, opened 2026-09-17
configure: add --pic flag for position-independent kernel builds

xpusostomos:
I'm doing a plugin for a project and am making a Chez extension plugin as a shared library. That means I need to be able to build the kernel compiled as position-independent code. 

--pic appends -fPIC to the kernel CFLAGS. The build system already uses -fPIC -shared for its foreign-loadable objects, so this follows an anticipated configuration. With the flag, kernel.a ends up being PIC code, good for linking into a .so


mflatt:
Does supplying `CFLAGS+=-fPIC` to `configure` work for your case, instead of adding a `--pic` flag?

xpusostomos:
Good question. Yes it does work. I could have sworn I tried something along those lines, but I guess I didn't.  All I would say is it might be nice to make that more explicit somewhere because embedding chez into an add on shared object might a common scenario. If nothing else it, it alerts people that yes chez works in a shared library. (That might be obvious to you, but knowing vaguely what goes on inside of a sophisticated runtime like chez, I didn't know until I tried it).

PR #1067, opened 2026-09-17
update doc to contain openbsd related restrictions

PeterHajdu:


BUILDING
```diff
@@ -348,6 +348,11 @@ The makefile supports several targets:
    Removes all built elements from the workarea, and then removes
    `bin/zuo` (unless configured with `ZUO=<zuo>`).
 
+OPENBSD
+
+For building on OpenBSD the source code needs to be on a filesystem
```
mflatt, line 353: I'd add a comma after `OpenBSD`.

BUILDING
```diff
@@ -348,6 +348,11 @@ The makefile supports several targets:
...
+that is mounted with the wxallowed option. Also it needs to be
```
mflatt, line 354: How about `Chez Scheme` in place of `it`?

BUILDING
```diff
@@ -348,6 +348,11 @@ The makefile supports several targets:
    Removes all built elements from the workarea, and then removes
    `bin/zuo` (unless configured with `ZUO=<zuo>`).
 
+OPENBSD
```
mflatt, line 351: I think this is useful, but ok if we move it down to before `EMSCRIPTEN`, since Windows compilation is probably more common than OpenBSD?

BUILDING
```diff
@@ -348,6 +348,11 @@ The makefile supports several targets:
...
+
+For building on OpenBSD the source code needs to be on a filesystem
+that is mounted with the wxallowed option. Also it needs to be
+installed to such filesystem.
```
mflatt, line 355: `such` -> `such a`

mflatt (commented):
Some suggested changes. I'd be happy to adjust and merge, if you prefer.

Also, assuming that this change is merged first, I can adjust #1066 to note the alternative of configuring with `--enable-wx` (in exchange for losing `__collect_safe` functionality).

PeterHajdu:
Thanks for the review. Is there anything else I can fix?

mflatt:
Thanks!

PR #1068, opened 2026-09-18
small floating-point performance improvements

mflatt:
Mark `fl-make-rectangular` as accepting unboxed arguments, which is needed to comunicate unboxing potential to `let` bindings.

On x86_64, use `movapd` instead of `movsd` for register-to-register moves, since that avoids creating a false dependency on some processors to preserve high bits.


mflatt:
Another instruction-selection improvement from @samth: clear bits before `cvtsi2sd`.

